Issue #9, opened 2015-09-22
Simple volume driver with zero configuration for development.


kunalkushwaha:
I think a simple volume which resides on local host can be useful for new / first time user of osd.

A btrfs or nfs needs some configuration, but a file/folder based volume will be helpful while trying the osd, which require zero configuration.

jvinod:
@kunalkushwaha Sounds good. Let's do a local driver with zero config. Thanks

peter-edge:
Can this be closed? @kunalkushwaha

kunalkushwaha:
Code is merged, Closing.

Issue #7, opened 2015-09-21
Support docker volume API


jvinod:
Docker 1.9 will have a new volumes API - add openstorage support for https://github.com/docker/docker/pull/14242

kunalkushwaha:
I am currently working on this issue.

Findings:

Docker volume APIs identify volumes with their Name. 
Name of volume either system generated uuid or user can specify the name in argument.


```
$ docker volume  create -d vfs --name=vol1
vol1

$ docker volume  create -d vfs
fd2913a52b3e2d7f8de0d83e8e53142d4bc8c093cd0c9c33033ffa1f60116e0d

$ docker volume  ls
DRIVER              VOLUME NAME
local               8b26f006dc081110eeb8aee577289c25b58d8db7d780eb42a5209d36e4d42976
vfs                 vol1
vfs                 fd2913a52b3e2d7f8de0d83e8e53142d4bc8c093cd0c9c33033ffa1f60116e0d
```

In OSD, uuid generated by OSD is key to identify the volumes. 
So to integrate with Volume APIs, mapping with osd uuid and Name is required, and Path API of docker, should lookup with Volume Name instead of current osd's uuid.

- Also, driver specific parameters can be passed as ``--opts``, so nothing is required there.
I feel one more Method in Docker Volume API is required to print help of driver parameters.

jvinod:
What stops us from creating a volume with the same name as passed in from the volumes API? I.e the VolumeID == Volume name from from the volumes API

kunalkushwaha:
Creation is successful, but path of volume cannot be found, as Docker invokes path for "Name" not uuid.

NOTE: Also, we need to keep track with https://github.com/docker/docker/issues/15997 & https://github.com/docker/docker/pull/16534

jvinod:
Understood. I am suggesting we don't create the UUID if the name is provided. I.e the volumeID  it the name that is provided. If it is not provided, we generate a UUID.  

So in your examples:
```
$ docker volume  create -d osd --name=vol1
vol1
#VolumeID == vol1

$ docker volume  create -d osd
cfbb378d-1e0b-4321-8946-63cbb76ccd4d
#VolumeID = cfbb378d-1e0b-4321-8946-63cbb76ccd4d

$ docker volume  ls
DRIVER              VOLUME NAME
osd                 vol1
osd                 cfbb378d-1e0b-4321-8946-63cbb76ccd4d
```

kunalkushwaha:
[...]

PR #10, opened 2015-09-23
localfs, a zero-configuration driver 

kunalkushwaha:
Sample code for review and understanding driver code.

Few things is not clear to me.
1. Create()- In create function, the folder structure that is created, doesnt make any sense while mounting.
2. I have not implement, Mount function, since it is not required.
3. After mounting the volume in docker container, the data changes in that mounted folder is reflected in ``/var/lib/docker/volume/<volume-name>``
4. I want to understand, how the create/mount  folder/file is reflected to actual IO Path. It is not clear to me.

NOTE: This converted to local from nfs-driver.

Signed-off-by: Kunal Kushwaha <kunal.kushwaha@gmail.com>

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -4,6 +4,7 @@ osd:
     #nodeid: "1"
     #clusterid: "deadbeeef"
   drivers:
+      vfs:
 #    pwx:
 #   nfs:
 #     server: "localhost"
--- a/drivers.go
+++ b/drivers.go
@@ -3,6 +3,7 @@ package main
 import (
 	"github.com/libopenstorage/openstorage/drivers/aws"
 	"github.com/libopenstorage/openstorage/drivers/btrfs"
+	"github.com/libopenstorage/openstorage/drivers/vfs"
 	"github.com/libopenstorage/openstorage/drivers/nfs"
 	"github.com/libopenstorage/openstorage/drivers/pwx"
 	"github.com/libopenstorage/openstorage/volume"
@@ -25,5 +26,7 @@ var (
 		{driverType: btrfs.Type, name: btrfs.Name},
 		// PWX driver provisions storage from PWX cluster.
 		{driverType: pwx.Type, name: pwx.Name},
+		// VFS driver provisions storage from local filesystem
+		{driverType: vfs.Type, name: vfs.Name},
 	}
 )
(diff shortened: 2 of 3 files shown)


kunalkushwaha:
Okay, I got the issue.

``local`` is default driver of docker volume and Should not be used by anyone.

Though still facing issue of kvdb. 

```
Starting volume driver:  newlocal
INFO[0000] Starting REST service on /var/lib/osd/driver/newlocal.sock 
INFO[0000] Starting REST service on /run/docker/plugins/newlocal.sock 
DEBU[0008]                                               Driver=newlocal ID=vol5 Request=create
DEBU[0008]                                               Driver=newlocal ID=vol5 Request=create
WARN[0008] 404 Failed to locate volume: Cannot locate volume vol5  Driver=newlocal ID=vol5 Request=create


```

jvinod:
@kunalkushwaha 
- Note that the nfs driver in the absence of a server defaults to a vfs driver
- regarding the 404 : api/server/docker.go (the volume plugin implementation) expects the vol to exist on a create. The volume should be created separately using 'osd cli' before using it with docker. This will not be necessary after we address https://github.com/libopenstorage/openstorage/issues/7
- you'd still need to implement Mount (bind mount to specified path) and Unmount
- you don't need to implement the pseudo block device inside the vfs path.

kunalkushwaha:
@jvinod Problem is with kvdb driver.
The Path() function of docker.go , tries to get path from kvdb, but it is not getting the ID back.
I think issue is with kvdb code.

Will try to debug in detail.

kunalkushwaha:
@jvinod I got the issue now. Bacially as ID it expects the hex is e.g. ``d9f6d6bd-7cfb-4b89-acf5-26ecd994e213`` instead of name.

Shouldn't be a mapping between ID and name be there, so both can work?

kunalkushwaha:
Fix for https://github.com/libopenstorage/openstorage/issues/9

kunalkushwaha:
all review comments are incorporated.

jvinod:
Everthing else looks good ..

jvinod:
@kunalkushwaha could we add a vfs_test.go? Thanks..

PR #15, opened 2015-10-04
Builds osd binary in docker container and copy osd image to host.

kunalkushwaha:
Signed-off-by: Kunal Kushwaha <kunal.kushwaha@gmail.com>

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+echo "Building osd in docker"
+make docker
+docker stop osd-dev
+docker rm osd-dev
+docker run --name osd-dev -itd osd bash
+echo "Copying osd binary to host"
+docker cp osd-dev:/bin/osd .


peter-edge:
This won't work for OS X, for example, if there was ever a desire for a platform-agnostic client (you'd have to do a cross-compile build). With the new docker images, I'm not sure if this is relevant anymore, shoot.

jvinod:
@peter-edge what's the recommended way to get a binary on your box with the build done in docker

peter-edge:
docker cp, I am on my phone right now but I'll link you an example in a bit.

jvinod:
cool...how;s that different from just -v 'ing?

peter-edge:
Remote machines - if you are running from Mac, connecting to a machine
that's not a local vm, host volumes do not apply.

jvinod:
ah

kunalkushwaha:
Hi,

In my dev environment, I use two ways and both works fine.

1. Build osd in docker container and copy to host.
2. Build osd in docker container with folders like /var/lib/plugins,
/var/lib/osd etc mounted in container in privileged mode.
    - Then run the osd using docker exec and another docker exec for volume
creation etc.
    - Benefit of this approach is the whole build and runtime environment
is shared with docker environment, but still independent.

I think second approach can work for OSX also. If docker runs in one
container, folders can be shared with build container of osd.

Regards,
Kunal Kushwaha

------------------------------

peter-edge:
@jvinod what are we doing about this?

kunalkushwaha:
This is no more valid now on current file. Closing this.

PR #21, opened 2015-10-06
Makefile suggestions

peter-edge:
This is not ready to merge, but more of a conversation piece that I thought you might find helpful :)

Two reference points for Makefiles I use:

- https://github.com/pachyderm/pachyderm/blob/master/Makefile
- https://github.com/peter-edge/go-flights/blob/master/Makefile

So some themes here:

- I figure GOPATH is up to the user to set, esp as some people like to have it as an individual thing per repository.
- No need to do a blank assignment for BUILD_OPTIONS
- Not everyone uses ctags, no real use for the Makefile
- Removed @'s, I like to see what is going on
- Added how I update code :)
- Added godep usage with golang 1.5 vendor experiment. We can talk about this.
- My pretest check suite: lint, vet, errcheck. Not everyone likes this, but put it in there for reference
- Tests in docker (I think all the tests here should just be in Docker, we do this in Pachyderm, but we can discuss/I'm happy to help out).
- This is in preparation for all tests in docker/CI.


peter-edge:
Oh, also removes the docker rmi line - this kills the caching

gourao:
Removing the docker rmi looks good...  Thanks for catching this!

jvinod:
I like all the suggestions. I am not caught up with GO15VENDOREXPERIMENT yet, need to understand that...

peter-edge:
Check the link I posted in the associated issue for the doc on it :)

peter-edge:
https://github.com/libopenstorage/openstorage/pull/24

PR #27, opened 2015-10-09
Switch to go-dockerclient

peter-edge:
https://github.com/fsouza/go-dockerclient

In my experience, it's the most actively maintained golang docker API client. Kubernetes uses it, among a lot of people. And I also happen to be a maintainer of it :)


jvinod:
SGTM, we're not tied to the samalba client .. The only potential "breaking" change for our deployment might be that it takes in its config from environment variables::
```
-	docker, err := dockerclient.NewDockerClient(dockerHost, nil)
+	docker, err := docker.NewClientFromEnv()
```
which might affect our deployment.

ping @gourao

peter-edge:
It's the same thing as you have now, if DOCKER_HOST is not set, it defaults
to the default UNIX socket. This constructor also handles DOCKER_TLS_VERIFY
and DOCKER_CERT_PATH.

jvinod:
sure,  we can always set the env. Hypothetically, how do simultaneously talk to two different docker api endpoints?

peter-edge:
Heh. More fun configuration that we can add. You'll have to have a strict
with docker host, TLS verify, CERT path, and then pass it into the
constructor.

jvinod:
:) Are you saying there's already such a constructor ? I  see there's a NewClient(endpoint string) -- that might help?

peter-edge:
Yes :) there's about 20 constructors, half of which I put in the codebase
over time for various reasons :)

jvinod:
Cool :)  Let's  pass in the dockerclient as an explicit argument  instead of env vars?  what do you think ?

peter-edge:
Oh sure, I was just going with what was already there. I'll make that
change now.

peter-edge:
Can we get this in? Want to move on from it :)

jvinod:
oh yes, missed this.

PR #28, opened 2015-10-09
Fix mount test panic

peter-edge:
If the length is not 0, then `p[0]` will panic. `require.Equal` is to `assert.Equal` as `t.Fatalf` is to `t.Errorf`.


jvinod:
Thanks!

PR #29, opened 2015-10-09
Move osd.go to cmd/osd/main.go, move drivers.go to drivers/drivers.go

peter-edge:


peter-edge:
As it is now, the Dockerfile actually specifies the wrong command because "make install" will not name the binary "osd".

jvinod:
Nice..

PR #31, opened 2015-10-09
Fix make docker-test

peter-edge:
This takes care of varying places where required programs can be, along with flags needed for docker run. In the long term, the exec.Command calls should go away (I have a way to set this up with btrfs I can do in the future so there is no need to mount in the test).

The goal here is a consistent test environment, regardless of installed state.


jvinod:
Looks good . Since 'which' can find it, would it have been enough to blow away the absolute path and just pass the exec name ?

peter-edge:
Actually I think so, heh.

peter-edge:
I removed the which function, I also removed building btrfs-progs from source and just installing btrfs-tools in the Dockerfile, it passes :)

jvinod:
awesome :)

Issue #25, opened 2015-10-09
Running OSD  aws driver in docker


jvinod:
OSD aws  driver in docker doesn't work:
ubuntu@ip-172-31-17-72:~/go/src/github.com/libopenstorage/openstorage$ docker run --privileged -v $PWD:/etc -v /usr/share/docker/plugins:/usr/share/docker/plugins -v /var/lib/osd/driver:/var/lib/osd/driver -v /mnt:/mnt osd /bin/osd -d -f /etc/config.yaml
Starting volume driver:  aws
time="2015-10-08T14:16:58Z" level=info msg="AWS instance i-75ecbeb5 zone us-west-1b" 
Unable to start volume driver:  aws RequestError: send request failed
caused by: Post https://ec2.us-west-1.amazonaws.com/: x509: failed to load system roots and no roots provided

gourao:
@peter-edge Could you take a look at this?  It happens only when the OSD runs  in Docker.  It works when OSD is run at the host.

gourao:
Closing... AWS lib issue

lpabon:
Closing since it has had no update for more than 6 months. Please reopen if needed.

PR #34, opened 2015-10-09
fix ca certificates

peter-edge:
https://github.com/libopenstorage/openstorage/issues/25

Have not tested but almost guarantee this fixes it.

Another alternative is https://github.com/peter-edge/go-cacerts but I actually do not use that as much anymore.


jvinod:
I tested it with that change, I still get that error. Seems to be coming from the aws-go package .. regardless the change makes sense ..

PR #35, opened 2015-10-09
Fix go vet

peter-edge:


jvinod:
hmph, i wish the compiler would catch those :)

PR #37, opened 2015-10-09
Add btrfs unsupported.go for not linux and cgo

peter-edge:
There are other places we should have similar logic, but we'll get there.


jvinod:
Yes!

PR #39, opened 2015-10-09
Base FUSE volume driver

peter-edge:
This isn't ready for merging yet, I need to find a good example FS to test with.

The concept here is that if someone implements bazil.org/fuse/fs.Filesystem, they can automatically add a volume.VolumeDriver. This would let Pachyderm immediately/generically add PFS as a volume.VolumeDriver, for example.

I basically took the code from the vfs driver, and the common code should be combined I would think. This was more a test for me to get some of the basics.


peter-edge:
Why this is cool: https://godoc.org/bazil.org/fuse/fs?importers

jvinod:
I see how this can be useful in a variety of ways - even perhaps a remote attach..

peter-edge:
Can we just merge this in for now and I'll work off it?

sathieu:
How can I enable this driver?

PR #73, opened 2015-12-14
Change all log.Warn/Warnf calls to log.Error/Errorf

peter-edge:
When I tried to boot up osd, I got this:

```
$ osd -d -f config_btrfs.yaml
INFO[0000] Starting volume driver: btrfs
WARN[0000] Unable to start volume driver: btrfs, no such file or directory
```

When I saw it was just a warning, my first thought went to "did the osd daemon still boot up? this seems fatal", so I did `ps -ef | grep osd`, which of course returned nothing.

Warn in Go World means "this is a warning, but I'm going to keep going". Error/Fatal are things where you expect the execution path to stop, although personally I dislike using Fatal as it does an `os.Exit(...)` call usually, which I think is bad for the middle of execution.

### DIFF
--- a/api/server/docker.go
+++ b/api/server/docker.go
@@ -59,7 +59,7 @@ func volDriverPath(method string) string {
 
 func (d *driver) volNotFound(request string, id string, e error, w http.ResponseWriter) error {
 	err := fmt.Errorf("Failed to locate volume: " + e.Error())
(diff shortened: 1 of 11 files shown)


peter-edge:
https://github.com/libopenstorage/openstorage/pull/94

Issue #74, opened 2015-12-14
Getting started script for btrfs


peter-edge:
I tried to do `osd -d -f config_btrfs.yaml`, which of course does not work with out all the normal btrfs fun times :) I think that the btrfs driver should be the good getting started case, as it does not require any external dependencies from your local (linux) machine like nfs and aws do (thoughts on that?).

I have this https://github.com/peter-edge/dotfiles/blob/master/bin/setup_docker_storage.sh which I can adapt for osd, what does everyone think?

PR #75, opened 2015-12-14
add etc/btrfs/init.sh and make launch-local-btrfs

peter-edge:
I have a feeling that the mount in the script should not be there, when I killed osd after running make launch-local-btrfs, I have no way to unmount /var/lib/openstorage/btrfs, it is always busy.

We need to visit this and do documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,10 @@ launch: docker-build-osd
 		-v /var/lib/osd/:/var/lib/osd/\
 		openstorage/osd -d -f /etc/config.yaml
 
+launch-local-btrfs: install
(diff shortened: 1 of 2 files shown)


peter-edge:
cc @gourao

peter-edge:
https://github.com/libopenstorage/openstorage/issues/74

peter-edge:
Note that vfs DOES do the setup of the directory, so we should choose one or the other

peter-edge:
ping @gourao @jvinod

peter-edge:
ping! can we just merge this?

peter-edge:
@gourao @jvinod

PR #94, opened 2015-12-16
Change all logrus.Warn/Warnf calls to logrus.Error/Errorf

peter-edge:


peter-edge:
I'm going to do this separately, and probably with go.pedge.io/lion

PR #111, opened 2016-01-03
Refactor beginning

peter-edge:
This is the beginning of the refactor. I want to merge what there is so far sooner than later because the primary change here is moving the data objects to protobuf, which made a lot of changes in other places (especially around always using pointers). My hope is that we can "fully" test this (ie use your any internal testing you have with internal drivers) and then merge it.

We should talk about this change not over email.


peter-edge:
Ping

peter-edge:
@jvinod cli protobuf json stuff fixed

peter-edge:
You might want to check that I didnt mess up https://github.com/libopenstorage/openstorage/commit/632cfaf02b5dabf8a4be22af80f7b5b0c9d4d91a but this commit needs work

PR #114, opened 2016-01-13
CoprHD Prototype Support

rodriguise:
Provides initial support for creating volumes using the coprhd REST API.


volume/drivers/coprhd/coprhd.go
```diff
@@ -0,0 +1,262 @@
+package coprhd
+
+import (
+	"crypto/tls"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	log "github.com/Sirupsen/logrus"
```
peter-edge: Logrus

volume/drivers/coprhd/coprhd.go
```diff
@@ -0,0 +1,262 @@
...
+
+	"github.com/portworx/kvdb"
+
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/libopenstorage/openstorage/volume"
+
+	napping "gopkg.in/jmcvetta/napping.v3"
```
peter-edge: Why does this need naming?

volume/drivers/coprhd/coprhd.go
```diff
@@ -0,0 +1,262 @@
...
+)
+
+const (
+	Name = "coprhd"
+	Type = api.Block
+
+	// URI_LOGIN path to create a authentication token
+	URI_LOGIN = "login.json"
```
peter-edge: Camel case

volume/drivers/coprhd/coprhd.go
```diff
@@ -0,0 +1,262 @@
...
+		} `json:"task"`
+	}
+)
+
+func init() {
+	volume.Register(Name, Init)
+}
+
+func Init(params volume.DriverParams) (volume.VolumeDriver, error) {
+
```
peter-edge: Newline

rodriguise:
The goimports tool removes unused imports, since this package has a unconventional name it gets removed without an explicit name.

volume/drivers/coprhd/coprhd.go
rodriguise: The goimports tool removes unused imports, since this package has a unconventional name it gets removed without an explicit name.

volume/drivers/coprhd/coprhd.go
rodriguise: Is this a question?

jvinod: I think the comment is to use logrus as-is without renaming the import. (as part of cleanup we did that in the repo)

peter-edge:
Are you using a plugin like vim-go for goimports? It does not remove my imports with unconventional names

PR #126, opened 2016-03-21
Flexvolume for k8s integration

adityadani:
- Implemented the attach/detach/mount/unmount functions in flexvolume server plugin.
- The flexvolume client now writes to stdout the way kubernetes flexvolume plugin expects for each command.
- Modified the mounts package to provide devicePath given a mount point.

Note: Currently we are storing a mapping of devices and their volume drivers in a in-memory map. If the osd daemon dies and upon restart it does not find a volume driver for a device then it defaults to pwx driver.

### DIFF
--- a/api/server/flexvolume.go
+++ b/api/server/flexvolume.go
@@ -1,27 +1,145 @@
 package server
 
-type flexVolumeClient struct{}
+import (
+	"fmt"
+
+	"github.com/libopenstorage/openstorage/pkg/mount"
+	"github.com/libopenstorage/openstorage/volume"
+
+	"go.pedge.io/dlog"
+)
+
+const (
+	osdDriverKey = "osdDriver"
+	volumeIdKey  = "volumeID"
+	pxDriverKey  = "pxd"
+)
+
+var (
+	deviceDriverMap map[string]string
+)
+
+type flexVolumeClient struct {
+}
 
 func newFlexVolumeClient() *flexVolumeClient {
+	deviceDriverMap = make(map[string]string)
+
 	return &flexVolumeClient{}
 }
 
+func (c *flexVolumeClient) getVolumeDriver(driverName string) (volume.VolumeDriver, error) {
+	driver, err := volume.Get(driverName)
+	if err != nil {
+		return nil, fmt.Errorf("%v driver not initialized in OSD", driverName)
+	}
+	return driver, nil
+}
 func (c *flexVolumeClient) Init() error {
 	return nil
 }
 
 func (c *flexVolumeClient) Attach(jsonOptions map[string]string) error {
+	driverName, ok := jsonOptions[osdDriverKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No OSD driver specified in flexvolume")
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	mountDevice, ok := jsonOptions[volumeIdKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No volumeID specified in flexvolume")
+	}
+
+	_, err = driver.Attach(mountDevice)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
 func (c *flexVolumeClient) Detach(mountDevice string) error {
+	var driverName string
+	driverName, ok := deviceDriverMap[mountDevice]
+	if !ok {
+		dlog.Infof("Could not find driver for (%v). Defaulting to pxd", mountDevice)
+		driverName = pxDriverKey
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	err = driver.Detach(mountDevice)
+	if err != nil {
+		return err
+	}
 	return nil
 }
 
-func (c *flexVolumeClient) Mount(targetMountDir string, mountDevice string, jsonOptions map[string]string) error {
+func (c *flexVolumeClient) Mount(targetMountDir string, mountDevice string,
+	jsonOptions map[string]string) error {
+	driverName, ok := jsonOptions[osdDriverKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No OSD driver specified in flexvolume")
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	if targetMountDir == "" {
+		return fmt.Errorf("Missing target mount path")
+	}
+
+	err = driver.Mount(mountDevice, targetMountDir)
+	if err != nil {
+		return fmt.Errorf("Failed in mount. Driver returned error : (%v)", err)
+	}
+
+	// Update the deviceDriverMap
+	mountManager, _ := mount.New(mount.DeviceMount, "")
+	devPath, err := mountManager.GetDevPath(targetMountDir)
+	if err != nil {
+		return fmt.Errorf("Mount Failed. Could not find mountpoint in /proc/self/mountinfo")
+	}
+	deviceDriverMap[devPath] = driverName
+
 	return nil
 }
 
 func (c *flexVolumeClient) Unmount(mountDir string) error {
+	var driverName string
+
+	// Get the mountDevice from mount manager
+	mountManager, _ := mount.New(mount.DeviceMount, "")
+	mountDevice, err := mountManager.GetDevPath(mountDir)
+	if err != nil {
+		return fmt.Errorf("Invalid mountDir (%v). No device mounted on this path", mountDir)
+	}
+
+	driverName, ok := deviceDriverMap[mountDevice]
+	if !ok {
+		dlog.Infof("Could not find driver for (%v). Defaulting to pxd", mountDevice)
+		driverName = pxDriverKey
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	err = driver.Unmount(mountDevice, mountDir)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/api/server/flexvolume.go
+++ b/api/server/flexvolume.go
@@ -13,11 +13,9 @@ import (
 const (
 	osdDriverKey = "osdDriver"
 	volumeIDKey  = "volumeID"
-	pxDriverKey  = "pxd"
 )
 
 var (
-	deviceDriverMap map[string]string
 	// ErrInvalidSpecDriver is returned when incorrect or no OSD driver is specified in spec file
 	ErrInvalidSpecDriver = errors.New("Invalid kubernetes spec file : No OSD driver specified in flexvolume")
 	// ErrInvalidSpecVolumeID is returned when incorrect or no volumeID is specified in spec file
@@ -26,9 +24,12 @@ var (
 	ErrNoMountInfo = errors.New("Could not read mountpoints from /proc/self/mountinfo. ")
 	// ErrMissingMountPath is returned when no mountPath specified in mount call
 	ErrMissingMountPath = errors.New("Missing target mount path")
+	deviceDriverMap     = make(map[string]string)
 )
 
-type flexVolumeClient struct{}
+type flexVolumeClient struct {
+	defaultDriver string
+}
 
 func (c *flexVolumeClient) Init() error {
 	return nil
@@ -39,18 +40,15 @@ func (c *flexVolumeClient) Attach(jsonOptions map[string]string) error {
 	if !ok {
 		return ErrInvalidSpecDriver
 	}
-
 	driver, err := c.getVolumeDriver(driverName)
 	if err != nil {
 		return err
 	}
-
 	mountDevice, ok := jsonOptions[volumeIDKey]
 	if !ok {
 		return ErrInvalidSpecVolumeID
 	}
-
-	if _, err = driver.Attach(mountDevice); err != nil {
+	if _, err := driver.Attach(mountDevice); err != nil {
 		return err
 	}
 	return nil
@@ -59,16 +57,14 @@ func (c *flexVolumeClient) Attach(jsonOptions map[string]string) error {
 func (c *flexVolumeClient) Detach(mountDevice string) error {
 	driverName, ok := deviceDriverMap[mountDevice]
 	if !ok {
-		dlog.Infof("Could not find driver for (%v). Defaulting to pxd", mountDevice)
-		driverName = pxDriverKey
+		dlog.Infof("Could not find driver for (%v). Resorting to default driver ", mountDevice)
+		driverName = c.defaultDriver
 	}
-
 	driver, err := c.getVolumeDriver(driverName)
 	if err != nil {
 		return err
 	}
-
-	if err = driver.Detach(mountDevice); err != nil {
+	if err := driver.Detach(mountDevice); err != nil {
 		return err
 	}
 	return nil
@@ -81,27 +77,22 @@ func (c *flexVolumeClient) Mount(targetMountDir string, mountDevice string,
 	if !ok {
 		return ErrInvalidSpecDriver
 	}
-
 	driver, err := c.getVolumeDriver(driverName)
 	if err != nil {
 		return err
 	}
-
 	if targetMountDir == "" {
 		return ErrMissingMountPath
 	}
-
-	if err = driver.Mount(mountDevice, targetMountDir); err != nil {
+	if err := driver.Mount(mountDevice, targetMountDir); err != nil {
 		return err
 	}
-
 	// Update the deviceDriverMap
 	mountManager, err := mount.New(mount.DeviceMount, "")
 	if err != nil {
 		dlog.Infof("Could not read mountpoints from /proc/self/mountinfo. Device - Driver mapping not saved!")
 		return nil
 	}
-
 	devPath, err := mountManager.GetDevPath(targetMountDir)
 	if err != nil {
 		return fmt.Errorf("Mount Failed. Could not find mountpoint in /proc/self/mountinfo")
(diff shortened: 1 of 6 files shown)


adityadani:
@jvinod Can you also take a look at this ?

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+type flexVolumeClient struct {
+}
 
 func newFlexVolumeClient() *flexVolumeClient {
+	deviceDriverMap = make(map[string]string)
+
 	return &flexVolumeClient{}
 }
 
+func (c *flexVolumeClient) getVolumeDriver(driverName string) (volume.VolumeDriver, error) {
```
peter-edge: private functions below public functions https://github.com/libopenstorage/openstorage/blob/master/STYLEGUIDE.md

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	return driver, nil
+}
 func (c *flexVolumeClient) Init() error {
 	return nil
 }
 
 func (c *flexVolumeClient) Attach(jsonOptions map[string]string) error {
+	driverName, ok := jsonOptions[osdDriverKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No OSD driver specified in flexvolume")
```
peter-edge: constant

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	mountDevice, ok := jsonOptions[volumeIdKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No volumeID specified in flexvolume")
```
peter-edge: constant

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+		dlog.Infof("Could not find driver for (%v). Defaulting to pxd", mountDevice)
+		driverName = pxDriverKey
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	err = driver.Detach(mountDevice)
```
peter-edge: scope err inside if https://github.com/libopenstorage/openstorage/blob/master/STYLEGUIDE.md

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	if err != nil {
+		return err
+	}
+
+	mountDevice, ok := jsonOptions[volumeIdKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No volumeID specified in flexvolume")
+	}
+
+	_, err = driver.Attach(mountDevice)
```
peter-edge: scope err inside if https://github.com/libopenstorage/openstorage/blob/master/STYLEGUIDE.md

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	}
 	return nil
 }
 
-func (c *flexVolumeClient) Mount(targetMountDir string, mountDevice string, jsonOptions map[string]string) error {
+func (c *flexVolumeClient) Mount(targetMountDir string, mountDevice string,
+	jsonOptions map[string]string) error {
+	driverName, ok := jsonOptions[osdDriverKey]
+	if !ok {
+		return fmt.Errorf("Invalid kubernetes spec file : No OSD driver specified in flexvolume")
```
peter-edge: constant

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
+	if targetMountDir == "" {
+		return fmt.Errorf("Missing target mount path")
+	}
+
+	err = driver.Mount(mountDevice, targetMountDir)
```
peter-edge: scope err inside if - last comment of this, but do it for the rest of the changes

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	if err != nil {
+		return fmt.Errorf("Failed in mount. Driver returned error : (%v)", err)
```
peter-edge: just return err?

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	}
+
+	// Update the deviceDriverMap
+	mountManager, _ := mount.New(mount.DeviceMount, "")
```
peter-edge: is the second argument an error? if so, do not ignore

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	devPath, err := mountManager.GetDevPath(targetMountDir)
+	if err != nil {
+		return fmt.Errorf("Mount Failed. Could not find mountpoint in /proc/self/mountinfo")
+	}
+	deviceDriverMap[devPath] = driverName
+
```
peter-edge: less newlines https://github.com/libopenstorage/openstorage/blob/master/STYLEGUIDE.md

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
 	return nil
 }
 
 func (c *flexVolumeClient) Unmount(mountDir string) error {
+	var driverName string
+
+	// Get the mountDevice from mount manager
+	mountManager, _ := mount.New(mount.DeviceMount, "")
```
peter-edge: error on 2nd argument?

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
 		&AttachRequest{
 			JsonOptions: jsonOptions,
 		},
 	)
+	if err == nil {
+		a := AttachSuccessOutput{
+			Status: "Success",
+			Device: jsonOptions["volumeID"],
+		}
+		output, _ = json.Marshal(a)
```
peter-edge: do not ignore error

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+	} else {
+		a := FailureOutput{
+			Status:  "Failure",
+			Message: err.Error(),
+		}
+		output, _ = json.Marshal(a)
```
peter-edge: do not ignore error, same for rest of the changes

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+	}
+
+	os.Stdout.Write(output)
```
peter-edge: check error

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
 	return err
 }
 
 func (c *client) Detach(mountDevice string) error {
+	var output []byte
```
peter-edge: just do the output outside the if statements, no need to declare

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
-	_, err := c.apiClient.Mount(
-		context.Background(),
-		&MountRequest{
-			TargetMountDir: targetMountDir,
-			MountDevice:    mountDevice,
-			JsonOptions:    jsonOptions,
-		},
-	)
+	err := os.MkdirAll(targetMountDir, os.ModeDir)
+	var output []byte
```
peter-edge: output outside of if statements, declare a

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+			context.Background(),
+			&MountRequest{
+				TargetMountDir: targetMountDir,
+				MountDevice:    mountDevice,
+				JsonOptions:    jsonOptions,
+			},
+		)
+		if err == nil {
+			a := SuccessOutput{
+				Status: "Success",
```
peter-edge: Success or Failure should be a constant

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+
 	_, err := c.apiClient.Detach(
 		context.Background(),
 		&DetachRequest{
 			MountDevice: mountDevice,
 		},
 	)
+
+	if err == nil {
+		a := SuccessOutput{
```
peter-edge: success output can just be a constant

```go
var (
  successBytes = []byte(`{"Status":"Success"}`)
)
```

func init() {
  successBytes_, err := json.Marshal(
    map[string]
  )
}
```

File: pkg/flexvolume/client.go [hunk 19]
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+			Status: "Success",
+		}
+		output, _ = json.Marshal(a)
+	} else {
+		a := FailureOutput{
```
peter-edge:
```go
func newFailureOutput(err error) []byte {
  return []byte(fmt.Sprintf(`{"Status":"Failure", "Message":"%s"}`, err.Error())
}
```

pkg/flexvolume/client.go
```diff
@@ -9,6 +12,20 @@ type client struct {
 	apiClient APIClient
 }
 
+type AttachSuccessOutput struct {
```
peter-edge: Does kubernetes actually parse this in some way?

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
 	_, err := c.apiClient.Attach(
 		context.Background(),
 		&AttachRequest{
 			JsonOptions: jsonOptions,
 		},
 	)
+	if err == nil {
+		a := AttachSuccessOutput{
+			Status: "Success",
+			Device: jsonOptions["volumeID"],
```
peter-edge: check if volumeID is a key?

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+			Status: "Success",
+		}
+		output, _ = json.Marshal(a)
+	} else {
+		a := FailureOutput{
+			Status:  "Failure",
+			Message: err.Error(),
+		}
+		output, _ = json.Marshal(a)
+	}
```
peter-edge: func newOutput(err error) []byte {
  if err != nil {
    return newFailureOutput(err)
  }
  return successOutput
}

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
+
+	os.Stdout.Write(output)
```
peter-edge: check error

pkg/flexvolume/client.go
```diff
@@ -22,43 +39,119 @@ func (c *client) Init() error {
...
 func (c *client) Mount(targetMountDir string, mountDevice string, jsonOptions map[string]string) error {
-	_, err := c.apiClient.Mount(
-		context.Background(),
-		&MountRequest{
-			TargetMountDir: targetMountDir,
-			MountDevice:    mountDevice,
-			JsonOptions:    jsonOptions,
-		},
-	)
+	err := os.MkdirAll(targetMountDir, os.ModeDir)
```
peter-edge:
```go
if err := os.MkdirAll(targetMountDir, os.ModeDir); err == nil {

} else {

}
```

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	_, err = driver.Attach(mountDevice)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
 func (c *flexVolumeClient) Detach(mountDevice string) error {
+	var driverName string
```
peter-edge: no need to declare driverName

peter-edge:
Did a pass for style mostly. The main things are scoping of errors and other variables, variable names, newlines, no unchecked errors, and split out the common logic for the JSON output (I added a suggestion(s) for it). I'll go over other stuff after this. Check out the style guide for more details.

peter-edge:
Two other things:

- The global variable map - strongly do not like :)
- Defaulting to pwx - let's not rename this to storage, and keep it named openstorage :p

api/server/flexvolume.go
```diff
@@ -1,27 +1,145 @@
...
+	osdDriverKey = "osdDriver"
+	volumeIdKey  = "volumeID"
+	pxDriverKey  = "pxd"
+)
+
+var (
+	deviceDriverMap map[string]string
+)
+
+type flexVolumeClient struct {
```
peter-edge: one line, see near the bottom of the style guide https://github.com/libopenstorage/openstorage/blob/master/STYLEGUIDE.md

pkg/flexvolume/client.go
adityadani: Yes kubernetes flexvolume util watches the stdout when it calls the flexvolume plugin's commands. They expect a json output with pre-defined key value pairs in it.

pkg/flexvolume/client.go
adityadani: If we have an invalid volumeID, then the flexvolume server itself will catch it and return an error. We reach this if condition only when everything went good and we actually had a volumeID as a key.

pkg/flexvolume/client.go
adityadani: I have completely removed json marshalling and replaced it with your suggestion of []byte success and failure constants.

adityadani:
I second your feeling of not maintaining a global variable map. I am taking this is as a temporary solution to get the wheels rolling. My next task is to see how we can use etcd/kvdb configured in k8s for our use in openstorage. Once that is figured out we can get rid of this map.

I did not quite follow your second comment about defaulting to pwx.

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+)
+
+const (
+	osdDriverKey = "osdDriver"
+	volumeIDKey  = "volumeID"
+	pxDriverKey  = "pxd"
+)
+
+var (
+	deviceDriverMap map[string]string
```
peter-edge: private below public

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	if err != nil {
+		return err
+	}
+
+	mountDevice, ok := jsonOptions[volumeIDKey]
+	if !ok {
+		return ErrInvalidSpecVolumeID
+	}
+
+	if _, err = driver.Attach(mountDevice); err != nil {
```
peter-edge:
```go
if _, err := driver.Attach(...)
```

`:=` instead of `=`

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	if !ok {
+		dlog.Infof("Could not find driver for (%v). Defaulting to pxd", mountDevice)
+		driverName = pxDriverKey
+	}
+
+	driver, err := c.getVolumeDriver(driverName)
+	if err != nil {
+		return err
+	}
+
```
peter-edge: so many newlines :)

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	if err = driver.Detach(mountDevice); err != nil {
```
peter-edge: same comment about making `err` actually scoped to the if statement, `:=` instead of `=` for the rest of the changes

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	}
+
+	if err = driver.Mount(mountDevice, targetMountDir); err != nil {
+		return err
+	}
+
+	// Update the deviceDriverMap
+	mountManager, err := mount.New(mount.DeviceMount, "")
+	if err != nil {
+		dlog.Infof("Could not read mountpoints from /proc/self/mountinfo. Device - Driver mapping not saved!")
```
peter-edge: can you explain why not returning the error and ignoring it?

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	devPath, err := mountManager.GetDevPath(targetMountDir)
+	if err != nil {
+		return fmt.Errorf("Mount Failed. Could not find mountpoint in /proc/self/mountinfo")
+	}
+	deviceDriverMap[devPath] = driverName
 	return nil
 }
 
 func (c *flexVolumeClient) Unmount(mountDir string) error {
+	var driverName string
```
peter-edge: why declaring this here?

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	}
+
+	if err = driver.Unmount(mountDevice, mountDir); err != nil {
+		return err
+	}
 	return nil
 }
+
+func newFlexVolumeClient() *flexVolumeClient {
+	deviceDriverMap = make(map[string]string)
```
peter-edge: setting a global variable in a constructor? why not just doing this when you declare deviceDriverMap?

api/server/flexvolume.go
```diff
@@ -1,27 +1,155 @@
...
+	return &flexVolumeClient{}
+}
+
+func (c *flexVolumeClient) getVolumeDriver(driverName string) (volume.VolumeDriver, error) {
+	driver, err := volume.Get(driverName)
+	if err != nil {
+		return nil, fmt.Errorf("%v driver not initialized in OSD", driverName)
+	}
+	return driver, nil
```
peter-edge: why not just return `volume.Get(driverName)`?

pkg/flexvolume/client.go
```diff
@@ -22,12 +34,22 @@ func (c *client) Init() error {
...
 		&AttachRequest{
 			JsonOptions: jsonOptions,
 		},
 	)
+	if err == nil {
+		output = []byte(fmt.Sprintf(`{"Status":"Success", "Device":"%s"}`, jsonOptions[volumeIDKey]))
+	} else {
+		output = newFailureBytes(err)
+	}
+	if _, osErr := os.Stdout.Write(output); osErr != nil {
```
peter-edge: can just call this `err`, it's scoped to the if statement

pkg/flexvolume/client.go
```diff
@@ -38,18 +60,33 @@ func (c *client) Detach(mountDevice string) error {
 			MountDevice: mountDevice,
 		},
 	)
+
+	output := newOutput(err)
+	if _, osErr := os.Stdout.Write(output); osErr != nil {
```
peter-edge: same thing, just `err`

pkg/flexvolume/client.go
```diff
@@ -22,12 +34,22 @@ func (c *client) Init() error {
...
+	var output []byte
+
 	_, err := c.apiClient.Attach(
 		context.Background(),
 		&AttachRequest{
 			JsonOptions: jsonOptions,
 		},
 	)
+	if err == nil {
+		output = []byte(fmt.Sprintf(`{"Status":"Success", "Device":"%s"}`, jsonOptions[volumeIDKey]))
```
peter-edge: just abstract this to a function like `newOutput`?

pkg/flexvolume/client.go
peter-edge: might as well just do:

```go
func writeOutput(err error) {
  return os.Stdout.Write(newOutput(err))
}```

and then instead of doing this logging below (which can be done by the caller), just do:

```go
return writeOutput(err)
```

pkg/flexvolume/client.go
```diff
@@ -38,18 +60,33 @@ func (c *client) Detach(mountDevice string) error {
...
+		_, err = c.apiClient.Mount(
+			context.Background(),
+			&MountRequest{
+				TargetMountDir: targetMountDir,
+				MountDevice:    mountDevice,
+				JsonOptions:    jsonOptions,
+			},
+		)
+	}
+	output := newOutput(err)
```
peter-edge: why not just:

```go
if err := os.MkdirAll(...); err != nil {
  return writeOutput(err)
}
_, err := c.apiClient.Mount(...)
return writeOutput(err)
```

api/server/flexvolume.go
adityadani: Returning an error would mean that the mount command failed in k8s flexvolume. But here it did succeed and our book-keeping effort failed. If this fails the only down side is our map will not be populated and in detach or unmount we might default to pwx driver.

peter-edge:
On the pwx default: the question I have is why the choice to default to pwx? Openstorage is meant to be driver-agnostic and not biased towards any given implementation, and pwx isn't even open source, so I would not default at all.

pkg/flexvolume/client.go
peter-edge: if you really want to log here (not sure), then you can do it inside the `writeOutput` function too

api/server/flexvolume.go
peter-edge: same comment about defaulting to pwx - I very much disagree with this :)

peter-edge: I really dislike silent failures - if there is some downside, then you can have some other function that handles this error in that manner and makes the side effect explicit, ie defaulting to pwx needs to be an explicit decision in the code that is just just a side effect of this behavior (and defaulting to pwx needs to be able to be changed, ie I don't agree with it, ha :) )

peter-edge:
LGTM

Issue #133, opened 2016-04-01
update api.SeverityType, api.ResourceType to match names of other enums


peter-edge:

adityadani:
Sent out a PR addressing this
https://github.com/libopenstorage/openstorage/pull/134

adityadani:
Merged PR.

Issue #130, opened 2016-04-01
rename Alerts struct to Alert, add relevant data to protobuf definitions


peter-edge:
Specifically the enums, also the package name alerts should be changed

adityadani:
Sent out a PR addressing this
https://github.com/libopenstorage/openstorage/pull/134

adityadani:
Merged PR

PR #134, opened 2016-04-03
Alert Package Cleanup

adityadani:
This PR addresses the following two issues:
https://github.com/libopenstorage/openstorage/issues/133
https://github.com/libopenstorage/openstorage/issues/130

- Renamed alerts package to alert
- Renamed Alerts structure in api.proto to Alert
- Renamed enums in proto file to follow the style


api/api.proto
```diff
@@ -220,6 +212,10 @@ message Alerts {
   bool cleared = 8;
 }
 
+message Alerts {
+  repeated Alert alerts = 1;
```
peter-edge: this should be `alert`, not `alerts`.

peter-edge:
I agree this fixes the enum naming, however two issues that I should have specified more clearly in the package bug:

1.

`AlertAction` should be a protobuf enum.

2.

```go
// Resource is equaivalent to api.ResourceType and is used in the alerts instance
// so that callers of the instance don't have to worry about api.*
type Resource int
```

Resource should be completely deleted. Right now, a second public type is literally added for typing convenience - I think that's all I should need to say on that, but if you need additional details on why we don't want to do that, ping me :)

Issue #136, opened 2016-04-11
runtime error: invalid memory address or nil pointer dereference


seterrychen:
When I using `docker volume ls`, it took longer time than expected and error message displayed.

```
vagrant@vagrant-ubuntu-trusty-64:~$ sudo time docker volume ls
list nfs: Post http://%2Frun%2Fdocker%2Fplugins%2Fnfs.sock/VolumeDriver.List: http: ContentLength=3 with Body length 0
DRIVER              VOLUME NAME
nfs                 test
0.02user 0.00system 0:15.04elapsed 0%CPU (0avgtext+0avgdata 15256maxresident)k
0inputs+0outputs (0major+4156minor)pagefaults 0swaps
```

and Output of osd:

```
2016/04/11 07:08:36 http: panic serving @: runtime error: invalid memory address or nil pointer dereference
goroutine 30 [running]:
net/http.(*conn).serve.func1(0xc8202129a0, 0x7f51835664a0, 0xc820026ae8)
	/usr/local/go/src/net/http/server.go:1287 +0xb5
github.com/libopenstorage/openstorage/volume.match(0xc8201e8840, 0x0, 0x0, 0xfc81e0)
	/home/vagrant/gopath/src/github.com/libopenstorage/openstorage/volume/enumerator.go:206 +0x1c
github.com/libopenstorage/openstorage/volume.(*DefaultEnumerator).Enumerate(0xc82016f500, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/vagrant/gopath/src/github.com/libopenstorage/openstorage/volume/enumerator.go:123 +0x2c3
github.com/libopenstorage/openstorage/api/server.(*driver).list(0xc82019ee40, 0x7f51835665f0, 0xc820212d10, 0xc8201f96c0)
	/home/vagrant/gopath/src/github.com/libopenstorage/openstorage/api/server/docker.go:314 +0x39f
github.com/libopenstorage/openstorage/api/server.(*driver).(github.com/libopenstorage/openstorage/api/server.list)-fm(0x7f51835665f0, 0xc820212d10, 0xc8201f96c0)
	/home/vagrant/gopath/src/github.com/libopenstorage/openstorage/api/server/docker.go:81 +0x3e
net/http.HandlerFunc.ServeHTTP(0xc8201ad7f0, 0x7f51835665f0, 0xc820212d10, 0xc8201f96c0)
[...]

PR #153, opened 2016-05-31
fix: nil pointer dereference when using "docker volume ls"

seterrychen:
Signed-off-by: Terry Chen <seterrychen@gmail.com>

When using "docker volume ls", the value of locator and volumeLables of `match` method will be nil. It's the root cause of the `nil pointer dereference` problem. (issue #136)

To fix the problem, elem is just added into volumes and `match` function doesn't need to check when locator and labels are nil.

### DIFF
--- a/volume/drivers/common/default_store_enumerator.go
+++ b/volume/drivers/common/default_store_enumerator.go
@@ -181,6 +181,9 @@ func match(
 	locator *api.VolumeLocator,
 	volumeLabels map[string]string,
 ) bool {
+	if locator == nil {
+		return hasSubset(v.Spec.VolumeLabels, volumeLabels)
+	}
 	if locator.Name != "" && v.Locator.Name != locator.Name {
 		return false
 	}


volume/drivers/common/default_store_enumerator.go
```diff
@@ -98,7 +98,8 @@ func (e *defaultStoreEnumerator) Enumerate(
 		if err := json.Unmarshal(v.Value, elem); err != nil {
 			return nil, err
 		}
-		if match(elem, locator, labels) {
+		if (locator == nil && labels == nil) ||
```
jvinod: Perhaps you could do this check within the match function itself?

seterrychen: Sure.

volume/drivers/common/default_store_enumerator.go
```diff
@@ -181,6 +181,9 @@ func match(
 	locator *api.VolumeLocator,
 	volumeLabels map[string]string,
 ) bool {
+	if locator == nil && volumeLabels == nil {
```
jvinod: it should be:

```
 if locator == nil {
return hasSubset(v.Spec.VolumeLabels, volumeLabels)
}
```

seterrychen: got it, thx

PR #154, opened 2016-05-31
Move the registration of drivers to another package for import purposes

peter-edge:
This is somewhat temporary until I think of a better solution


volume/drivers/all/all.go
```diff
@@ -0,0 +1,38 @@
+package all
```
jvinod: Maybe just call the package 'registry' ?

PR #156, opened 2016-07-15
K8s refactor

adityadani:
Refactoring subpackages
- api
- volume/drivers
  so that import of libopenstorage does not add unnecessary packages like libopenstorage/gossip and portworx/kvdb.


api/client/client.go
```diff
@@ -76,12 +78,15 @@ func NewDriverClient(driverName, version string) (*Client, error) {
 // standard unix domain socket
 func GetSupportedDriverVersions(driverName, serverPath string) ([]string, error) {
 	// Get a client handler
-	sockPath := "unix://" + config.DriverAPIBase + driverName + ".sock"
-	volumeClient, err := NewClient(sockPath, "")
+	if serverPath == "" {
```
sangleganesh: since "unix://" and "osd.sock" is repeated more than once, make them constants and create a function for getting server path ?

api/client/client.go
```diff
@@ -76,12 +78,15 @@ func NewDriverClient(driverName, version string) (*Client, error) {
...
+		serverPath = "unix://" + config.DriverAPIBase + driverName + ".sock"
+	}
+
+	client, err := NewClient(serverPath, "")
 	if err != nil {
 		return []string{}, err
 	}
-	versions, err := volumeClient.Versions()
+	versions, err := client.Versions("osd-volumes")
```
sangleganesh: make osd-volume some kind of constant ?

PR #168, opened 2016-12-11
Reduce the duration for which cluster lock is held

adityadani:
Do not hold the lock when Listener's Init functions
are being called.

Cluster And Node Init workflow

1. Take cluster lock
2. Read cluster data from kvdb. If cluster not initialized initialize
   the cluster and update cluster data in kvdb as initialized.
3. Release cluster lock
4. Call Listener's Init functions.
5. Take cluster lock
6. Update cluster data with new node info obtained from listener.
7. Release cluster lock
8 Start cluster watch based on last index obtained step 6.

### DIFF
--- a/cluster/manager.go
+++ b/cluster/manager.go
@@ -318,8 +318,6 @@ func (c *ClusterManager) getLatestNodeConfig(nodeId string) *NodeEntry {
 }
 
 func (c *ClusterManager) initNode(db *ClusterInfo) (*api.Node, bool) {
-	c.nodeCache[c.selfNode.Id] = *c.getCurrentState()
-
 	_, exists := db.NodeEntries[c.selfNode.Id]
 
 	// Add us into the database.
@@ -335,6 +333,7 @@ func (c *ClusterManager) initNode(db *ClusterInfo) (*api.Node, bool) {
 		Hostname:   c.selfNode.Hostname,
 		NodeLabels: labels,
 	}
+
 	db.NodeEntries[c.config.NodeId] = nodeEntry
 
 	dlog.Infof("Node %s joining cluster...", c.config.NodeId)
@@ -438,7 +437,7 @@ func (c *ClusterManager) joinCluster(
 		dlog.Panicln("Fatal, Unable to find self node entry in local cache")
 	}
 
-	err = c.updateNodeEntryDB(selfNodeEntry)
+	_, _, err = c.updateNodeEntryDB(selfNodeEntry)
 	if err != nil {
 		return err
 	}
@@ -477,6 +476,7 @@ func (c *ClusterManager) startHeartBeat(clusterInfo *ClusterInfo) {
 	gossipStoreKey := types.StoreKey(heartbeatKey + c.config.ClusterId)
 
 	node := c.getCurrentState()
+	c.nodeCache[c.selfNode.Id] = *node
 	c.gossip.UpdateSelf(gossipStoreKey, *node)
 	var nodeIps []string
 	for nodeId, nodeEntry := range clusterInfo.NodeEntries {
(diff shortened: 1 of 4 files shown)


cluster/manager.go
```diff
@@ -738,77 +720,97 @@ func (c *ClusterManager) startListeners(db kvdb.Kvdb, exist *bool) (
...
+	kvClusterInfo, err := readClusterInfo()
 	if err != nil {
 		dlog.Errorln("Failed to read cluster info. ", err)
-		return nil, nil, err
+		return -0, nil, err
 	}
-	clusterInfo.NodeEntries[c.config.NodeId] = selfNodeEntry
-	if clusterInfo.Status == api.Status_STATUS_INIT {
+	kvClusterInfo.NodeEntries[c.config.NodeId] = selfNodeEntry
+	if kvClusterInfo.Status == api.Status_STATUS_INIT {
```
sangleganesh: this is no longer true, you should assert that cluster state is OK.

cluster/manager.go
```diff
@@ -738,77 +720,97 @@ func (c *ClusterManager) startListeners(db kvdb.Kvdb, exist *bool) (
...
+		return nil, nodeExists, nil, errors.New("Fatal, Cluster is in an unexpected state.")
 	}
+	// Cluster database max size... 0 if unlimited.
+	c.size = clusterInfo.Size
+	c.status = api.Status_STATUS_OK
+	return &clusterInfo, nodeExists, self, nil
+}
 
-	selfNodeEntry, ok = initState.ClusterInfo.NodeEntries[c.config.NodeId]
+func (c *ClusterManager) startListeners(db kvdb.Kvdb, nodeExists *bool) (
```
sangleganesh: let us call this initListeners, to differentiate with actually starting listeners and to clarify what it is doing.

cluster/manager.go
```diff
@@ -738,77 +720,97 @@ func (c *ClusterManager) startListeners(db kvdb.Kvdb, exist *bool) (
...
+		dlog.Errorln("Failed to initialize node in cluster.", err)
+		return 0, nil, err
+	}
+
+	selfNodeEntry, ok := clusterInfo.NodeEntries[c.config.NodeId]
 	if !ok {
 		dlog.Panicln("Fatal, Unable to find self node entry in local cache")
 	}
+
+	kvlock, err := db.LockWithID(clusterLockKey, c.config.NodeId)
```
sangleganesh: use updateNodeDB to update this node's entry in DB.

cluster/manager.go
```diff
@@ -867,15 +869,15 @@ func (c *ClusterManager) Start() error {
 	// Start heartbeating to other nodes.
 	go c.startHeartBeat(clusterInfo)
```
sangleganesh: After this point cluster info becomes stale, to disallow others from using it, probably we should move start heart beat and start listeners into a single function.

cluster/manager.go
```diff
@@ -867,15 +869,15 @@ func (c *ClusterManager) Start() error {
...
+	c.startClusterDBWatch(lastIndex, kvdb)
 
 	err = c.waitForQuorum(exist)
 	if err != nil {
 		return err
 	}
 
 	go c.updateClusterStatus()
-	go c.replayNodeDecommission(initState)
+	go c.replayNodeDecommission(clusterInfo)
```
sangleganesh: cluster info is stale, remove it as a param here and instead do readClusterInfo() as a part of replayNodeDecommision (or i can do it as a part of my clean up too).

cluster/manager.go
```diff
@@ -738,77 +720,97 @@ func (c *ClusterManager) startListeners(db kvdb.Kvdb, exist *bool) (
...
-		return nil, nil, errors.New(msg)
+		return nil, true, nil, errors.New(msg)
 	}
-
-	// Cluster database max size... 0 if unlimited.
-	c.size = initState.ClusterInfo.Size
 	// Set the clusterID in db
-	initState.ClusterInfo.Id = c.config.ClusterId
+	clusterInfo.Id = c.config.ClusterId
+	self, nodeExists := c.initNode(&clusterInfo)
```
sangleganesh: initNode calls getCurrentState, which was fine until we added call in getCurrentState to get listener states. Listeners are not yet initialized - so I suggest removing getCurrentState from initNode.

PR #169, opened 2017-02-09
A discovery interface with a kvdb implementation for Cluster Manager

adityadani:


cluster/discovery/discovery.go
```diff
@@ -0,0 +1,33 @@
...
+
+	// RemoveNode removes a node from a cluster
+	RemoveNode(dne NodeEntry) (*ClusterInfo, error)
+
+	// Enumerate enumerates the nodes that have been discovered in the cluster
+	Enumerate() (*ClusterInfo, error)
+
+	// WatchCluster starts a watch on the cluster and calls the provided
+	// callback function when a node is added or removed.
+	WatchCluster(wcb WatchClusterCB, index uint64) error
```
sangleganesh: nit: remove "Cluster" from Watch

cluster/manager.go
```diff
@@ -504,33 +520,32 @@ func (c *ClusterManager) startClusterDBWatch(lastIndex uint64,
...
-		labels := nodeEntry.NodeLabels
-		version, ok := labels[gossipVersionKey]
-		if !ok || version != c.gossipVersion {
+		if nodeEntry.GossipVersion != c.gossipVersion {
 			// Do not add nodes with mismatched version
 			continue
 		}
 
-		nodeIps = append(nodeIps, nodeEntry.DataIp+":9002")
+		nodeIps = append(nodeIps, nodeEntry.Ip+":9002")
```
sangleganesh: make "9002" a constant

cluster/discovery/discovery.go
```diff
@@ -0,0 +1,33 @@
+package discovery
+
+// NodeEntry is used to discovery nodes in the cluster
+type NodeEntry struct {
+	Id            string
+	Ip            string
+	GossipVersion string
```
sangleganesh: GossipVersion seems to be out of place here ? Also, isn't the version actually passed over wire ? (not sure why we need to put it here ..)

cluster/manager.go
```diff
@@ -248,6 +250,20 @@ func (c *ClusterManager) getNonDecommisionedPeers(db ClusterInfo) map[types.Node
 	return peers
 }
 
+func (c *ClusterManager) getBootstrapPeers(dci *discovery.ClusterInfo) map[types.NodeId]string {
+	peers := make(map[types.NodeId]string)
+	for _, bne := range dci.Nodes {
+		peers[types.NodeId(bne.Id)] = bne.Ip + ":9002"
```
sangleganesh: nit: "9002" can be a constant. Not sure why you need to add the constant - seems like if all of them have same port then we probably don't need to add it at all .. ?

cluster/manager.go
```diff
@@ -1394,6 +1433,15 @@ func (c *ClusterManager) NodeRemoveDone(nodeID string, result error) {
 
 	err := c.deleteNodeFromDB(nodeID)
 	if err != nil {
+		goto error_done
+	}
+	_, err = c.discoverService.RemoveNode(discovery.NodeEntry{Id: nodeID})
```
sangleganesh: what if we crash before removing it from discoverService but it is removed from all other dbs ?
(This is a third db that needs to be kept in sync with the /storage and /cluster ).

cluster/manager.go
```diff
@@ -958,9 +956,47 @@ func (c *ClusterManager) Start(
 	)
 	c.gossipVersion = types.GOSSIP_VERSION_2
 
+	dne := discovery.NodeEntry{
+		Id:            c.config.NodeId,
+		Ip:            c.selfNode.DataIp,
+		GossipVersion: types.GOSSIP_VERSION_2,
+	}
+	clusterInfo, err := c.discoverService.AddNode(dne)
```
sangleganesh: I can't understand the flow here - but does it happen after NodeInit is called ? 
If not, what happens if NodeInit fails ? 
I don't see cleanup code .. ?

cluster/cluster.go
```diff
@@ -163,6 +164,12 @@ type ClusterRemove interface {
 
 // Cluster is the API that a cluster provider will implement.
 type Cluster interface {
+
+	// DiscoveryStart has to be called before Start and needs to be provided with
+	// a discovery implementation that could be used to bootstrap a node in the
+	// cluster
+	DiscoveryStart(discovery.Cluster) error
```
sangleganesh: This requirement seems awkward. Why not roll it up as a part of Start and make start accept the "discovery" thing ?

cluster/discovery/discovery.go
adityadani: Currently we look at this GossipVersion and then decide if we should add it in our gossip map. Basically the list of Ips that we pass to gossip are decided based on gossip version matches.

cluster/cluster.go
adityadani: I wanted to separate out the two things
- starting discovery
- starting the cluster manager and cluster state machine.

cluster/discovery/discovery.go
adityadani: Currently we check gossip version before we pass in the list of Ips to gossip.Start.

cluster/discovery/discovery.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+// ClusterInfo is the cluster info used while discoveryping nodes
+// and discovering peer nodes using gossip
+type ClusterInfo struct {
+	Size    int
```
adityadani: yes will remove it.

cluster/discovery/discovery.go
```diff
@@ -0,0 +1,33 @@
+package discovery
+
+// NodeEntry is used to discovery nodes in the cluster
+type NodeEntry struct {
+	Id            string
+	Ip            string
+	GossipVersion string
+}
+
+// ClusterInfo is the cluster info used while discoveryping nodes
```
adityadani: It is an incorrect description. Will change it

cluster/discovery/discovery.go
```diff
@@ -0,0 +1,33 @@
...
+// and discovering peer nodes using gossip
+type ClusterInfo struct {
+	Size    int
+	Nodes   map[string]NodeEntry
+	Version uint64
```
adityadani: In case of a kvdb discovery service it will be KvdbVersion. I will add a comment.

cluster/discovery/discovery.go
adityadani: will remove.

cluster/discovery/discovery.go
```diff
@@ -0,0 +1,33 @@
...
+	// AddNode adds a new node into a cluster so that other can discover
+	AddNode(dne NodeEntry) (*ClusterInfo, error)
+
+	// RemoveNode removes a node from a cluster
+	RemoveNode(dne NodeEntry) (*ClusterInfo, error)
+
+	// Enumerate enumerates the nodes that have been discovered in the cluster
+	Enumerate() (*ClusterInfo, error)
+
+	// WatchCluster starts a watch on the cluster and calls the provided
```
adityadani: We need some version to start watching from. Let me see if I can remove it from this API and in some way make it only kvdb implementation specific.

cluster/discovery/discovery_kvdb.go
```diff
@@ -0,0 +1,113 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/portworx/kvdb"
+	"go.pedge.io/dlog"
+)
+
+const (
+	// ClusterDiscoveryKey is the key at which discovery info is stored in kvdb
+	ClusterDiscoveryKey = "cluster/discoverydb"
```
adityadani: Will change this.

cluster/discovery/discovery_kvdb.go
```diff
@@ -0,0 +1,113 @@
...
+	if ci.Size == 0 {
+		ci.Nodes = make(map[string]NodeEntry)
+	}
+
+	if _, exists := ci.Nodes[ne.Id]; !exists {
+		// We do not exist in the map.
+		ci.Size = ci.Size + 1
+	}
+	ci.Nodes[ne.Id] = ne
+	kvp, err := b.kv.Put(ClusterDiscoveryKey, &ci, 0)
```
adityadani: Will use CaS

cluster/manager.go
adityadani: NodeInit is called after this routine. This one only adds self node into the discovery db and tries to connect with other node in gossip.

Yes I need to add some clean up code if NodeInit fails.

adityadani:
Closing this PR as it is not required anymore.

PR #172, opened 2017-02-16
167: Fixing AWS Create volume

zoxpx:
AWS Create Volume was allocating device names, but it was never releasing them.
As a result, no more than 11 AWS volumes could be created.


volume/drivers/aws/aws.go
```diff
@@ -603,6 +610,13 @@ func (d *Driver) Detach(volumeID string) error {
 
 		return err
 	}
+
+	if "" != device {
```
adityadani: nit: For consistency - if  device != ""

PR #173, opened 2017-03-07
PWX-2772: Implemented REST cluster/inspect/node

zoxpx:
Old REST "GET Inspect Cluster Node" call was not implemented.
This fix implements the call, and is loosely based on REST "GET
Enumerate" call.


zoxpx:
Merging into 'master' as verbally agreed w/ Ganesh (will do a post-review, and re-fix if required).

Issue #177, opened 2017-03-15
Cluster.Start() ignores persisted MaxClusterNodes


zoxpx:
The `Cluster.Start(maxClusterNodes int, nodeInitialized bool)` API call always takes the "maximum cluster nodes" parameter, and it ignores the kvdb-persisted "maximum nodes" previously set by `Cluster.SetSize()`.

We should introduce a distinct `maxClusterNodes` parameter value, such that the `Cluster.Start()` would take the persisted "max cluster nodes" value instead of overriding it.

sangleganesh:
I am not sure if this is an issue at all - the Cluster.SetSize() afaik never worked, and that value is not used anywhere. Also, not sure what this issue is tracking ? The current mechanism works fine in controlling the size of the cluster, and also allows to seamlessly increase the size when started with a new size. Can you clarify what is the issue here and what are its effects ?

zoxpx:
The `Cluster.SetSize(#nodes)` does actually persist the value (see [ref](https://github.com/libopenstorage/openstorage/blob/master/cluster/manager.go#L1222)), but it was not connected in any way to the `Cluster.Start(#nodes)` - which didn't seem right.
Note that there is still a discrepancy when one defines number of nodes via `Cluster.SetSize()` smaller than defined by `Cluster.Start()`, which could cause all the nodes in the cluster to begin restarting (see [ref](https://github.com/libopenstorage/openstorage/blob/master/cluster/manager.go#L623)).

sangleganesh:
Thats exactly what I said, that field is not used because it did not work before. IMO the issue should be to remove that unused field.

zoxpx:
Agree w/ previous -- the `SetSize()` is advisory number, required at start to set up the initial cluster size.  Based on this number, the cluster will auto-eject itself from the cluster if it finds its over the count.
The node restart issue via gossip count-check may still need to be addressed, but should be in a different tracker (technically, it's a different issue).

Closing this tracker.

PR #178, opened 2017-03-15
Cluster.Start() and MaxClusterNodes parameter fix (#177)

zoxpx:
Introduced special "maxNodes" value 0, which will cause Cluster.Start() to use
persisted "maxNodes" instead of overriding.


zoxpx:
Changed sequencing in PX code -- this change no longer necessary (closing MR)

PR #183, opened 2017-04-21
Add REST client and server for Alerts

adityadani:
Add REST APIs to handle alerts
- Enumerate
- Clear
- Erase

These APIs can be called through the volume or cluster routes.

### DIFF
--- a/api/client/cluster/client.go
+++ b/api/client/cluster/client.go
@@ -167,26 +167,18 @@ func (c *clusterClient) GetGossipState() *cluster.ClusterState {
 	return status
 }
 
-func (c *clusterClient) EnumerateAlerts(resource api.ResourceType) (*api.Alerts, error) {
+func (c *clusterClient) EnumerateAlerts(ts, te time.Time, resource api.ResourceType) (*api.Alerts, error) {
 	a := api.Alerts{}
 	request := c.c.Get().Resource(clusterPath+"/alerts/" + strconv.FormatInt(int64(resource), 10))
+	if !te.IsZero() {
+		request.QueryOption("timestart", ts.Format(api.TimeLayout))
+		request.QueryOption("timeend", te.Format(api.TimeLayout))
+	}
 	if err := request.Do().Unmarshal(&a); err != nil {
 		return nil, err
 	}
 	return &a, nil
 }
-
-func (c *clusterClient) EnumerateAlertsWithinTimeRange(ts, te time.Time, resource api.ResourceType) (*api.Alerts, error) {
-	a := api.Alerts{}
-	request := c.c.Get().Resource(clusterPath+"/alerts/" + strconv.FormatInt(int64(resource), 10))
-	request.QueryOption("timestart", ts.Format(api.TimeLayout))
-	request.QueryOption("timeend", te.Format(api.TimeLayout))
-	if err := request.Do().Unmarshal(&a); err != nil {
-		return nil, err
-	}
-	return &a, nil
-}
-
 
 func (c *clusterClient) ClearAlert(resource api.ResourceType, alertID int64) error {
 	path := clusterPath + "/alerts/" + strconv.FormatInt(int64(resource), 10) + "/" + strconv.FormatInt(alertID, 10)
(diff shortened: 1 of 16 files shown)


api/client/cluster/client.go
```diff
@@ -146,3 +147,44 @@ func (c *clusterClient) GetGossipState() *cluster.ClusterState {
 	}
 	return status
 }
+
+func (c *clusterClient) EnumerateAlerts(resource api.ResourceType) (*api.Alerts, error) {
```
sangleganesh: seems like there can be only one set of alert routes and pass resource type appropriate for volume or cluster alerts .. the code looks exactly the same except clusterPath and volumePath ?

api/server/cluster.go
```diff
@@ -29,6 +31,9 @@ func (c *clusterApi) Routes() []*Route {
 		{verb: "PUT", path: clusterPath("/disablegossip", cluster.APIVersion), fn: c.disableGossip},
 		{verb: "PUT", path: clusterPath("/shutdown", cluster.APIVersion), fn: c.shutdown},
 		{verb: "PUT", path: clusterPath("/shutdown/{id}", cluster.APIVersion), fn: c.shutdown},
+		{verb: "GET", path: clusterPath("/alerts/{resource}", cluster.APIVersion), fn: c.enumerateAlerts},
```
sangleganesh: can you pass resource type volume here to clear volume alerts ?

volume/drivers/buse/buse.go
```diff
@@ -109,12 +115,13 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
...
 	}
 
+	inst.cl = &clusterListener{}
 	c, err := cluster.Inst()
 	if err != nil {
 		dlog.Println("BUSE initializing in single node mode")
 	} else {
 		dlog.Println("BUSE initializing in clustered mode")
-		c.AddEventListener(inst)
+		c.AddEventListener(inst.cl)
```
adityadani: Yes I was not sure why it was added in the first place. I can remove it completely.

api/client/cluster/client.go
adityadani: Sure I can do that. I added it because if any driver implements only the volume interface and not the clustered osd interface, they should still be able to get alerts for only volumes.

cluster/manager.go
```diff
@@ -1542,6 +1542,64 @@ func (c *ClusterManager) HandleNotifications(culpritNodeId string, notification
...
+			dlog.Warnf("Failed to enumerate alerts from (%v): %v",
+				e.Value.(ClusterListener).String(), err)
+			continue
+		}
+		a.Alert = append(a.Alert, listenerAlerts.Alert...)
+	}
+	return &a, nil
+}
+
+func (c *ClusterManager) EnumerateAlertsWithinTimeRange(ts, te time.Time, resource api.ResourceType) (*api.Alerts, error) {
```
adityadani: Ok I will just have one API which takes in two time arguments

api/server/cluster.go
```diff
@@ -248,6 +253,139 @@ func (c *clusterApi) versions(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	err = inst.ClearAlert(resourceType, alertId)
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	msg := "Successfully cleared Alert "
+	json.NewEncoder(w).Encode(msg)
```
sangleganesh: nit: no need for msg var can just pass message to Encode..

api/server/cluster.go
```diff
@@ -248,6 +253,139 @@ func (c *clusterApi) versions(w http.ResponseWriter, r *http.Request) {
...
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	err = inst.EraseAlert(resourceType, alertId)
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	msg := "Successfully erased Alert "
```
sangleganesh: nit: msg not needed

PR #187, opened 2017-09-06
Deprecate protoeasy

adityadani:
- Removes usage of protoeasy from openstorage
- Generated protobuf definitions now use github.com/golang/protobuf instead of go.pedge.io/protobuf
- Removed all references of go.pedge.io/protobuf from other go files
- Build grpc and protobuf for flexvolumes without using protoeasy


pkg/flexvolume/api_server.go
```diff
@@ -3,48 +3,58 @@ package flexvolume
...
 }
 
 func newAPIServer(client Client) *apiServer {
-	return &apiServer{protorpclog.NewLogger("flexvolume.API"), client}
+	return &apiServer{client}
 }
 
-func (a *apiServer) Init(_ context.Context, _ *google_protobuf.Empty) (_ *google_protobuf.Empty, err error) {
-	defer func(start time.Time) { a.Log(nil, nil, err, time.Since(start)) }(time.Now())
+func (a *apiServer) Init(_ context.Context, _ *empty.Empty) (_ *empty.Empty, err error) {
```
sangleganesh: is this going to be backward compatible ?

api/api.pb.go
```diff
@@ -626,17 +658,17 @@ func (m *StorageResource) GetLastScan() *google_protobuf.Timestamp {
 // StoragePool groups different storage devices based on their CosType
 type StoragePool struct {
 	// ID pool ID
-	ID int32 `protobuf:"varint,1,opt,name=ID" json:"ID,omitempty"`
+	ID int32 `protobuf:"varint,1,opt,name=ID,json=iD" json:"ID,omitempty"`
```
sangleganesh: same question - is this backward compatible ?

pkg/flexvolume/api_server.go
adityadani: Yes this should be. It just changes the package from which Empty struct was being pulled from go.pedge.io/protobuf --> github.com/golang/protobuf

api/api.pb.go
adityadani: Adding a new tag would not break backward compatibility.

PR #191, opened 2017-10-05
Pre mount state

talakad:
PWX-3974: Revert immutable flag, only it was not set to begin with on mount failure.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -88,6 +88,7 @@ const (
 
 const mountPathRemoveDelay = 30 * time.Second
 const chattrBin = "/usr/bin/chattr"
+const lsattrBin = "/usr/bin/lsattr"
 
 var (
 	// ErrExist is returned if path is already mounted to a different device.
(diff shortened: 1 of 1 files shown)


pkg/mount/mount.go
```diff
@@ -543,6 +548,33 @@ func (m *Mounter) removeSoftlinkAndTarget(link string) error {
...
+		dlog.Errorf("Failed to stat mount path:%v", err)
+		return true
+	}
+	op, err := exec.Command(lsattrBin, "-d", mountpath).CombinedOutput()
+	if err != nil {
+		// Cannot get path status, return true so that immutable bit is not reverted
+		dlog.Errorf("error listing attrs for %v err:%v", mountpath, string(op))
+		return true
+	}
+	attrs := strings.Split(string(op), " ")
```
disrani-px: Can you add a comment on what the output should look like, so that we know what we are trying to parse.

pkg/mount/mount.go
```diff
@@ -543,6 +548,33 @@ func (m *Mounter) removeSoftlinkAndTarget(link string) error {
...
+// is immutable .
+func (m *Mounter) isPathSetImmutable(mountpath string) bool {
+	if _, err := os.Stat(mountpath); err != nil {
+		dlog.Errorf("Failed to stat mount path:%v", err)
+		return true
+	}
+	op, err := exec.Command(lsattrBin, "-d", mountpath).CombinedOutput()
+	if err != nil {
+		// Cannot get path status, return true so that immutable bit is not reverted
+		dlog.Errorf("error listing attrs for %v err:%v", mountpath, string(op))
```
disrani-px: Start message with capital letter

pkg/mount/mount.go
```diff
@@ -543,6 +548,33 @@ func (m *Mounter) removeSoftlinkAndTarget(link string) error {
...
+		return true
+	}
+	attrs := strings.Split(string(op), " ")
+	if len(attrs) < 2 {
+		// Cannot get path status, return true so that immutable bit is not reverted
+		dlog.Errorf("Invalid lsattr output %v", string(op))
+		return true
+	}
+	if strings.Contains(attrs[0], "i") {
+		dlog.Errorf("Mount Path already set to Immutable")
```
disrani-px: Should probably just be a warning, this function is just checking the state, it isn't an error for this function.

talakad: it isn't error. But Not expected. its good leave that log.

pkg/mount/mount.go
talakad: yes. good to have the sample output.

PR #194, opened 2017-10-25
Initial work for CSI Support

### LINKED ISSUE #201: csi: Add documentation to API
Add documentation to API

lpabon:
Closes #201


lpabon:
@adityadani @jvinod Please review when you can.

Is travis running on this?

csi/csi.go
```diff
@@ -0,0 +1,76 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package csi
+
+import (
+	"net"
+	"sync"
+
+	//"github.com/libopenstorage/openstorage/api/spec"
```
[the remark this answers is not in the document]
jvinod: Add golint target for csi in Makefile, eventually everything needs to pass golint, but lets alteast make new files go through it first.

```
LINT_TARGETS = csi
lint:                                                                                                                                                        │····················
        go get -v github.com/golang/lint/golint                                                                                                              │····················
        for pkg in $(LINT_TARGETS); do \                                                                                                                     │····················
                for file in $$(find $${pkg} -name '*.go' | grep -v vendor | grep -v '\.pb\.go' | grep -v '\.pb\.gw\.go'); do \                               │····················
                        $(GOPATH)/bin/golint $${file}; \                                                                                                     │····················
                        if [ -n "$$($(GOPATH)/bin/golint $${file})" ]; then \                                                                                │····················
                                exit 1; \                                                                                                                    │····················
                        fi; \                                                                                                                                │····················
                done; \                                                                                                                                      │····················
        done
```

tests/assert.go
```diff
@@ -0,0 +1,47 @@
+//
```
jvinod: Do we need this? 
see "github.com/stretchr/testify/require"
"github.com/stretchr/testify/assert"
https://github.com/stretchr/testify

csi/csi.go
```diff
@@ -0,0 +1,76 @@
...
+	csi "github.com/container-storage-interface/spec/lib/go/csi"
```
adityadani: The alias should not be required as the package itself is named csi

csi/identity.go
```diff
@@ -0,0 +1,64 @@
...
+var (
+	csiVersion = &csi.Version{
+		Major: 0,
+		Minor: 0,
+		Patch: 0,
+	}
+)
+
+func (s *OsdCsiServer) GetSupportedVersions(
+	context.Context, *csi.GetSupportedVersionsRequest) (*csi.GetSupportedVersionsResponse, error) {
```
adityadani: We try to follow the following pattern for functions in openstorage,
func (s *receiver) foo(
arg1,
arg2,
) (return values) {
}

csi/identity.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (s *OsdCsiServer) GetPluginInfo(
+	context.Context, *csi.GetPluginInfoRequest) (*csi.GetPluginInfoResponse, error) {
+	return &csi.GetPluginInfoResponse{
+		Reply: &csi.GetPluginInfoResponse_Result_{
+			Result: &csi.GetPluginInfoResponse_Result{
+				Name:          csiDriverName,
+				VendorVersion: csiDriverVersion,
+				Manifest: map[string]string{
+					"hello":                  "world",
```
adityadani: What would be the real data here in this map?

csi/csi.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"go.pedge.io/dlog"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+type OsdCsiServerConfig struct {
+	Listener net.Listener
```
adityadani: If this is a config object shouldn't it just have ip and port number instead of a Listener object? This server can then create a listener object.

csi/identity.go
lpabon: Don't know yet. Most likely be `nil` for the moment.

csi/csi.go
lpabon: By providing "Listener" the caller can create either a TCP or Unix Domain socket connection.

csi/csi.go
lpabon: Yeah, this is carry over from the generated code. I'll take care of it.

tests/assert.go
lpabon: Ok, I will use that.

csi/csi.go
```diff
@@ -0,0 +1,76 @@
...
+
+	//"github.com/libopenstorage/openstorage/api/spec"
+	csi "github.com/container-storage-interface/spec/lib/go/csi"
+
+	"go.pedge.io/dlog"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+type OsdCsiServerConfig struct {
```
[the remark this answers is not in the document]
lpabon: It is only for that server, so I would assume so.

csi/csi.go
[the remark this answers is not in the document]
lpabon: Cool, I'll add this.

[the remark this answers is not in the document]
lpabon: @jvinod I think you can replace this shell script with the command:
` golint $(go list ./... | grep -v vendor)`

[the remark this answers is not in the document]
lpabon: Solved by #195

lpabon:
@adityadani @jvinod Please take a look at the new updates

csi/csi.go
disrani-px: Then maybe also have helper function like NewOcdCsiTcpServer() and NewOsdCsiSocketServer() which will create the listeners with sane defaults?

csi/csi.go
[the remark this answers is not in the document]
disrani-px: Since the package is called csi, it would be used as csi.OsdCsiServerConfig. "go lint" suggests to use just ServerConfig in these cases to avoid stuttering

csi/csi.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (s *OsdCsiServer) Address() string {
+	return s.listener.Addr().String()
+}
+
+func (s *OsdCsiServer) goServe() {
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+		s.server.Serve(s.listener)
```
disrani-px: Check return code for Serve()? Document says that it'll only return fatal errors, so would be good to log that here.

csi/csi_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+// csiTestService is a simple struct used abstract
+// the creation and setup of the gRPC CSI service
+type csiTestService struct {
```
disrani-px: Same comment as above. This is within the csi package, so doesn't really need a qualifier for the struct names. If this was public ie CsiTestService, "go lint" would have complained.

lpabon: Cool, since this one is not exported, I'll change the name.

csi/csi.go
lpabon: Good point. I will take care of it.

csi/csi.go
```diff
@@ -0,0 +1,75 @@
...
+
+type OsdCsiServer struct {
+	listener net.Listener
+	server   *grpc.Server
+	wg       sync.WaitGroup
+}
+
+func NewOsdCsiServer(config *OsdCsiServerConfig) *OsdCsiServer {
+	return &OsdCsiServer{
+		listener: config.Listener,
```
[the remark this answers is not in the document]
sangleganesh: also, missing information/checks on the state of Listener.

csi/csi_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+)
+
+// testServer is a simple struct used abstract
```
sangleganesh: missing negative tests.

csi/csi.go
```diff
@@ -0,0 +1,75 @@
...
+	reflection.Register(s.server)
+
+	// Start listening for requests
+	dlog.Infof("CSI Server ready on %s", s.Address())
+	s.goServe()
+	return nil
+}
+
+func (s *OsdCsiServer) Stop() {
+	s.server.Stop()
```
sangleganesh: calling Stop() multiple times will panic, in the absence of an interface and function documentation it is unclear what is the expectation from clients.

csi/csi.go
```diff
@@ -0,0 +1,75 @@
...
+	return &OsdCsiServer{
+		listener: config.Listener,
+	}
+}
+
+func (s *OsdCsiServer) Start() error {
+	s.server = grpc.NewServer()
+
+	csi.RegisterIdentityServer(s.server, s)
+	reflection.Register(s.server)
```
sangleganesh: can you add comments on why reflection is being used ?

csi/csi.go
sangleganesh: agree with the comments above - passing Listener seems a error prone as well.

csi/csi_test.go
lpabon: These are coming in future PRs. I'll add an issue for this to track it.

csi/csi.go
lpabon: True, documentation will be added in the next PR. I'll add an issue for this

csi/csi.go
lpabon: Will add comment. I just followed the example, but I think it is used to report to gRPC callers which functions are available.

csi/csi.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func NewOsdCsiServer(config *OsdCsiServerConfig) *OsdCsiServer {
+	return &OsdCsiServer{
+		listener: config.Listener,
+	}
+}
+
+func (s *OsdCsiServer) Start() error {
+	s.server = grpc.NewServer()
```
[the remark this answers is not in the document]
lpabon: Good catch. I will fix this.

csi/csi.go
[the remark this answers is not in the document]
lpabon: Cannot check for nil in an interface. Maybe what I can do is ask for the parameters that are normally passed to Listen().

csi/csi_test.go
lpabon: Will fix in #200

csi/csi.go
lpabon: Will be fixed in #201

csi/csi.go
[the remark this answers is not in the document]
lpabon: Added locks and state of server.

csi/csi.go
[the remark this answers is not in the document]
lpabon: Fixed

PR #195, opened 2017-10-26
travis: Update Makefile and Travis.yml

lpabon:
Now Travis will test for fmt and vet of files. This WILL make
it fail when this patch is introduced (which is good) because
it will catch many items in the code that need to be fixed.


lpabon:
This failure in Travis means the patch is working 👍

adityadani:
Is it possible to do this per package?
This way we don't keep the build failing and fix every package incrementally.

lpabon:
I think we should fix it all. I have a commit coming with the changes needed to make this work.

lpabon:
This fixes four issues found by vet on the code

lpabon:
Nice! All works and fixed.

PR #203, opened 2017-11-01
Add SetTLS function to client so we can support https

pault84:
Hey we need this functionality so we can support https and certificates in our client.

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -87,6 +87,12 @@ type Client struct {
 	userAgent   string
 }
 
+func (c * Client) SetTLS(tlsConfig *tls.Config) {
+	c.httpClient = &http.Client{
+		Transport: &http.Transport{TLSClientConfig: tlsConfig},
+	}
+}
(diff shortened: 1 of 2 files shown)


lpabon:
Could you please provide a unit test function utilizing this new API?

lpabon:
Looks like go fmt issues: https://travis-ci.org/libopenstorage/openstorage/jobs/295951496#L540

pault84:
It changes the client indeed, if we change the actual client code there will be about 10x the amount of code changes required. I can fix the go fmt issues.

lpabon:
@pault84 Ok, Then we probably want some kind of check or test to make sure that once the server is running, this new call returns an error. Or is the expectation that making this call will restart the server at any point?

pault84:
I got your point, i'm going to sit with vinod today and hammer some things out and will do an update on this pr.

lpabon:
Looks good!

PR #207, opened 2017-11-03
csi: Controller Server initial support

lpabon:
This patch adds support for the following CSI interfaces in the
Controller Server:
    
* ValidateVolumeCapabilities
* ControllerPublishVolume
* ControllerUnpublishVolume
* ControllerGetCapabilities
    
It also adds support for the NewOsdCsiServer to support being
provided a driver name to use.
    
Lastly, this patch also introduces the use of Golang Mocks,
and provides a new file called volume/drivers/mock/driver.mock.go
    which can be automatically generated using the command:
    
    ```
    $ cd volume/drivers ; go generate ./...
    ```


lpabon:
Please take note of the new API added to volume/driver: https://github.com/libopenstorage/openstorage/pull/207/commits/4f6a24708dae4ed9b4a63b896c2b7ad3e1ba62f3#diff-24221b5a5a3de1557d5407304cd715ebR217

`Remove()` was necessary for unit tests since the variable which holds the registration is global.

csi/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	"golang.org/x/net/context"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// ControllerGetCapabilities is a CSI API functions which returns to the caller
+// the capabilities of the OSD CSI driver.
+func (s *OsdCsiServer) ControllerGetCapabilities(
+	ctx context.Context,
+	req *csi.ControllerGetCapabilitiesRequest) (*csi.ControllerGetCapabilitiesResponse, error) {
```
adityadani: Can you put all the function prototypes in the following format
func (s *osd) funcName(
arg1 string,
arg2 string,
) (return values)

We try to follow this [styleguide](https://github.com/libopenstorage/openstorage/blob/master/STYLEGUIDE.md)

csi/controller.go
```diff
@@ -0,0 +1,230 @@
...
+	}, nil
+
+}
+
+// ControllerPublishVolume is a CSI API implements the attachment of a volume
+// on to a node.
+func (s *OsdCsiServer) ControllerPublishVolume(
+	context.Context,
+	*csi.ControllerPublishVolumeRequest) (*csi.ControllerPublishVolumeResponse, error) {
+	return nil, status.Error(codes.Unimplemented, "This request is not supported")
```
adityadani: Is the CSI client expecting this error code as the error for No-op?

csi/controller.go
```diff
@@ -0,0 +1,230 @@
...
+				"Cannot have both mount and block be undefined")
+		}
+
+		// Check access mode is setup correctly
+		mode := capability.GetAccessMode()
+		switch {
+		case mode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER:
+			if v.Spec.Shared {
+				result.Supported = false
+				result.Message = "Volume is a multi node volume"
```
adityadani: Could define all the messages as constants.

csi/controller.go
```diff
@@ -0,0 +1,230 @@
...
+				break
+			}
+		case mode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY:
+			if v.Spec.Shared {
+				result.Supported = false
+				result.Message = "Volume is a multi node volume"
+				break
+			}
+			if !v.Readonly {
+				result.Supported = false
```
adityadani: @jvinod This could be a separate discussion, but do we want a writable volume to be provided as a readyOnly volume to a container orchestrator? Some drivers will not have a notion of readOnly volumes.

However if we allow, then we defer the control to CO then it might schedule apps which perform write operation on such a volume.

csi/controller.go
```diff
@@ -0,0 +1,230 @@
...
+				break
+			}
+		case mode.Mode == csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY:
+			if !v.Spec.Shared {
+				result.Supported = false
+				result.Message = "Volume is not a multi node volume"
+				break
+			}
+			if !v.Readonly {
+				result.Supported = false
```
adityadani: same concern as single node read only volume

csi/controller.go
```diff
@@ -0,0 +1,230 @@
...
+				result.Message = "Volume is not a multi node volume"
+				break
+			}
+			if v.Readonly {
+				result.Supported = false
+				result.Message = "Volume is read only"
+				break
+			}
+		default:
```
adityadani: IMO the default case should also return the result.Supported as false, instead of error.

lpabon: default is a coding error on the client and a spec contract break

csi/controller.go
lpabon: Ah! I forgot the last return. Yeah, i'll make these changes.

csi/controller.go
lpabon: This is for clients that are incorrectly written

csi/controller.go
lpabon: Good point.

csi/controller.go
lpabon: Created issue #211 to track this

csi/controller.go
lpabon: On issue #211

csi/controller.go
adityadani: discussed offline with @lpabon

PR #208, opened 2017-11-03
csi: Create an interface for the gRPC server

### LINKED ISSUE #206: Create an interface for CSI server
The CSI server function `NewOsdCsiServer` should return an interface probably called `CsiServer` which includes all the interfaces in the CSI spec plus `Start()` and `Stop()`.

lpabon:
Closes #206

### DIFF
--- a/csi/csi.go
+++ b/csi/csi.go
@@ -37,6 +37,7 @@ type OsdCsiServerConfig struct {
 // OsdCsiServer is a OSD CSI compliant server which
 // proxies CSI requests for a single specific driver
 type OsdCsiServer struct {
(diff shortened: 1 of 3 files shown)


csi/csi.go
```diff
@@ -37,6 +37,7 @@ type OsdCsiServerConfig struct {
 // OsdCsiServer is a OSD CSI compliant server which
 // proxies CSI requests for a single specific driver
 type OsdCsiServer struct {
+	Server
```
adityadani: Just curious, why do we need this? It should work without adding the nested interface as well.

PR #213, opened 2017-11-10
Extracting out method to parse name string in an options map as well.

piyush-nimbalkar:


lpabon:
@piyush-nimbalkar please add unit tests next time.

PR #220, opened 2017-11-14
csi: CreateVolume and DeleteVolume support

lpabon:


csi/controller.go
```diff
@@ -315,11 +316,152 @@ func osdVolumeAttributes(v *api.Volume) map[string]string {
...
+		}
+	} else {
+		// Get Capabilities and Size
+		spec.Size = req.GetCapacityRange().GetRequiredBytes()
+		for _, cap := range req.GetVolumeCapabilities() {
+			// Check access mode is setup correctly
+			mode := cap.GetAccessMode().GetMode()
+			if mode == csi.VolumeCapability_AccessMode_MULTI_NODE_MULTI_WRITER ||
+				mode == csi.VolumeCapability_AccessMode_MULTI_NODE_SINGLE_WRITER {
+				spec.Shared = true
```
piyush-nimbalkar: we can break here

csi/controller.go
```diff
@@ -315,11 +316,152 @@ func osdVolumeAttributes(v *api.Volume) map[string]string {
...
+			req.GetVolumeId(),
+			err.Error())
+		dlog.Errorln(e)
+		return nil, status.Error(codes.Internal, e)
+	}
+
+	return &csi.DeleteVolumeResponse{
+		Reply: &csi.DeleteVolumeResponse_Result_{
+			Result: &csi.DeleteVolumeResponse_Result{},
+		},
```
piyush-nimbalkar: is the response supposed to be empty?

pkg/util/volume_test.go
```diff
@@ -0,0 +1,162 @@
...
+	"testing"
+
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/libopenstorage/openstorage/volume/drivers/mock"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestVolumeFromNameFailedToLocateDueToTooManyVolumes(t *testing.T) {
```
piyush-nimbalkar: should we add tests for inspect returning more than one volume or zero volumes (without error), although it does not happen?
similarly one for enumerate returning zero volumes without error?

csi/controller_test.go
```diff
@@ -911,3 +911,222 @@ func TestControllerListVolumes(t *testing.T) {
...
+
+	// Setup mock
+	s.MockDriver().EXPECT().Delete(myid).Return(fmt.Errorf("TEST")).Times(1)
+
+	_, err := c.DeleteVolume(context.Background(), req)
+	assert.NotNil(t, err)
+	serverError, ok := status.FromError(err)
+	assert.True(t, ok)
+	assert.Equal(t, serverError.Code(), codes.Internal)
+	assert.Contains(t, serverError.Message(), "Unable to delete")
```
piyush-nimbalkar: nit: should we check for "TEST" to be sure the error is coming because of the delete call failing? in case, we are using "unable to delete" in multiple places.

csi/controller_test.go
```diff
@@ -911,3 +911,222 @@ func TestControllerListVolumes(t *testing.T) {
 	}
 	assert.Equal(t, found, len(mockVolumeList))
 }
+
+func TestControllerCreateVolume(t *testing.T) {
```
piyush-nimbalkar: we could add some negative test cases for create volume

csi/controller.go
lpabon: Good catch, but if in the future we add more settings, having a break there will have a bad side-effect. Having it go through the whole thing will provide easier enhancement in the near future (since this code and the spec are still heavily in flux).

csi/controller.go
lpabon: Yeah, according to the spec. I found that really weird also :)

csi/controller_test.go
lpabon: Ah! I forgot. You got it.

csi/controller_test.go
lpabon: 👍

pkg/util/volume_test.go
lpabon: yeah, good point. I will add.

csi/controller.go
piyush-nimbalkar: That is why we have tests :)
But I am fine with not adding the break as this code is changing and it does not have a very bad impact if we don't have the break
If you don't see this changing it in near future, then you have add it, else it's fine.

lpabon:
still working on the unit tests.. do not review yet.

csi/controller.go
lpabon: ha! yeah, you are correct. as long as we add tests with each change. You know, I agree, and I will add the change 👍

lpabon:
All set. please review @piyush-nimbalkar @adityadani

pkg/util/volume_test.go
```diff
@@ -0,0 +1,221 @@
...
+	// Setup calls
+	gomock.InOrder(
+		// Return that it was not found
+		driver.
+			EXPECT().
+			Inspect([]string{name}).
+			Return(nil, fmt.Errorf("not found")).
+			Times(1),
+
+		// Return too many volumes
```
piyush-nimbalkar: nit: wrong comment

csi/controller_test.go
```diff
@@ -912,6 +912,504 @@ func TestControllerListVolumes(t *testing.T) {
...
+		},
+	}
+	_, err = c.CreateVolume(context.Background(), req)
+	assert.NotNil(t, err)
+	serverError, ok = status.FromError(err)
+	assert.True(t, ok)
+	assert.Equal(t, serverError.Code(), codes.InvalidArgument)
+	assert.Contains(t, serverError.Message(), "CapacityRange")
+
+	// RequredBytes is zero
```
piyush-nimbalkar: nit: *required

csi/csi.go
```diff
@@ -43,12 +44,13 @@ type OsdCsiServerConfig struct {
 // proxies CSI requests for a single specific driver
 type OsdCsiServer struct {
 	Server
-	listener net.Listener
-	server   *grpc.Server
-	driver   volume.VolumeDriver
-	wg       sync.WaitGroup
-	running  bool
-	lock     sync.Mutex
+	listener    net.Listener
```
sangleganesh: Please add comments for member.

csi/csi.go
```diff
@@ -43,12 +44,13 @@ type OsdCsiServerConfig struct {
...
+	server      *grpc.Server
+	driver      volume.VolumeDriver
+	wg          sync.WaitGroup
+	running     bool
+	lock        sync.Mutex
```
sangleganesh: Does the lock protect access to all members ? 
If so embed the lock within the struct.
If not comments can explain what is being protected using this lock.

pkg/util/volume_test.go
```diff
@@ -0,0 +1,219 @@
...
+	defer mc.Finish()
+
+	driver := mock.NewMockVolumeDriver(mc)
+
+	// Setup calls
+	name := "myvolume"
+	gomock.InOrder(
+		driver.
+			EXPECT().
+			Inspect([]string{name}).
```
sangleganesh: Add test for Inspect(nil) and Inspect([]string{}) if not already.

csi/controller.go
```diff
@@ -315,11 +316,154 @@ func osdVolumeAttributes(v *api.Volume) map[string]string {
...
 
+// CreateVolume is a CSI API which creates a volume on OSD
+// This function supports snapshots if the parent volume id is supplied
+// in the parameters.
+func (s *OsdCsiServer) CreateVolume(
+	ctx context.Context,
+	req *csi.CreateVolumeRequest,
+) (*csi.CreateVolumeResponse, error) {
+
+	// Log request
```
sangleganesh: Add nil check for req ?

csi/controller.go
```diff
@@ -315,11 +316,154 @@ func osdVolumeAttributes(v *api.Volume) map[string]string {
...
+	resp := &csi.CreateVolumeResponse{
+		Reply: &csi.CreateVolumeResponse_Result_{
+			Result: &csi.CreateVolumeResponse_Result{
+				VolumeInfo: volume,
+			},
+		},
+	}
+
+	// Check if the volume has already been created or is in process of creation
+	v, err := util.VolumeFromName(s.driver, req.GetName())
```
sangleganesh: This looks like an error. 
Create must return error if already created - atleast that seems to be general behavior.

csi/controller.go
```diff
@@ -315,11 +316,154 @@ func osdVolumeAttributes(v *api.Volume) map[string]string {
...
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume id must be provided")
+	}
+
+	err := s.driver.Delete(req.GetVolumeId())
+	if err != nil {
+		e := fmt.Sprintf("Unable to delete volume with id %s: %s",
+			req.GetVolumeId(),
+			err.Error())
+		dlog.Errorln(e)
```
sangleganesh: We should come up with guidelines on logging - automated schedulers sometimes keep sending invalid/stale requests and that can cause us to fill up logs.
In this case since the error is returned to the caller one can debug the issue - so the log does not help - so i would not log (my preference is to not log on invalid input, just like we do not log on version or volume name being empty). 
If not anything, for consistency purposes.

csi/controller.go
```diff
@@ -315,11 +316,154 @@ func osdVolumeAttributes(v *api.Volume) map[string]string {
...
+	spec, locator, source, err := s.specHandler.SpecFromOpts(req.GetParameters())
+	if err != nil {
+		e := fmt.Sprintf("Unable to get parameters: %s\n", err.Error())
+		dlog.Errorln(e)
+		return nil, status.Error(codes.InvalidArgument, e)
+	}
+
+	// Check if the caller is asking to create a snapshot or for a new volume
+	var id string
+	if source != nil && len(source.GetParent()) != 0 {
```
sangleganesh: This should be left to the drivers to handle ?

csi/controller.go
lpabon: afaik, grpc will not call with nil, it will not allow it.

csi/controller.go
lpabon: If already created, it will return the information according to the spec.

csi/controller.go
lpabon: That's how it works for the docker manager in osd.

csi/controller.go
lpabon: I'll create an issue for this to follow up.

lpabon: Created #223

csi/csi.go
lpabon: Not sure what you mean, could you show me an example? or are you saying we should add comments from now on for all osd structs?

csi/csi.go
lpabon: locks access to start and stop only

pkg/util/volume_test.go
lpabon: Wouldn't that mean that someone has to pass nil into Inspect()? Could you show me an example of the code in CSI where that would happen?

sangleganesh: If it can be done, it needs to be handled, regardless of someone is doing it today. Can you guarantee that no one in future will do it ?

csi/csi.go
sangleganesh: https://github.com/libopenstorage/openstorage/blob/master/api/api.go

pkg/util/volume_test.go
lpabon: I think I am confused because I am not testing Inspect(), or expecting to have inspect be passed with nil. If the unit test passes nil to Inspect() then it will fail, since it expected something else. I think what you want to be tested is already being tested, right?

csi/csi.go
lpabon: I see some done and some not. This is an internal struct, does it need to be commented in that manner? I agree with you that `OsdCsiServerConfig` should definitely have each member commented since it is used by the caller.

lpabon: Created #224

PR #221, opened 2017-11-15
PWX-4196: Adding /nodehealth REST call

zoxpx:
Adding new REST call /nodehealth to report node health status.  Returned status
is similar to /nodestatus, except that it returns '200 OK' only when the node
is OK and ready to serve storage externally.

Ie [HTTP code // body]:
> 503 Service Unavailable // Node status not OK (STATUS_NOT_IN_QUORUM)
> 503 Service Unavailable // Node status not OK (STATUS_INIT)
> 503 Service Unavailable // Node status not OK (STATUS_MAINTENANCE)
> 200 OK // Node status OK

### DIFF
--- a/api/server/cluster.go
+++ b/api/server/cluster.go
@@ -23,7 +23,8 @@ func (c *clusterApi) Routes() []*Route {
 		{verb: "GET", path: "/cluster/versions", fn: c.versions},
 		{verb: "GET", path: clusterPath("/enumerate", cluster.APIVersion), fn: c.enumerate},
 		{verb: "GET", path: clusterPath("/gossipstate", cluster.APIVersion), fn: c.gossipState},
-		{verb: "GET", path: clusterPath("/nodestatus", cluster.APIVersion), fn: c.nodestatus},
+		{verb: "GET", path: clusterPath("/nodestatus", cluster.APIVersion), fn: c.nodeStatus},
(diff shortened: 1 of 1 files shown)


api/server/cluster.go
```diff
@@ -296,22 +297,48 @@ func (c *clusterApi) status(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(cluster.Status)
 }
 
-func (c *clusterApi) nodestatus(w http.ResponseWriter, r *http.Request) {
-	method := "nodestatus"
-
+func nodeStatusIntl() (api.Status, error) {
```
sangleganesh: nit: Internal is better

api/server/cluster.go
```diff
@@ -296,22 +297,48 @@ func (c *clusterApi) status(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	if st != api.Status_STATUS_OK {
+		err = fmt.Errorf("Node status not OK (%s)", api.Status_name[int32(st)])
+		c.sendError(c.name, method, w, err.Error(), http.StatusServiceUnavailable)
+		return
+	}
 
+	w.Write([]byte("Node status OK\n"))
```
sangleganesh: nit: can this be a const ?

zoxpx:
Looks like UT's are missing for the _whole_ package (added https://github.com/libopenstorage/openstorage/issues/226 to track)

PR #222, opened 2017-11-15
csi: Update CSI spec to Nov 14 2017

lpabon:


csi/controller.go
```diff
@@ -273,10 +264,10 @@ func (s *OsdCsiServer) ListVolumes(
 		dlog.Errorln(errs)
 		return nil, status.Error(codes.Internal, errs)
 	}
-	entries := make([]*csi.ListVolumesResponse_Result_Entry, len(volumes))
+	entries := make([]*csi.ListVolumesResponse_Entry, len(volumes))
 	for i, v := range volumes {
 		// Initialize entry
-		entries[i] = &csi.ListVolumesResponse_Result_Entry{
+		entries[i] = &csi.ListVolumesResponse_Entry{
```
piyush-nimbalkar: just out of curiosity, why do we not append here instead of storing at the index?

lpabon: it is much faster to allocate the memory ahead of time if you already know the exact length of the slice. Calling append on each for loop is costly.

Sometimes you do not know the length and must use append

PR #228, opened 2017-11-16
Add method to remove tags from volume.

sangleganesh:


volume/drivers/aws/aws_test.go
```diff
@@ -4,19 +4,48 @@ import (
...
+		AvailabilityZone: &d.md.zone,
+		VolumeType:       &voltype,
+		Size:             &sz,
+	}
+	labelNames := []string{"label1", "label2"}
+	labels := make(map[string]string)
+	for _, name := range labelNames {
+		labels[name] = name
+	}
+	vol, err := d.ops.Create(ec2Vol, labels)
```
lpabon: Does this means that this test has to be run on aws?

volume/drivers/aws/aws_storage.go
```diff
@@ -226,6 +228,18 @@ func (s *ec2Ops) ApplyTags(
 	return err
 }
 
+func (s *ec2Ops) RemoveTags(
+	v *ec2.Volume,
+	labels map[string]string,
+) error {
+	req := &ec2.DeleteTagsInput{
+		Resources: []*string{v.VolumeId},
```
lpabon: That's weird that the SDK requires a `*string`. Aren't strings already a reference? Still code looks good.

sangleganesh: i think the only time you want to pass string pointers is to differentiate between empty string and no string (i.e. nil). Another use case would be to set those values - since golang strings are immutable - so you need to pass its pointer to modify its value.

volume/drivers/aws/aws_test.go
sangleganesh: yes. The test is run only if it determines that it can be run.

PR #233, opened 2017-11-21
AWS Driver: Do not error out when the block device name is in the "/dev/xvd[b-c][a-z]" format.

adityadani:
- Previously we were able to parse devices only in "/dev/xvd[f-p]" format and unnecessarily error out for devices like /dev/xvdcg

### DIFF
--- a/volume/drivers/aws/aws_storage.go
+++ b/volume/drivers/aws/aws_storage.go
@@ -270,9 +270,8 @@ func (s *ec2Ops) DeviceMappings() (map[string]string, error) {
 	for _, d := range instance.BlockDeviceMappings {
 		if d.DeviceName != nil && d.Ebs != nil && d.Ebs.VolumeId != nil {
 			devName := *d.DeviceName
-			// Per AWS docs EC instances have the root mounted at
-			// /dev/sda1, /dev/xvda this label should be skipped
-			if devName == "/dev/sda1" || devName == "/dev/xvda" {
+			// Skip the root device
+			if devName == *instance.RootDeviceName {
 				continue
 			}
 			// AWS EBS volumes get mapped from /dev/sdN -->/dev/xvdN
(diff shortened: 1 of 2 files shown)


volume/drivers/aws/aws_storage.go
```diff
@@ -271,8 +271,8 @@ func (s *ec2Ops) DeviceMappings() (map[string]string, error) {
 		if d.DeviceName != nil && d.Ebs != nil && d.Ebs.VolumeId != nil {
 			devName := *d.DeviceName
 			// Per AWS docs EC instances have the root mounted at
-			// /dev/sda1, this label should be skipped
-			if devName == "/dev/sda1" {
+			// /dev/sda1, /dev/xvda this label should be skipped
```
sangleganesh: why not use instance.RootDeviceName ?

adityadani: Just to check the RootDeviceName we will have to make an AWS request to get the instance. Seems to me as an overkill. In the AWS docs they state that the root device name is either /dev/sda or /dev/xvda

sangleganesh: instance is available already, why do we need an extra call again ? 
My reading of AWS docs states that the root device is configurable, and they provide examples as using /dev/sda.

adityadani: I was referring to this AWS doc - http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/device_naming.html#available-ec2-device-names

Also, we do not have the EC2 instance object, we just have a string name of the instance. Hence the extra call to EC2 to get the instance object and thus the root device name

sangleganesh: // describe current instance.
func (s *ec2Ops) describe() (***ec2.Instance**, error) 

instance is ec2.Instance


ec2.Instance has 

    // The root device name (for example, /dev/sda1 or /dev/xvda).
    RootDeviceName *string `locationName:"rootDeviceName" type:"string"`

adityadani: Done in incremental.

PR #241, opened 2017-11-30
csi: UnpublishNode

pault84:
Adresses: #236 

```
INFO[0187] Connecting to NBD...
INFO[0187] Attempting to open device /dev/nbd0
INFO[0187] Formatting /dev/nbd0 with FS_TYPE_EXT4
INFO[0187] BUSE mapped NBD device /dev/nbd0 (size=10000000) to block file /var/lib/openstorage/buse/92208801-f13c-4b5f-bca7-045be1369b0f
INFO[0303] BUSE mounted NBD device /dev/nbd0 at /mnt
INFO[0303] Voume 92208801-f13c-4b5f-bca7-045be1369b0f mounted on /mnt
INFO[0329] Volume 92208801-f13c-4b5f-bca7-045be1369b0f unmounted
```

### DIFF
--- a/csi/node.go
+++ b/csi/node.go
@@ -166,6 +166,14 @@ func (s *OsdCsiServer) NodeUnpublishVolume(
 		return nil, status.Error(codes.InvalidArgument, "Target path must be provided")
 	}
 
+	// Get volume information
+	_, err := util.VolumeFromName(s.driver, req.GetVolumeId())
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "Volume id %s not found: %s",
+			req.GetVolumeId(),
+			err.Error())
+	}
+
 	// Verify target location is an existing directory
 	// See: https://github.com/container-storage-interface/spec/issues/60
 	if err = verifyTargetLocation(req.GetTargetPath()); err != nil {
--- a/csi/node.go
+++ b/csi/node.go
@@ -148,9 +148,6 @@ func (s *OsdCsiServer) NodePublishVolume(
 }
 
 // NodeUnpublishVolume is a CSI API call which unmounts the volume.
-//
-// TODO: Support READ ONLY Mounts
-//
 func (s *OsdCsiServer) NodeUnpublishVolume(
 	ctx context.Context,
 	req *csi.NodeUnpublishVolumeRequest,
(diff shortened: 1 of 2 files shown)


csi/node.go
```diff
@@ -140,13 +140,71 @@ func (s *OsdCsiServer) NodePublishVolume(
...
+	dlog.Infof("Volume %s mounted on %s",
 		req.GetVolumeId(),
 		req.GetTargetPath())
 
 	return &csi.NodePublishVolumeResponse{}, nil
 }
 
+// NodeUnpublishVolume is a CSI API call which unmounts the volume.
+//
+// TODO: Support READ ONLY Mounts
```
lpabon: This is probably not needed.

csi/node.go
```diff
@@ -140,13 +140,71 @@ func (s *OsdCsiServer) NodePublishVolume(
...
+	if err = s.driver.Unmount(v.GetId(), req.GetTargetPath(), nil); err != nil {
+		return nil, status.Errorf(
+			codes.Internal,
+			"Unable to unmount volume %s onto %s: %s",
+			req.GetVolumeId(),
+			req.GetTargetPath(),
+			err.Error())
+	}
+
+	if err = s.driver.Detach(req.GetVolumeId(), nil); err != nil {
```
lpabon: We should only do deatch for Type() Block, afaik

pault84: @adityadani is that correct? if so i'll make the fix

prabirpaul: Adding a check here similar to one in NodePublishVolume () means csi code would need to change if a driver implementation changes. This could be prevented by appropriate handling of unimplemented calls in the driver.

csi/node.go
```diff
@@ -140,13 +140,70 @@ func (s *OsdCsiServer) NodePublishVolume(
...
+	}
+
+	// Verify target location is an existing directory
+	// See: https://github.com/container-storage-interface/spec/issues/60
+	if err = verifyTargetLocation(req.GetTargetPath()); err != nil {
+		return nil, status.Error(codes.NotFound, err.Error())
+	}
+
+	// Mount volume onto the path
+	if err = s.driver.Unmount(v.GetId(), req.GetTargetPath(), nil); err != nil {
```
adityadani: Either use v.GetId() OR req.GetVolumeId() for consistency.

csi/node.go
adityadani: I agree with @prabirpaul . All the drivers in openstorage do implement Detach calls. It could just be a no-op for such drivers. Ideally we won't need this extra check.

lpabon: That's fine, but then we would need to create a new issue for this to solve at another PR.

csi/node.go
lpabon: yeah, that makes sense. I changed from v.GetId() to req.GetVolumeId() in NodePublishVolume

csi/node.go
pault84: So keep the check for block or not? my last commit has this check.

PR #248, opened 2017-12-12
Adding credential APIs.

talakad:
These APIs allow clients to configure and manage credentials for
different cloud providers.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -62,12 +62,37 @@ const (
 	OptTimeoutSec = "TimeoutSec"
 	// OptQuiesceID query parameter use for quiesce
 	OptQuiesceID = "QuiesceID"
+	// OptCredUUID is the UUID of the credential
+	OptCredUUID = "CredUUID"
+	// OptCredType  indicates type of credential
+	OptCredType = "CredType"
+	// OptCredEncrKey indicates if the key is encrypted
+	OptCredEncrKey = "CredEncrypt"
+	// OptCredRegion indicates the region
+	OptCredRegion = "CredRegion"
+	// OptCredDisableSSL indicated if SSL should be disabled
+	OptCredDisableSSL = "CredDisableSSL"
+	// OptCredEndpoint indicate the cloud endpoint
+	OptCredEndpoint = "CredEndpoint"
+	// OptCredAccKey for s3
+	OptCredAccessKey = "CredAccessKey"
+	// OptCredSecretKey
+	OptCredSecretKey = "CredSecretKey"
+	// OptCredGoogleProjectID
+	OptCredGoogleProjectID = "CredProjectID"
+	// OptCredGoogleJsonKey
+	OptCredGoogleJsonKey = "CredJsonKey"
+	// OptCredAzureAccountName
+	OptCredAzureAccountName = "CredAccountName"
+	// OptOptCredAzureAccountKey
+	OptCredAzureAccountKey = "CredAccountKey"
 )
 
 // Api client-server Constants
 const (
 	OsdVolumePath   = "osd-volumes"
 	OsdSnapshotPath = "osd-snapshot"
+	OsdCredsPath    = "osd-creds"
 	TimeLayout      = "Jan 2 15:04:05 UTC 2006"
 )
 
@@ -150,6 +175,20 @@ type StatPoint struct {
 	Fields map[string]interface{}
 	// Timestamp in Unix format
 	Timestamp int64
+}
+
+// CredCreateRequest is the input for CredCreate command
+type CredCreateRequest struct {
+	// InputParams is map describing cloud provide
+	InputParams map[string]string
+}
+
+// CredCreateResponse is returned for CredCreate command
+type CredCreateResponse struct {
+	// UUID of the credential that was just created
+	UUID string
+	// CredErr indicates reasonfor failed CredCreate
+	CredErr error
 }
 
 // DriverTypeSimpleValueOf returns the string format of DriverType
--- a/api/api.go
+++ b/api/api.go
@@ -66,9 +66,9 @@ const (
 	OptCredUUID = "CredUUID"
 	// OptCredType  indicates type of credential
 	OptCredType = "CredType"
-	// OptCredEncrKey indicates if the key is encrypted
+	// OptCredEncrKey is the key used to encrypt data
 	OptCredEncrKey = "CredEncrypt"
-	// OptCredRegion indicates the region
+	// OptCredRegion indicates the region for s3
 	OptCredRegion = "CredRegion"
 	// OptCredDisableSSL indicated if SSL should be disabled
 	OptCredDisableSSL = "CredDisableSSL"
@@ -76,15 +76,17 @@ const (
 	OptCredEndpoint = "CredEndpoint"
 	// OptCredAccKey for s3
 	OptCredAccessKey = "CredAccessKey"
-	// OptCredSecretKey
+	// OptCredSecretKey for s3
 	OptCredSecretKey = "CredSecretKey"
-	// OptCredGoogleProjectID
+	// OptCredGoogleProjectID projectID for google cloud
 	OptCredGoogleProjectID = "CredProjectID"
-	// OptCredGoogleJsonKey
+	// OptCredGoogleJsonKey for google cloud
 	OptCredGoogleJsonKey = "CredJsonKey"
-	// OptCredAzureAccountName
+	// OptCredAzureAccountName is the account name for
+	// azure as the cloud provider
 	OptCredAzureAccountName = "CredAccountName"
-	// OptOptCredAzureAccountKey
+	// OptOptCredAzureAccountKey is the accountkey for
+	// azure as the cloud provider
 	OptCredAzureAccountKey = "CredAccountKey"
 )
 
--- a/api/api.go
+++ b/api/api.go
@@ -190,7 +190,7 @@ type CredCreateResponse struct {
 	// UUID of the credential that was just created
 	UUID string
 	// CredErr indicates reasonfor failed CredCreate
-	CredErr error
+	CredErr string
 }
 
 // DriverTypeSimpleValueOf returns the string format of DriverType
--- a/api/client/volume/client.go
+++ b/api/client/volume/client.go
@@ -4,19 +4,19 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
-	"io"
-	"io/ioutil"
-	"strconv"
-
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/client"
 	"github.com/libopenstorage/openstorage/volume"
+	"io"
+	"io/ioutil"
+	"strconv"
 )
 
 const (
 	graphPath  = "/graph"
 	volumePath = "/osd-volumes"
 	snapPath   = "/osd-snapshot"
+	credsPath  = "/osd-creds"
 )
 
 type volumeClient struct {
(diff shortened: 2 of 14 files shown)


api/api.go
```diff
@@ -62,12 +62,37 @@ const (
...
+	OptCredEncrKey = "CredEncrypt"
+	// OptCredRegion indicates the region
+	OptCredRegion = "CredRegion"
+	// OptCredDisableSSL indicated if SSL should be disabled
+	OptCredDisableSSL = "CredDisableSSL"
+	// OptCredEndpoint indicate the cloud endpoint
+	OptCredEndpoint = "CredEndpoint"
+	// OptCredAccKey for s3
+	OptCredAccessKey = "CredAccessKey"
+	// OptCredSecretKey
```
lpabon: Could you add some comments on the next few? The comments is only the name of the variable at the moment.

api/client/volume/client.go
```diff
@@ -435,3 +435,51 @@ func (v *volumeClient) Unquiesce(volumeID string) error {
 	}
 	return nil
 }
+
+func (v *volumeClient) CredsList() (map[string]interface{}, error) {
+	creds := make(map[string]interface{}, 0)
+	err := v.c.Get().Resource(credsPath + "/credslist").Do().Unmarshal(&creds)
```
lpabon: I'll skip these for the this review, because it looks like all APIs do this, but there should be a check on the returned HTTP status of the header. This is more of an overall question about all of OSD client calls.

volume/volume.go
```diff
@@ -194,6 +195,17 @@ type BlockDriver interface {
 	Detach(volumeID string, options map[string]string) error
 }
 
+// CredsDriver provides methods to handle credentials
+type CredsDriver interface {
+	CredsCreate(params map[string]string) (string, error)
```
lpabon: These need to be commented

api/server/volume.go
```diff
@@ -555,6 +555,101 @@ func (vd *volAPI) unquiesce(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+
+	d, err := vd.getVolDriver(r)
+	if err != nil {
+		notFound(w, r)
+		return
+	}
+
+	if err := d.CredsValidate(uuid); err != nil {
```
lpabon: You can use the mockdriver to test these calls. Here is a sample of how to set it up: https://github.com/libopenstorage/openstorage/blob/master/csi/csi_test.go#L50-L56

lpabon: Also, you will need to deregister it on every test: https://github.com/libopenstorage/openstorage/blob/master/csi/csi_test.go#L87

api/api.go
talakad: Sure. Will add.

api/server/volume.go
talakad: Ok. I will look into this.

volume/volume.go
talakad: Yes. I will add those.

volume/volume.go
```diff
@@ -194,6 +195,18 @@ type BlockDriver interface {
...
 }
 
+// CredsDriver provides methods to handle credentials
+type CredsDriver interface {
+	// CredsCreate creates credential for a given cloud provider
+	CredsCreate(params map[string]string) (string, error)
+	// CredsList lists the configured credentials in the cluster
+	CredsList() (map[string]interface{}, error)
+	// CredsDelete deletes the credential associated schedUUID
+	CredsDelete(schedUUID string) error
```
sangleganesh: creds interface should not deal with *sched* . Interface should just be Create/Delete/Enumerate

volume/volume.go
```diff
@@ -194,6 +195,18 @@ type BlockDriver interface {
...
+	// CredsValidate validates the credential associated schedUUID
+	CredsValidate(schedUUID string) error
```
sangleganesh: creds should not be related to sched.

volume/volume.go
```diff
@@ -194,6 +195,18 @@ type BlockDriver interface {
 	Detach(volumeID string, options map[string]string) error
 }
 
+// CredsDriver provides methods to handle credentials
+type CredsDriver interface {
+	// CredsCreate creates credential for a given cloud provider
+	CredsCreate(params map[string]string) (string, error)
+	// CredsList lists the configured credentials in the cluster
+	CredsList() (map[string]interface{}, error)
```
sangleganesh: s/List/Enumerate

api/server/volume.go
```diff
@@ -598,5 +697,9 @@ func (vd *volAPI) Routes() []*Route {
 		{verb: "POST", path: snapPath("", volume.APIVersion), fn: vd.snap},
 		{verb: "GET", path: snapPath("", volume.APIVersion), fn: vd.snapEnumerate},
 		{verb: "POST", path: snapPath("/restore/{id}", volume.APIVersion), fn: vd.restore},
+		{verb: "GET", path: credsPath("/credslist", volume.APIVersion), fn: vd.credsList},
```
sangleganesh: credsPath("/credslist") can be simply credsPath("/endpoint").
Also - get without parameter is enumerate all
get with parameter is single get

api/server/volume.go
```diff
@@ -598,5 +697,9 @@ func (vd *volAPI) Routes() []*Route {
...
+		{verb: "POST", path: credsPath("/credscreate", volume.APIVersion), fn: vd.credsCreate},
+		{verb: "DELETE", path: credsPath("/credsdelete", volume.APIVersion), fn: vd.credsDelete},
```
sangleganesh: since the verb is DELETE, no need for "/credsdelete".

api/server/volume.go
```diff
@@ -598,5 +697,9 @@ func (vd *volAPI) Routes() []*Route {
 		{verb: "POST", path: snapPath("", volume.APIVersion), fn: vd.snap},
 		{verb: "GET", path: snapPath("", volume.APIVersion), fn: vd.snapEnumerate},
 		{verb: "POST", path: snapPath("/restore/{id}", volume.APIVersion), fn: vd.restore},
+		{verb: "GET", path: credsPath("/credslist", volume.APIVersion), fn: vd.credsList},
+		{verb: "POST", path: credsPath("/credscreate", volume.APIVersion), fn: vd.credsCreate},
```
sangleganesh: Similarly, a POST to the "/creds" can be considered simply as create, no need for "/credscreate"

api/server/volume.go
```diff
@@ -598,5 +697,9 @@ func (vd *volAPI) Routes() []*Route {
...
+		{verb: "DELETE", path: credsPath("/credsdelete", volume.APIVersion), fn: vd.credsDelete},
+		{verb: "POST", path: credsPath("/credsvalidate", volume.APIVersion), fn: vd.credsValidate},
```
sangleganesh: not sure what this API is doing ?

api/server/volume.go
```diff
@@ -555,6 +555,101 @@ func (vd *volAPI) unquiesce(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(volumeResponse)
 }
 
+func (vd *volAPI) credsList(w http.ResponseWriter, r *http.Request) {
```
sangleganesh: it may be more cleaner to add creds related code to creds.go

api/api.go
```diff
@@ -62,12 +62,39 @@ const (
...
+	// OptCredGoogleProjectID projectID for google cloud
+	OptCredGoogleProjectID = "CredProjectID"
+	// OptCredGoogleJsonKey for google cloud
+	OptCredGoogleJsonKey = "CredJsonKey"
+	// OptCredAzureAccountName is the account name for
+	// azure as the cloud provider
+	OptCredAzureAccountName = "CredAccountName"
+	// OptOptCredAzureAccountKey is the accountkey for
+	// azure as the cloud provider
+	OptCredAzureAccountKey = "CredAccountKey"
```
sangleganesh: Seems some of these parameters are similar (azure account key == google json key, azure account name == google project id == was access key ? )

talakad: Better to name those parameters as they are called.. otherwise it can be confusing.

api/server/volume.go
talakad: That was the the way i started. However unit tests with httptest does not work with that mode.

api/server/volume.go
talakad: Yes. same as above.

api/server/volume.go
talakad: OK. will do this.

api/server/volume.go
talakad: Returns no error if it can access cloud provider with the given credential.

volume/volume.go
talakad: OK.

volume/volume.go
talakad: Good catch with naming!. Will fix it.

volume/volume.go
talakad: same as above.

api/server/volume.go
sangleganesh: @lpabon can you help @talakad  with UT questions ? thanks !

lpabon: @talakad I'm not sure what you mean. Let's talk tomorrow and see what we can do.

api/client/volume/client.go
```diff
@@ -4,19 +4,19 @@ import (
...
+	"io"
+	"io/ioutil"
+	"strconv"
 )
 
 const (
 	graphPath  = "/graph"
 	volumePath = "/osd-volumes"
 	snapPath   = "/osd-snapshot"
+	credsPath  = "/osd-creds"
```
sangleganesh: nit: credsPath = "/" + api.OsdCredsPath

api/client/volume/client.go
```diff
@@ -435,3 +435,51 @@ func (v *volumeClient) Unquiesce(volumeID string) error {
...
+	}
+	if response.Error != "" {
+		return errors.New(response.Error)
+	}
+	return nil
+}
+
+func (v *volumeClient) CredsValidate(uuid string) error {
+	response := &api.VolumeResponse{}
+	req := v.c.Post().Resource(credsPath + "/credsvalidate")
```
sangleganesh: can remove "creds" from "credsvalidate"

api/client/volume/client_volume_test.go
```diff
@@ -29,3 +28,144 @@ func TestClientTLS(t *testing.T) {
...
+			http.Error(w, "Missing uuid param", http.StatusBadRequest)
+			return
+		}
+		response := &api.VolumeResponse{}
+		json.NewEncoder(w).Encode(response)
+
+	}))
+	defer ts.Close()
+
+	//clnt, err := NewDriverClient(ts.URL, "pxd", volume.APIVersion, "")
```
sangleganesh: nit: remove commented code here and below

api/client/volume/client_volume_test.go
```diff
@@ -29,3 +28,144 @@ func TestClientTLS(t *testing.T) {
 
 	require.NoError(t, err)
 }
+
+func TestClientCredCreate(t *testing.T) {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
sangleganesh: If i understand correctly, this test code starts its own server, returns its own responses and bypasses the OSD code all together. Don't see why is it needed .. ?

talakad: This tests the client code alone. Mock driver for volumeDriver would support end-to-end ( with the volume driver implementation being mocked).Still under development.

api/client/volume/client.go
talakad: Will leave this as it is to match the rest of the file.

lpabon:
I'll review this tomorrow 12/19

lpabon:
Reviewed 6 of 13 files at r1, 1 of 2 files at r2, 4 of 7 files at r3, 3 of 3 files at r4.
Review status: all files reviewed at latest revision, 19 unresolved discussions.

---

*[api/server/creds.go, line 1 at r4](https://reviewable.io:443/reviews/libopenstorage/openstorage/248#-L0lcvMdA4I1NcU-TcxK:-L0lcvMdA4I1NcU-TcxL:b-ug1w1f) ([raw file](https://github.com/libopenstorage/openstorage/blob/5ab3cf34fece0ffb729498f44ed454edaba25c05/api/server/creds.go#L1)):*
> ```Go
> package server
> ```

This file is not unit tested. This needs to be fixed.


Please create an issue to add this following this PR.

---

*[api/server/creds.go, line 26 at r4](https://reviewable.io:443/reviews/libopenstorage/openstorage/248#-L0lbyIpxLd8OVX9HkQD:-L0lbyIpxLd8OVX9HkQE:b-y13ibz) ([raw file](https://github.com/libopenstorage/openstorage/blob/5ab3cf34fece0ffb729498f44ed454edaba25c05/api/server/creds.go#L26)):*
> ```Go
> 		return
> 	}
> 	json.NewEncoder(w).Encode(creds)
> ```

This returns an error, but it is silently ignored. We should not ignore it. If it returns an error, we should return an HTTP error status of 5XX to show it was an internal error, with the message showing the error.

The same situation happens throughout this file.

Please create an issue for a following fix.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/libopenstorage/openstorage/248)*

lpabon:
@talakad No show-stopper issues, but the change still needs a few updates.

PR #249, opened 2017-12-13
added proto for loopdev and ramdisk

antonipx:
Important: I only changed `api/api.proto` and `cli/cli_test.go`. I see changes in `flexvolume.pb.gw.go` and `api.pb.go` which I did not make. I'm unsure of their impact at present.


cli/cli_test.go
```diff
@@ -34,7 +34,9 @@ func TestCmdMarshalProto(t *testing.T) {
  "sticky": false,
  "group_enforced": false,
  "compressed": false,
- "cascaded": false
+ "cascaded": false,
+ "loopdev": false,
+ "ramdisk": false
```
lpabon: Are you testing that the values are always false? I think you should test for setting values to true or false, right? Who is going to take action with these values?

lpabon:
@antonipx Could you describe why these are needed?

antonipx:
@lpabon The changes are needed to implement two features for Portworx. Loopback device on top of PXD volume and ramdisk as backing store. Proto changes are required for these to be instrumented via pxctl, etc.

lpabon:
@antonipx Thank for the description. What tests need to be written to test this?

ggriffiths:

PR #253, opened 2017-12-19
Apidocs

pault84:
api docmentation generation...

make docs will generate swagger.json. in cmd/osd/ might move that  i guess.


pault84:
looks like i'm going to have to rebase

pault84:
Alright so i have to figure out how to add examples to some of the parameters from volumeEnumerate and snapshotEnumerate.

swagger 2.0 does not allow for serialized objects as a form parameter.
which causes the {} part of the object to disappear and that breaks the api call.

i can also add examples.

guidance would be nice.

api/api.go
```diff
@@ -80,77 +80,61 @@ const (
...
-	Id        string
-	Cpu       float64 // percentage.
-	MemTotal  uint64
-	MemUsed   uint64
-	MemFree   uint64
-	Avgload   int
-	Status    Status
+	// Id of the cluster.
+	Id string
+	// Id of the cluster.
```
adityadani: CpuUsage of the node in percentage.

api/api.go
```diff
@@ -80,77 +80,61 @@ const (
...
+	Cpu float64 // percentage.
+	// Total Memory of the node
```
adityadani: Shouldnt this be of the form
MemTotal is the total memory of the node
The same applies to the rest of the fields as well.

api/api.go
```diff
@@ -80,77 +80,61 @@ const (
...
+// swagger:model
 type Node struct {
-	Id        string
-	Cpu       float64 // percentage.
-	MemTotal  uint64
-	MemUsed   uint64
-	MemFree   uint64
-	Avgload   int
-	Status    Status
+	// Id of the cluster.
```
adityadani: Id of the node.

api/client/cluster/client.go
```diff
@@ -142,94 +142,6 @@ func (c *clusterClient) EnableUpdates() error {
 	return nil
 }
 
-func (c *clusterClient) SetLoggingURL(loggingURL string) error {
```
adityadani: Mark it as deprecated. And we will define a strategy to deprecate this if not needed.

api/server/cluster.go
```diff
@@ -57,6 +52,22 @@ func (c *clusterApi) String() string {
...
+// Lists cluster Nodes.
+//
+// This will return the entire cluster object and it's nodes.
+//
+// ---
+// produces:
+// - application/json
+// responses:
+//   '200':
+//      description: current cluster state
```
adityadani: Can we add description in the form of complete sentence as this is going to be a part of the swagger docs?

api/server/cluster.go
```diff
@@ -316,6 +238,18 @@ func nodeStatusIntl() (api.Status, error) {
...
+// swagger:operation GET /cluster/nodestatus node status nodeStatus
+//
+// This will return the node status .
+//
+// ---
+// produces:
+// - application/json
+// responses:
+//   '200':
+//      description: node status
```
adityadani: returns the status of the node serving this http request.

api/server/cluster.go
```diff
@@ -346,6 +292,18 @@ func (c *clusterApi) nodeHealth(w http.ResponseWriter, r *http.Request) {
...
+// This will return the peer node status
+//
+// ---
+// produces:
+// - application/json
+// responses:
+//   '200':
+//      description: anode
+//      schema:
+//         type: string
```
adityadani: this API takes in a nodeID as an argument

api/server/cluster.go
```diff
@@ -426,6 +436,25 @@ func (c *clusterApi) versions(w http.ResponseWriter, r *http.Request) {
...
+// - name: resource
+//   in: path
+//   description: resourcetype to get alerts with
+//   required: true
+//   schema:
+//    "$ref": "#/definitions/ResourceType"
+// responses:
+//   '200':
+//      description: Alerts object
+//      schema:
```
adityadani: type: array

api/server/volume.go
```diff
@@ -67,6 +67,30 @@ func (vd *volAPI) parseID(r *http.Request) (string, error) {
...
+//
+// Creates a single volume with given spec.
+//
+// ---
+// produces:
+// - application/json
+// parameters:
+// - name: spec
+//   in: body
+//   description: spec to create volume with
```
adityadani: not really spec but it takes the VolumeCreateRequest. You could say "request to create volume"

api/server/volume.go
```diff
@@ -531,6 +824,27 @@ func (vd *volAPI) quiesce(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(volumeResponse)
 }
 
+// swagger:operation GET /osd-volumes/unquiesce/{id} volume unquiesce unquiesceVolume
```
adityadani: this is POST

api/server/volume.go
```diff
@@ -488,6 +760,27 @@ func (vd *volAPI) requests(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(requests)
 }
 
+// swagger:operation GET /osd-volumes/quiesce/{id} volume quiesce quiesceVolume
```
adityadani: this is POST

api/api.go
pault84: good catch

api/api.go
pault84: Even better catch (copy paste hero)

api/api.go
pault84: You are right i'll change that

api/server/cluster.go
pault84: Yes we can i just took what they used in the swagger documentation

api/server/cluster.go
pault84: it does? i didn't see it in the route. but good catch!

api/server/cluster.go
pault84: Alerts is an object that has an array in it.

api/client/cluster/client.go
```diff
@@ -160,76 +161,6 @@ func (c *clusterClient) SetLoggingURL(loggingURL string) error {
 
 }
 
-func (c *clusterClient) SetManagementURL(managementURL string) error {
```
lpabon: Why is this removed for a docs PR? If this needs to be removed, please create a new PR just for this change.

pault84: None of these features have been used as far as i know but i can put them back and file a seperate pr for htose

pault84:
this is good too merge.

lpabon:
I'm going to review it tomorrow

lpabon:
Looks like there are a lot of errors but we can fix them after this PR. Could you create an issue to fix the errors? See: https://app.swaggerhub.com/apis/libopenstorage/osd-api/1.0.0

lpabon:
Two changes needed:

1. Could you add to the Makefile to automatically download go-swagger.
1. Output of swagger.json should not be in `cmd/osd`. Could we add it to `api/swagger` ?

PR #254, opened 2017-12-19
Backup ws

talakad:
Added APIs for Cloud backup driver.  These APIs allow management of volume backups in cloud including management of schedules for the cloud backups.


api/server/backup.go
```diff
@@ -0,0 +1,250 @@
+package server
+
+import (
+	"encoding/json"
+	"github.com/libopenstorage/openstorage/api"
+	"net/http"
+)
+
+func (vd *volAPI) backup(w http.ResponseWriter, r *http.Request) {
```
sangleganesh: seems like there is a slight inconsistency in the end points w.r.t to params, sometimes they only accept objects, sometimes multiple params are passed separately using key=val form. should we just pick and chose one of these ? (objects are better because adding fields is easier w.r.t to signature change).

volume/volume_not_supported.go
```diff
@@ -89,3 +92,104 @@ func (s *quiesceNotSupported) Quiesce(
...
+func (cl *cloudBackupNotSupported) BackupSchedCreate(
+	input *api.BackupSchedRequest,
+) *api.BackupSchedResponse {
+	response := api.BackupSchedResponse{
+		SchedCreateErr: ErrNotSupported.Error(),
+	}
+	return &response
+}
+
+// BackupSchedDelete delete a volume's cloud backup-schedule
```
sangleganesh: i think its not necessary to add comments here

api/api.go
```diff
@@ -152,6 +164,171 @@ type StatPoint struct {
...
+}
+
+type BackupSchedResponse struct {
+	// SchedUUID is the UUID of the newly created schedule
+	SchedUUID string
+	//SchedCreateErr indicates any error while creating backupschedule
+	SchedCreateErr string
+}
+
+type BackupScheduleInfo struct {
```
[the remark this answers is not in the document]
talakad: OK. I'll add common input and response structures.

api/server/backup.go
talakad: I see. I 'll object to all the apis.

volume/volume_not_supported.go
talakad: Ok

api/client/volume/client_volume_test.go
```diff
@@ -29,3 +29,302 @@ func TestClientTLS(t *testing.T) {
...
+	request := &api.BackupEnumerateRequest{}
+	response := VolumeDriver(clnt).BackupEnumerate(request)
+	require.NotEqual(t, response.EnumerateErr, "")
+
+	request.CredentialUUID = "3450987910078"
+	response = VolumeDriver(clnt).BackupEnumerate(request)
+	require.Equal(t, response.EnumerateErr, "")
+}
+
+func TestClientDeleteBackup(t *testing.T) {
```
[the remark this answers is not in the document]
talakad: Totally agree. Lets discuss offline.

lpabon:
Closed by #298

PR #255, opened 2017-12-20
Refactor AWS storage operations implementation to use common interface and add go tests

harsh-px:
This is primarily a move of the implementation from the drivers/volume/aws/aws_storage.go driver to pkg/storageops/aws. Changes were made to refactor it as a interface impl.

Also go tests were added to test the aws storage implementation.


pkg/storageops/aws/aws.go
```diff
@@ -18,90 +22,46 @@ type ec2Ops struct {
...
-	Snapshot(volumeID string, readonly bool) (*ec2.Snapshot, error)
-	// ApplyTags
-	ApplyTags(v *ec2.Volume, labels map[string]string) error
-	// RemoveTags removes labels from the volume
-	RemoveTags(v *ec2.Volume, labels map[string]string) error
-	// Tags
-	Tags(v *ec2.Volume) map[string]string
-}
+	_, err := credentials.NewEnvCredentials().Get()
+	if err != nil {
```
adityadani: nit: if _, err := credentials.NewEnvCredentials(); err != nil {

pkg/storageops/aws/aws.go
```diff
@@ -216,24 +176,31 @@ func (s *ec2Ops) waitAttachmentStatus(
...
 
+func (s *ec2Ops) Name() string { return "aws" }
+
 func (s *ec2Ops) ApplyTags(
-	v *ec2.Volume,
+	v interface{},
 	labels map[string]string,
 ) error {
+
+	vol := v.(*ec2.Volume)
```
adityadani: vol, ok := v.(*ec2.Volume)
check if type cast is successful and avoid go panics.

pkg/storageops/aws/aws.go
```diff
@@ -216,24 +176,31 @@ func (s *ec2Ops) waitAttachmentStatus(
...
 	_, err := s.ec2.CreateTags(req)
 	return err
 }
 
 func (s *ec2Ops) RemoveTags(
-	v *ec2.Volume,
+	v interface{},
 	labels map[string]string,
 ) error {
+	vol := v.(*ec2.Volume)
```
adityadani: vol, ok := v.(*ec2.Volume)
change all the typecasts in this file.

pkg/storageops/aws/aws.go
```diff
@@ -377,30 +360,47 @@ func (s *ec2Ops) available(v *ec2.Volume) bool {
...
+	var awsVols = make([]interface{}, len(resp.Volumes))
+	for i, v := range resp.Volumes {
+		awsVols[i] = v
+	}
+
+	return awsVols, nil
 }
 
-func (s *ec2Ops) Tags(v *ec2.Volume) map[string]string {
+func (s *ec2Ops) Tags(v interface{}) (map[string]string, error) {
```
adityadani: Can we just change the API itself to Tags(volumeID string)  instead of doing a refreshVol on an ec2Volume ?

pkg/storageops/aws/aws.go
```diff
@@ -529,40 +536,46 @@ func (s *ec2Ops) Detach(volumeID string) error {
...
-) (*ec2.Snapshot, error) {
+) (interface{}, error) {
 	request := &ec2.CreateSnapshotInput{
 		VolumeId: &volumeID,
 	}
 	return s.ec2.CreateSnapshot(request)
 }
 
-func (s *ec2Ops) DevicePath(vol *ec2.Volume) (string, error) {
+func (s *ec2Ops) DevicePath(v interface{}) (string, error) {
```
adityadani: same here. May be we could just change the API to DevicePath(volumeId string). So that we don't need to pass in templates or the caller doesn't need to create a ec2Volume object. What do you think?

pkg/storageops/storageops_test.go
```diff
@@ -0,0 +1,152 @@
...
+var drivers map[string]storageops.Ops
+var diskTemplates map[string]map[string]interface{}
+var diskName = fmt.Sprintf("%s-%s", newDiskPrefix, uuid.NewV4())
+
+func initDrivers(t *testing.T) {
+	drivers = make(map[string]storageops.Ops)
+	diskTemplates = make(map[string]map[string]interface{})
+
+	// AWS
+	if d, err := aws.NewEnvClient(); err != aws.ErrAWSEnvNotAvailable {
```
adityadani: nit: we could execute this AWS specific code under the aws/storageops_test.go and then pass in the StorageOps interface to this test suite. In this way this file would be cloud provider agnostic.

pkg/storageops/aws/aws.go
harsh-px: Fixed in incremental

pkg/storageops/aws/aws.go
harsh-px: Fixed in incremental

pkg/storageops/aws/aws.go
harsh-px: Changed API in incremental

pkg/storageops/aws/aws.go
harsh-px: Changed API in incremental

pkg/storageops/storageops_test.go
harsh-px: Changed API in incremental

pkg/storageops/aws/aws.go
harsh-px: Changed API in incremental

pkg/storageops/aws/aws.go
```diff
@@ -428,47 +414,49 @@ func (s *ec2Ops) Enumerate(
...
 	if len(labels) > 0 {
-		if err = s.ApplyTags(newVol, labels); err != nil {
-			return nil, s.rollbackCreate(*newVol.VolumeId, err)
+		if err = s.ApplyTags(*resp.VolumeId, labels); err != nil {
+			return nil, s.rollbackCreate(*resp.VolumeId, err)
 		}
 	}
 
-	return newVol, nil
+	return s.refreshVol(resp.VolumeId)
```
adityadani: good catch!

pkg/storageops/aws/aws.go
```diff
@@ -428,47 +414,49 @@ func (s *ec2Ops) Enumerate(
...
 	return sets, nil
 }
 
 func (s *ec2Ops) Create(
-	v *ec2.Volume,
+	v interface{},
 	labels map[string]string,
-) (*ec2.Volume, error) {
+) (interface{}, error) {
+	vol := v.(*ec2.Volume)
```
adityadani: safe-guard against go panics.
```
vol, ok := v.(*ec2.Volume)
if !ok {
    return nil, err
}
```
Do the same at all places where you are type casting.

volume/drivers/aws/aws.go
```diff
@@ -203,11 +205,13 @@ func (d *Driver) Create(
...
 		ec2Vol.Iops = iops
 	}
-	vol, err := d.ops.Create(ec2Vol, locator.VolumeLabels)
+	resp, err := d.ops.Create(ec2Vol, locator.VolumeLabels)
 	if err != nil {
 		dlog.Warnf("Failed in CreateVolumeRequest :%v", err)
 		return "", err
 	}
+
+	vol := resp.(*ec2.Volume)
```
adityadani: Doesn't hurt to add the type cast check here as well even if ops.Create is going to return an ec2.Volume

harsh-px:
Add checks for all typecasts.

PR #256, opened 2017-12-20
Adding unit test cases for Volume management rest endpoints.

ipochi:
Adding test cases for http endpoints for volume management functions
Volume create,delete,inspect,attach,detach,mount.


api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+		Spec:    &api.VolumeSpec{Size: size},
+	}
+
+	gomock.InOrder(
+		ts.MockDriver().
+			EXPECT().
+			Set(id, req.GetLocator(), req.GetSpec()).
+			Return(nil),
+
+		ts.MockDriver().
```
adityadani: If I am not wrong wouldn't the Set() call with this request object result into an Attach and Mount to be called as well ?

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+					},
+				},
+			}, nil),
+	)
+
+	// create driver client
+
+	driverclient := volumeclient.VolumeDriver(ts.client)
+
+	res := driverclient.Set(id, req.GetLocator(), req.GetSpec())
```
adityadani: I am  not sure how this test is passing?

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+	assert.Nil(t, err)
+
+	name := "myvol"
+	id := "myid"
+	size := uint64(10)
+
+	req := &api.VolumeSetRequest{
+		Options: map[string]string{},
+		Action: &api.VolumeStateAction{
+			Attach: api.VolumeActionParam_VOLUME_ACTION_PARAM_ON},
```
adityadani: nit:  follow styleguide

Attach: api.VolumeActionParam_VOLUME_ACTION_PARAM_ON,
}

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
```
adityadani: nit: 
Attach: api.VolumeActionParam_VOLUME_ACTION_PARAM_ON},
should be
Attach: api.VolumeActionParam_VOLUME_ACTION_PARAM_ON,
}

fix at other places as well.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+			Attach:    api.VolumeActionParam_VOLUME_ACTION_PARAM_ON,
+			Mount:     api.VolumeActionParam_VOLUME_ACTION_PARAM_ON,
+			MountPath: "/mnt"},
+		Locator: &api.VolumeLocator{Name: name},
+		Spec:    &api.VolumeSpec{Size: size},
+	}
+
+	gomock.InOrder(
+
+		ts.MockDriver().
```
adityadani: If the Attach Action_PARAM is ON, shouldn't the Attach API get called ? I dont see any expect for that?
What I am more curious is how is the test passing?

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+			EXPECT().
+			Mount(id, gomock.Any(), gomock.Any()).
+			Return(nil),
+
```
adityadani: Also the Locator and Spec is not nil which means d.Set() should also be called and there needs to be an equivalent Expect()
https://github.com/libopenstorage/openstorage/blob/master/api/server/volume.go#L150

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+	res, err := driverclient.Stats(id, true)
+
+	assert.Nil(t, err)
+	assert.Equal(t, bytesUsed, res.BytesUsed)
+
+}
+
+// MockDriver helper method.
+func (s *testServer) MockDriver() *mockdriver.MockVolumeDriver {
```
lpabon: Place all `testServer` and its receivers in its own generic file.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+
+	if err != nil {
+		dlog.Errorf("Error starting the server")
+	}
+
+	// adding sleep to avoid race condition of connection refused.
+	time.Sleep(1 * time.Second)
+}
+
+func setupMocks() *testServer {
```
lpabon: Since this is returning a new *testServer, please rename to newMocks() instead.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+
+	req := &api.VolumeCreateRequest{
+		Locator: &api.VolumeLocator{Name: name},
+		Source:  &api.Source{},
+		Spec:    &api.VolumeSpec{Size: size},
+	}
+
+	// Setup mock functions
+	id := "myid"
+	gomock.InOrder(
```
lpabon: `InOrder` is only needed when there is more than one matched mock call. Not needed here.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+		ts.MockDriver().
+			EXPECT().
+			Create(gomock.Any(), gomock.Any(), gomock.Any()).
```
lpabon: Please check to see which values can be matched here if any. It provides more testing than allowing _any_ value. Gomock is very good at matching structs.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+
+	req := &api.VolumeCreateRequest{}
+
+	// create a volume client
+	driverclient := volumeclient.VolumeDriver(ts.client)
+
+	res, err := driverclient.Create(req.GetLocator(), req.GetSource(), req.GetSpec())
+
+	assert.NotNil(t, err)
+	assert.Empty(t, res)
```
lpabon: Check that err.Error() contains `error in create`

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+	}
+
+	gomock.InOrder(
+		ts.MockDriver().
+			EXPECT().
+			Detach(id, gomock.Any()).
+			Return(fmt.Errorf("Error in detaching")),
+	)
+
+	// create client
```
lpabon: Update these comments for all functions

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
+package testing
```
lpabon: This file should be under `api/client/volume` since it is testing that package.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,686 @@
...
+	mgmtPort   uint16 = 2376
+	pluginPort uint16 = 2377
+	driver     string = "mock"
+	version    string = "v1"
+)
+
+// Init function to setup the http server
+
+func init() {
+	startServer()
```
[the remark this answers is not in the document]
lpabon: If, for some reason, this cannot be accomplished with the way the code is written today, go ahead with this global server, but create an issue that the code needs to be enhanced to be able to be tested.

For example, the repo has many examples of a code pattern where global variables are used to house an instance of a variable. This makes it almost impossible to test. If you find something like this interferes with making `startServer` not global, then please create an issue to have that fix, and we'll allow the global implementation of `startServer` for now.

api/testing/restapi_test.go
ipochi: Ideally it should but Set method takes only the VolumeLocator and Spec , which is what is being passed to the server by the client wrapper method when making the http rest call. 
VolumeSetRequest.VolumeStateAction is not passed by the client wrapper hence processed as nil by server.

api/testing/restapi_test.go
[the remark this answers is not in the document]
ipochi: Ideally every Test should be independent starting and stopping the server.
However in the current functionality 

1. the startServer method in api/server/server.go returns only an err instance after creating the server.
https://github.com/libopenstorage/openstorage/blob/7b586013e8dba9004af52893b75c71853444c6e0/api/server/server.go#L130

2. There is no StopvServer functionality  in api/server/server.go

Hence the need for starting the server globally in the tests.

I will create a issue regarding this

api/testing/restapi_test.go
ipochi: Client wrapper method makes the call to doVolumeSet (in case of detach , mount and unmount) and dovolumeSetGetResponse( in case of attach)

which creates a VolumeSetRequest instance  with the Action attach and options but locator and spec are nil.
Hence on the server side its nil and no call is made to the Set.

api/testing/restapi_test.go
adityadani: @lpabon this test suite tests both client as well as server code. All the tests will invoke both http client and server with the driver as MockDriver. So does it not make sense for the tests to be under `apt/test`.

api/testing/restapi_test.go
```diff
@@ -0,0 +1,1528 @@
...
+		mgmtPort,
+	); err != nil {
+		dlog.Errorf("Error starting the server")
+	}
+
+	// adding sleep to avoid race condition of connection refused.
+	time.Sleep(1 * time.Second)
+}
+
+func newMocks() *testServer {
```
lpabon: This works, but it would be better to place this in testServer.go as a `newTestServer() *testServer` function.

adityadani:
overall looks good to me.

PR #258, opened 2017-12-22
 Add GCE storage operations implementation

harsh-px:


pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,552 @@
...
+
+	return &gceOps{
+		inst:    i,
+		service: service,
+	}, nil
+}
+
+func (s *gceOps) waitStatus(id string, desired string) error {
+	actual := ""
+	for retries, maxRetries := 0, 10; actual != desired && retries < maxRetries; retries++ {
```
adityadani: nit: constant for max retries. If possible same for both AWS and GCE.

adityadani: Also for all sleep intervals. Looks like they are common between AWS and GCE.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,552 @@
...
+		logrus.Warnf("Rollback failed volume %v, Error %v", id, err)
+	}
+	return createErr
+}
+
+func (s *gceOps) available(v *compute.Disk) bool {
+	return v.Status == "READY"
+}
+
+func (s *gceOps) addResource(
```
adityadani: this function can be common between AWS and GCE

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,552 @@
...
+) {
+	if s, ok := sets[key]; ok {
+		sets[key] = append(s, disk)
+	} else {
+		sets[key] = make([]interface{}, 0)
+		sets[key] = append(sets[key], disk)
+	}
+}
+
+func (s *gceOps) ApplyTags(diskName string,
```
adityadani: nit: ApplyTags(
diskName string,
labels map[string]string,
) (err error)

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,552 @@
...
+		time.Sleep(interval)
+	}
+
+	err = fmt.Errorf("disk: %s is not attached to instance: %s", disk.Name, s.inst.Name)
+	return
+}
+
+func (s *gceOps) Name() string { return "gce" }
+
+func (s *gceOps) describe() (*compute.Instance, error) {
```
adityadani: nit: rename it to describeInstance()

pkg/storageops/gce/gce.go
harsh-px: Fixed in incremental

pkg/storageops/gce/gce.go
harsh-px: Fixed in incremental

pkg/storageops/gce/gce.go
harsh-px: Fixed in incremental

pkg/storageops/gce/gce.go
harsh-px: Fixed in incremental

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,552 @@
...
+
+type gceOps struct {
+	inst    *Instance
+	service *compute.Service
+	mutex   sync.Mutex
+}
+
+// Instance stores the metadata of the running GCE instance
+type Instance struct {
+	ID         string
```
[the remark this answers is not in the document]
harsh-px: Fixed in incremental

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+		return
+	}
+
+	err = fmt.Errorf("env variable %s is not set", key)
+	return
+}
+
+// gceInfo fetches the GCE instance metadata from the metadata server
+func gceInfo(inst *instance) {
+	a := &assigner{}
```
lpabon: This probably should not be a single instance. `assigner` overwrites the error returned if two or more errors were saved. This causes the overwritten errors to be lost.

Is there a need for `assign`? Could we remove it? Maybe replace the whole thing with a for loop and struct of the variable to save to and the function to get the information from.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+
+	return &gceOps{
+		inst:    i,
+		service: service,
+	}, nil
+}
+
+func (s *gceOps) waitStatus(id string, desired string) error {
+	actual := ""
+	for retries, maxRetries := 0, storageops.ProviderOpsMaxRetries; actual != desired && retries < maxRetries; retries++ {
```
lpabon: Take a look at https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go . We should instead have a `util` pkg in osd, so that it can be used by other packages.

lpabon: This does not have to be done for this PR. If you want, please create an issue that solves this and add it here.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+		d, err := s.service.Disks.Get(s.inst.Project, s.inst.Zone, id).Do()
+		if err != nil {
+			if ignore := notFoundRegex.MatchString(err.Error()); ignore {
+				time.Sleep(storageops.ProviderOpsRetryInterval)
```
lpabon: Since this is the same as 162, you can try to combine the logic. One way is to use `goto` , but I am sure there are others.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+func (s *gceOps) waitForDetach(
+	diskURL string,
+	timeout time.Duration,
+) (err error) {
+	var inst *compute.Instance
+
+	interval := 5 * time.Second
+	for elapsed := 0 * time.Second; elapsed < timeout; elapsed += interval {
```
lpabon: This would also benefit from the wait.Poll() function.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+		inst, err = s.describeinstance()
+		if err != nil {
+			return
```
lpabon: Why change the style here to set `err` and just return? It is not used in any other function that way here

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+				found = true
+				break
+			}
+		}
+
+		if found {
+			time.Sleep(interval)
+			continue
+		} else {
+			return
```
lpabon: The value of `err` is not explicitly being set here. Just need to be careful with something else setting it before returning. I would suggest using `return <value>`

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+	return
+}
+
+// waitForAttach checks if given disk is attached to the local instance
+func (s *gceOps) waitForAttach(
+	disk *compute.Disk,
+	timeout time.Duration,
+) (devicePath string, err error) {
+	interval := 2 * time.Second
+	for elapsed := 0 * time.Second; elapsed < timeout; elapsed += interval {
```
lpabon: another use of wait.Poll()

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+		devicePath, err = s.DevicePath(disk.Name)
+		if err == nil {
+			return
+		}
+
+		time.Sleep(interval)
+	}
+
+	err = fmt.Errorf("disk: %s is not attached to instance: %s", disk.Name, s.inst.Name)
+	return
```
lpabon: Please use explicit returns unless absolutely necessary.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+func (s *gceOps) Name() string { return "gce" }
```
lpabon: Do you mind grouping the exposed functions?

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+	logrus.Warnf("Rollback create volume %v, Error %v", id, createErr)
+	err := s.Delete(id)
+	if err != nil {
+		logrus.Warnf("Rollback failed volume %v, Error %v", id, err)
+	}
+	return createErr
+}
+
+func (s *gceOps) available(v *compute.Disk) bool {
+	return v.Status == "READY"
```
lpabon: Make `READY` a const. Does GCE have a const for this already?

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+func (s *gceOps) ApplyTags(
+	diskName string,
+	labels map[string]string) (err error) {
+	d, err := s.service.Disks.Get(s.inst.Project, s.inst.Zone, diskName).Do()
+	if err != nil {
```
lpabon: I'm not 100% sure of this API, but if it is a regular RESTful one, there could be HTTP status available which needs to be used or logged.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+	for k, v := range labels {
+		currentLabels[k] = v
+	}
+
+	rb := &compute.ZoneSetLabelsRequest{
+		LabelFingerprint: d.LabelFingerprint,
+		Labels:           currentLabels,
+	}
+
+	_, err = s.service.Disks.SetLabels(s.inst.Project, s.inst.Zone, d.Name, rb).Do()
```
lpabon: Check err and any HTTP status if any.

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+	return
```
lpabon: Return explicit error

pkg/storageops/gce/gce.go
```diff
@@ -0,0 +1,540 @@
...
+}
+
+func (s *gceOps) Snapshot(
+	disk string,
+	readonly bool,
+) (interface{}, error) {
+	rb := &compute.Snapshot{
+		Name: fmt.Sprintf("snap-%d%02d%02d", time.Now().Year(), time.Now().Month(), time.Now().Day()),
+	}
+	return s.service.Disks.CreateSnapshot(s.inst.Project, s.inst.Zone, disk, rb).Do()
```
lpabon: How do the snapshots get deleted?

pkg/storageops/gce/gce_test.go
```diff
@@ -0,0 +1,53 @@
...
+		diskTemplates := make(map[string]map[string]interface{})
+
+		d, disks := initGCE(t)
+		drivers[d.Name()] = d
+		diskTemplates[d.Name()] = disks
+
+		test.RunTest(drivers, diskTemplates, t)
+	} else {
+		fmt.Printf("skipping GCE tests as environment is not set...\n")
+		t.Skip("skipping GCE tests as environment is not set...")
```
lpabon: Is there a test for snapshots?

harsh-px: Added snapshots tests.

pkg/storageops/gce/gce.go
harsh-px: Added a new api.

pkg/storageops/gce/gce.go
harsh-px: Removed named returns from all files.

pkg/storageops/gce/gce.go
harsh-px: As discussed offline, this won't be necessary and the API internally does it.

pkg/storageops/gce/gce.go
harsh-px: GCE doesn't have a constant. Added an internal const.

pkg/storageops/gce/gce.go
harsh-px: done

pkg/storageops/gce/gce.go
harsh-px: done

pkg/storageops/gce/gce.go
harsh-px: We already had https://github.com/portworx/sched-ops/blob/master/task/task.go#L15

So I have now changed all retry loops to use this.

pkg/storageops/gce/gce.go
harsh-px: assigner doesn't override errors as the assign method checks for existing errors before invoking the next function.

```
func (a *assigner) assign(getVal func() (string, error)) string {
	if a.err != nil {
		return ""
	}
```

harsh-px:
@lpabon I have updated the review. Let me know if you have more questions.

pkg/storageops/gce/gce.go
lpabon: Correct, but the value of `a.err` is lost and instead an empty string is returned. It should at least log the error before returning and if it logs, it should log the name it was trying to assign.

harsh-px:
@lpabon removed the assigner pattern.

PR #260, opened 2018-01-02
csi: Update identity name

### LINKED ISSUE #245: GetPluginInfoResponse needs to be updated
```
message GetPluginInfoResponse {
  // The name MUST follow reverse domain name notation format
  // (https://en.wikipedia.org/wiki/Reverse_domain_name_notation).
  // It SHOULD include the plugin's host company name and the plugin
  // name, to minimize the possibility of collisions. It MUST be 63
  // characters or less, beginning and ending with an alphanumeric
  // character ([a-z0-9A-Z]) with dashes (-), underscores (_),
  // dots (.), and alphanumerics between. This field is REQUIRED.
  string name = 1;
```

lpabon:
Closes #245

### DIFF
--- a/csi/identity.go
+++ b/csi/identity.go
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	csiDriverVersion = "0.1.0"
-	csiDriverName    = "csi-osd"
+	csiDriverVersion    = "0.1.0"
+	csiDriverNamePrefix = "com.openstorage."
 )
 
 var (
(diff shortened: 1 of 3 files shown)


csi/identity.go
```diff
@@ -22,8 +22,8 @@ import (
 )
 
 const (
-	csiDriverVersion = "0.1.0"
-	csiDriverName    = "csi-osd"
+	csiDriverVersion    = "0.1.0"
+	csiDriverNamePrefix = "com.portworx.openstorage."
```
adityadani: The prefix does not need to have portworx in it? CSI/osd should be driver agnostic.
It could simply be "com.openstorage."

lpabon: Great point. I will change that.

PR #268, opened 2018-01-04
 Add option 'journal' to spec. If

jvinod:
Add option 'journal' to spec. If set to true data for this device will be journalled. The default value is false. Unset value from old versions will translate to false.


lpabon:
I think the commit message of the patch is messed up. The first list has a maximum of 50 if I recall.

jvinod:
I don't think we need to follow the numbering scheme - the protobuf doesn't
go over the wire..

PR #273, opened 2018-01-06
Support for bind mount of fuse mounts.

talakad:
Bind mount of fuse mounts show up with fuse name  as device rather than device path being bind mounted. Added mechanism to pass in device name for fuse mounts. 

To elaborate, Mount function is passed with a device parameter and a path to mount the device with. In case of bind mount, both device and mount path are simply paths. After the bind mount, both of them will show with filesystem/device  that is on source path.  So mounter needs to track the mounts with this filesystem "string" passed in the options for bind mounts to reflect the correct mount status.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -8,6 +8,10 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/libopenstorage/openstorage/pkg/keylock"
+	"github.com/libopenstorage/openstorage/pkg/options"
+	"github.com/libopenstorage/openstorage/pkg/sched"
+	"go.pedge.io/dlog"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -16,11 +20,6 @@ import (
 	"sync"
 	"syscall"
 	"time"
-
-	"github.com/libopenstorage/openstorage/pkg/keylock"
-	"github.com/libopenstorage/openstorage/pkg/options"
-	"github.com/libopenstorage/openstorage/pkg/sched"
-	"go.pedge.io/dlog"
 )
 
 // Manager defines the interface for keep track of volume driver mounts.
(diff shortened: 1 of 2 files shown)


pkg/mount/mount.go
```diff
@@ -323,12 +322,17 @@ func (m *Mounter) deletePath(path string) bool {
...
 func (m *Mounter) Mount(
 	minor int,
-	device, path, fs string,
+	devPath, path, fs string,
 	flags uintptr,
 	data string,
 	timeout int,
 	opts map[string]string,
 ) error {
+	device := devPath
```
lpabon: Could you add a comment here that device is initialized to devPath, but could be overwritten by opts[]?

pkg/mount/mount.go
```diff
@@ -323,12 +322,20 @@ func (m *Mounter) deletePath(path string) bool {
...
 	flags uintptr,
 	data string,
 	timeout int,
 	opts map[string]string,
 ) error {
+	// device gets overwritten if opts specifies fuse mount with
+	// options.OptionsDeviceFuseMount.
+	device := devPath
+	if value, ok := opts[options.OptionsDeviceFuseMount]; ok {
+		// fuse mounts show-up with with this key as device.
```
lpabon: 'with with'

pkg/options/options.go
```diff
@@ -17,6 +17,8 @@ const (
 	OptionsWaitBeforeDelete = "WAIT_BEFORE_DELETE"
 	// OptionsRedirectDetach Redirect detach to the node where volume is attached
 	OptionsRedirectDetach = "REDIRECT_DETACH"
+	// OptionsDeviceFuseMount name of fuse mount device
+	OptionsDeviceFuseMount = "DEV_FUSE_MOUNT"
```
disrani-px: Fuse for shared volumes is a px specific implementation. libopenstorage doesn't know about it, right? So should this be called DeviceVirtualMount (or something similar) instead?

pkg/mount/mount.go
```diff
@@ -8,6 +8,10 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/libopenstorage/openstorage/pkg/keylock"
```
disrani-px: This was fine before, separated from the std imports

PR #276, opened 2018-01-12
test: Test nodeStatus

lpabon:
With this change it will now be possible to unit test Cluster manager restful endpoints.  This patter can also be used for testing other singleton object receivers.

Authors of this work:
lpabon, ipochi

### DIFF
--- a/api/server/cluster_test.go
+++ b/api/server/cluster_test.go
@@ -1,0 +1,80 @@
+package server
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/libopenstorage/openstorage/api"
+	client "github.com/libopenstorage/openstorage/api/client/cluster"
+	"github.com/libopenstorage/openstorage/cluster"
+	mockcluster "github.com/libopenstorage/openstorage/cluster/mock"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/golang/mock/gomock"
+	"github.com/kubernetes-csi/csi-test/utils"
+)
+
+type testCluster struct {
+	c       *mockcluster.MockCluster
+	mc      *gomock.Controller
+	oldInst func() (cluster.Cluster, error)
+}
+
+func newTestClutser(t *testing.T) *testCluster {
+	tester := &testCluster{}
+
+	// Save already set value of cluster.Inst to set it back
+	// when we finish the tests by the defer()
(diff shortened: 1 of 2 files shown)


api/server/cluster_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	// Save already set value of cluster.Inst to set it back
+	// when we finish the tests by the defer()
+	oldInst := cluster.Inst
+	defer func() {
+		cluster.Inst = oldInst
+	}()
+
+	// Create mock cluster
+	mc := gomock.NewController(&utils.SafeGoroutineTester{})
```
adityadani: Can you separate out the setting of the instance section of the code into a function? All the tests are going to need it.

api/server/cluster_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"github.com/golang/mock/gomock"
+	"github.com/kubernetes-csi/csi-test/utils"
+)
+
+func TestServerNodeStatus(t *testing.T) {
+
+	// Save already set value of cluster.Inst to set it back
+	// when we finish the tests by the defer()
+	oldInst := cluster.Inst
```
adityadani: If running in a test context, why would you need this? You can always override the cluster.Inst function definition.

cluster/cluster.go
```diff
@@ -17,6 +17,12 @@ var (
 
 	errClusterInitialized    = errors.New("openstorage.cluster: already initialized")
 	errClusterNotInitialized = errors.New("openstorage.cluster: not initialized")
+
+	// Inst returns an instance of an already instantiated cluster manager.
+	// This function can be overritten for testing purposes
```
adityadani: Why not just add Set and Get for the global "inst" object? It is more intuitive than setting the Inst function object.

api/server/cluster_test.go
lpabon: We need this, because you should always place tests back to a clean state before running the the next test. The next one may want to use the real Cluster.

cluster/cluster.go
lpabon: Since it is just for testing and it is a global variable, I didn't see a need.

api/server/cluster_test.go
lpabon: Sure I'll take care of it.

PR #290, opened 2018-01-19
Add operation flags definition

prabirpaul:
Adds new option/flags definition to be used when requesting forceful device detach.


adityadani:
lgtm. 
You need to sign of your commit. Hence the build failed. 
`git commit -s` should do the trick.

PR #292, opened 2018-01-23
sanity: GetGossipState SetSize

lpabon:

Added two new tests to osd-sanity

Worked with Imran on these tests.


lpabon:
Good catch, thanks @sangleganesh

PR #294, opened 2018-01-24
osdconfig package implementation

sdeoras:
This PR is a step towards implementing osd config data in a backend service and not local json file

Hi Folks, thanks for taking a look at code. Pl. only focus on libopenstorage/openstorage/osdconfig folder.
kvdb packaged in vendor dir is old and does not allow in-mem kvdb functionality, so I am unable to test that interface. Also I would like to discuss with you on semantics of kvdp api and parameters in osdconfig struct.


sdeoras:
[meeting Jan/25] we decided to break this down into smaller parts, hence closing this one and will create a new one.

PR #295, opened 2018-01-25
osdconfig: proto def review

sdeoras:
This is a step towards putting config params behind an abstraction/service

Closes #294 

As discussed in Jan/25th meeting, this PR is just for proto def review.


osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
+syntax = "proto3";
+
+// this package defines the configuration parameters and interfaces
+// requires to run portwork services
```
sangleganesh: change should not be specific to portworx as such.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+    string created = 4;
+    string cluster_id = 5;
+    string logging_url = 6;
+    string alerting_url = 7;
+    string scheduler = 8;
+    bool multicontainer = 9;
+    bool nolh = 10;
+    bool callhome = 11;
+    bool bootstrap = 12;
+    string tunnel_end_point = 13;
```
sangleganesh: afaik tunnel, nolh, logging/alert url are gone. 
callhome is not osd config param.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+}
+
+message StorageConfig {
+    repeated string devices_md = 1;
+    int32 max_count = 2;
+    int32 max_drive_set_count = 3;
+    repeated string devices = 4;
+    string raid_level = 5;
+    string raid_level_md = 6;
+    bool async_io = 7;
```
sangleganesh: async_io, num_theads are not related to osd.

osdconfig/proto/config.proto
sdeoras: thanks @sangleganesh , will do

osdconfig/proto/config.proto
sdeoras: will confirm by Tue

sdeoras:
@sangleganesh, we are still discussing design, so I don't have a whole lot to describe at the moment. Basically we are waiting to settle on struct definitions and RPC calls first. To the end user it will get served via a higher level abstraction.... Here are some thoughts but this will likely change as I get more inputs: https://github.com/sdeoras/openstorage/tree/sd-dev-bkp/osdconfig

api/client/volume/client.go
```diff
@@ -4,12 +4,13 @@ import (
...
-	"github.com/libopenstorage/openstorage/api"
-	"github.com/libopenstorage/openstorage/api/client"
-	"github.com/libopenstorage/openstorage/volume"
 	"io"
 	"io/ioutil"
 	"strconv"
+
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/libopenstorage/openstorage/api/client"
+	"github.com/libopenstorage/openstorage/volume"
```
lpabon: Remove this file from the PR

api/client/volume/client_volume_test.go
```diff
@@ -3,11 +3,12 @@ package volume
...
 	"encoding/json"
-	"github.com/libopenstorage/openstorage/api"
-	"github.com/stretchr/testify/require"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/stretchr/testify/require"
 )
```
lpabon: Remove this file from the PR

api/client/volume/volume.go
```diff
@@ -2,6 +2,7 @@ package volume
 
 import (
 	"fmt"
+
 	"github.com/libopenstorage/openstorage/api"
```
lpabon: Remove this file from the PR

lpabon:
Also, please squash all your commits since there are too many.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+}
+
+message NodesConfig {
+    map<string, NodeConfig> node_conf = 1;
+}
+
+message NodeConfig {
+    string node_id = 1;
+    NetworkConfig network = 2;
+    SecretsConfig secrets = 3;
```
adityadani: SecretsConfig should be at cluster level.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+    bool multicontainer = 9;
+    bool nolh = 10;
+    bool callhome = 11;
+    bool bootstrap = 12;
+    string tunnel_end_point = 13;
+    repeated string tunnel_certs = 14;
+    string driver = 15;
+    string debug_level = 16;
+    string domain = 17;
+    string mgmtip = 18;
```
adityadani: mgmtip and dataip should be node specific. You already have it as a part of NetworkConfig as mgt_iface and data_iface

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
+syntax = "proto3";
+
+// this package defines the configuration parameters and interfaces
+// requires to run portwork services
+package proto;
+
+service Spec {
+    rpc GetGlobalSpec(Empty) returns (GlobalConfig) {}
```
adityadani: Can you add comments for each RPC?

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message NodeConfig {
+    string node_id = 1;
+    NetworkConfig network = 2;
+    SecretsConfig secrets = 3;
+    StorageConfig storage = 4;
+}
+
+message KVDBConfig {
+    string username = 1;
```
adityadani: Can you also add comments for each field so that it would show up in the swagger docs? I can help you with that.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+    repeated string devices = 4;
+    string raid_level = 5;
+    string raid_level_md = 6;
+    bool async_io = 7;
+    int32 num_threads = 8;
+}
+
+
+
+
```
adityadani: nit: remove blank lines.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+    ClusterConfig cluster_conf = 1;
+    NodesConfig nodes_conf = 2;
+}
+
+message NodesConfig {
+    map<string, NodeConfig> node_conf = 1;
+}
+
+message NodeConfig {
+    string node_id = 1;
```
piyush-nimbalkar: what is the difference between this node_id and the message NodeID? why not use the same here or the string in RPC?

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+    KVDBConfig kvdb = 20;
+}
+
+message NetworkConfig {
+    string mgt_iface = 1;
+    string data_iface = 2;
+}
+
+message SecretsConfig {
+    string secret_type = 1;
```
piyush-nimbalkar: @adityadani should we use an enum here? or will it break backward compatibility?

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+// requires to run portwork services
+package proto;
+
+service Spec {
+    rpc GetGlobalSpec(Empty) returns (GlobalConfig) {}
+    rpc SetGlobalSpec(GlobalConfig) returns (Ack) {}
+    rpc GetClusterSpec(Empty) returns (ClusterConfig) {};
+    rpc SetClusterSpec(ClusterConfig) returns (Ack) {};
+    rpc GetNodeSpec(NodeID) returns (NodeConfig) {};
+    rpc SetNodeSpec(NodeConfig) returns (Ack) {};
```
piyush-nimbalkar: nit: not sure if need the semicolon at the end

osdconfig/proto/config.proto
piyush-nimbalkar: nit: Should this have been node_i_d to be consistent with the golint?
Also, I am not sure how i_d works for other languages, just in case some one decides to have the Spec service implemented in another language.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+}
+
+message NodeConfig {
+    string node_id = 1;
+    NetworkConfig network = 2;
+    SecretsConfig secrets = 3;
+    StorageConfig storage = 4;
+}
+
+message KVDBConfig {
```
piyush-nimbalkar: nit: Should we call this KvdbConfig because we use Kvdb in the kvdb repo?

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+service Spec {
+    rpc GetGlobalSpec(Empty) returns (GlobalConfig) {}
+    rpc SetGlobalSpec(GlobalConfig) returns (Ack) {}
+    rpc GetClusterSpec(Empty) returns (ClusterConfig) {};
+    rpc SetClusterSpec(ClusterConfig) returns (Ack) {};
+    rpc GetNodeSpec(NodeID) returns (NodeConfig) {};
+    rpc SetNodeSpec(NodeConfig) returns (Ack) {};
+}
+
+message Empty {}
```
piyush-nimbalkar: Is it mandatory to pass a message in the rpc? If yes, should we create one object for every rpc like,
GetClusterSpecRequest

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+message Ack {
+    int64 n = 1;
+}
+message NodeID {
```
piyush-nimbalkar: nit: newlines between these messages and for Empty we can write it - 
message Empty {
}

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
+syntax = "proto3";
+
+// this package defines the configuration parameters and interfaces
+// requires to run portwork services
+package proto;
+
+service Spec {
```
piyush-nimbalkar: nit: SpecService or ConfigService? Spec sounds more like a message.

osdconfig/proto/config.proto
```diff
@@ -0,0 +1,120 @@
...
+    rpc SetGlobalSpec(GlobalConfig) returns (Ack) {}
+    rpc GetClusterSpec(Empty) returns (ClusterConfig) {};
+    rpc SetClusterSpec(ClusterConfig) returns (Ack) {};
+    rpc GetNodeSpec(NodeID) returns (NodeConfig) {};
+    rpc SetNodeSpec(NodeConfig) returns (Ack) {};
+}
+
+message Empty {}
+message Ack {
+    int64 n = 1;
```
piyush-nimbalkar: nit: Should we rename 'n' to something more readable or have error/return codes for the ack responses. Just suggesting, I am fine with what the mass decides.

sdeoras:
@lpabon , @adityadani , @piyush-nimbalkar and @sangleganesh : Let's discuss tomorrow if your schedules permit. I'll setup a meeting. Thanks.

api/client/volume/client.go
sdeoras: done, will send new PR today if I can close with everyone on their requests

api/client/volume/client_volume_test.go
sdeoras: done. will send out new PR shortly

api/client/volume/volume.go
sdeoras: done, will send out new PR shortly.

osdconfig/proto/config.proto
sdeoras: done, will send out new PR shortly

osdconfig/proto/config.proto
sdeoras: ok, thanks. will move it at cluster level.

osdconfig/proto/config.proto
sdeoras: ok, deleted them from cluster level struct

osdconfig/proto/config.proto
sdeoras: :) sure, kept one at the end. been bitten in the past when code files did not end with a newline char

osdconfig/proto/config.proto
sdeoras: I am okay either way... toss.

osdconfig/proto/config.proto
sdeoras: both are okay, but let's remove them. done.

osdconfig/proto/config.proto
sdeoras: yes, that is mandatory to pass a message type... would like to know more why you think additional messages need to be defined. I tried to keep the number of messages to a minimum possible.

osdconfig/proto/config.proto
sdeoras: i got 'n' from typical Go's write() methods that return number of bytes written. Yes, some description on 'n' would help as it is not evident by itself as to what it does. Also okay to changing it to something else. Let's discuss today.

osdconfig/proto/config.proto
sdeoras: sure... now only if they had made gofmt for proto files as well we would be saving ourselves some review time :)

osdconfig/proto/config.proto
sdeoras: good question. there should just be one, but i needed a way to address node config using keys (nodeID's) and setting node config data by passing the struct. Also higher level code abstracts it out, so it is seamless as follows:
func (c *specClientIO) SetNodeSpec(ctx context.Context, in *NodeConfig) (*Ack, error) {
        // set node spec
	globalConf.NodesConf.NodeConf[in.NodeId] = in
}

sdeoras: good point. node_i_d will become nodeID in Go. let's make a call on it. i am okay both ways.

osdconfig/proto/config.proto
sdeoras: ok, you know the convention here best. done.

osdconfig/proto/config.proto
sdeoras: @piyush-nimbalkar, i also recall there were secrets specific changes you wanted to make. is this the only one?

sdeoras:
from @piyush-nimbalkar:
>> Personally I feel, we should have comment for every field considering proto is 'the' schema that everyone will refer to.

i need some help here. pl. feel free to add comments if you know what these fields do.

Also discussing with @sangleganesh yesterday it seems we need rpc calls to delete a node config. I am thinking we should add two more rpc calls as follows:
// get all node IDs
rpc GetNodeIDs(Empty) returns (NodeIDs) {}
    
// delete a set of nodes configs
rpc DeleteNodes(NodeIDs) returns (Ack) {}

sdeoras:
folks, here are some of the changes you requested: https://github.com/sdeoras/openstorage/blob/sd-dev-rev-1/osdconfig/proto/config.proto 
I can create new PR for this and close this one.

lpabon:
@sdeoras Sure, that would be fine, thanks!

sdeoras:
Thanks all for review and comments here. I did incorporate as many of these changes as I could in the struct definition. Pl. stay tuned for another PR after I close with @jvinod. 
cc: @venkatpx

PR #296, opened 2018-01-26
CSI Sanity tests and fixes

lpabon:
With the new csi-sanity tool and pkg we can now test the CSI APIs act as expected. This PR fixes issues found by csi-sanity and also adds it to the CI system to be used on every `docker-test` run.

* Added support to run CSI drivers from `cmd/osd/main.go`
* Added new `csi/csisanity_test.go` which may not be perfectly setting up NFS as the server. Please make sure I did the right thing.

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -19,6 +19,8 @@ package csi
 import (
 	"fmt"
 
+	"github.com/portworx/kvdb"
+
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/pkg/util"
 
@@ -34,6 +36,7 @@ const (
 	volumeCapabilityMessageNotMultinodeVolume = "Volume is not a multinode volume"
 	volumeCapabilityMessageReadOnlyVolume     = "Volume is read only"
 	volumeCapabilityMessageNotReadOnlyVolume  = "Volume is not read only"
+	defaultCSIVolumeSize                      = uint64(1024 * 1024 * 1024)
 )
 
 // ControllerGetCapabilities is a CSI API functions which returns to the caller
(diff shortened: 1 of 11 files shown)


csi/controller.go
```diff
@@ -34,6 +34,7 @@ const (
 	volumeCapabilityMessageNotMultinodeVolume = "Volume is not a multinode volume"
 	volumeCapabilityMessageReadOnlyVolume     = "Volume is read only"
 	volumeCapabilityMessageNotReadOnlyVolume  = "Volume is not read only"
+	defaultCSIVolumeSize                      = uint64(1024 * 1024 * 1024)
```
piyush-nimbalkar: we do have a default volume size in OSD right?
should we just use that constant here instead of declaring in two places?

csi/controller.go
```diff
@@ -433,13 +439,17 @@ func (s *OsdCsiServer) DeleteVolume(
...
 
-	err := s.driver.Delete(req.GetVolumeId())
-	if err != nil {
-		e := fmt.Sprintf("Unable to delete volume with id %s: %s",
-			req.GetVolumeId(),
-			err.Error())
-		dlog.Errorln(e)
-		return nil, status.Error(codes.Internal, e)
+	// If the volume is not found, then we can return OK
+	volumes, err := s.driver.Inspect([]string{req.GetVolumeId()})
```
piyush-nimbalkar: what if it is a not a "not found" error and something else?

csi/controller.go
lpabon: Yeah, absolutely. Do you know where that is?

csi/controller.go
lpabon: good point. I'll take a look to see if Inspect returns a known error code. If not, we may not be able to determine and may have to guess.

csi/controller.go
piyush-nimbalkar: Tried looking for it, couldn't find it so far.
@adityadani would have a better idea

lpabon: As discussed in email, we need a value in OSD. I think this is the right place since only CSI and not OSD supports default size.

csi/controller.go
lpabon: It doesn't look as if kvdb will guarantee the value of kvdb.ErrNotFound for all backends. Instead the GetVal implementation seems to return whatever error is returned from the interface implementation. I only see the memory kvdb backend use kvdb.ErrNotFound.


cc @adityadani

PR #297, opened 2018-01-29
Volume create,enumerate API osd-sanity

ipochi:

This PR works on the osd-sanity tests for Volume API's
Closes #

### DIFF
--- a/pkg/sanity/volume.go
+++ b/pkg/sanity/volume.go
@@ -37,10 +37,9 @@ var _ = Describe("Volume [Volume Tests]", func() {
 
 	BeforeEach(func() {
 		var err error
-		restClient, err = volumeclient.NewDriverClient(osdAddress, "nfs", volume.APIVersion, "")
+		restClient, err = volumeclient.NewDriverClient(osdAddress, volumeDriver, volume.APIVersion, "")
 
 		Expect(err).ToNot(HaveOccurred())
-		//volumedriver = clusterclient.ClusterManager(restClient)
 		volumedriver = volumeclient.VolumeDriver(restClient)
 	})
 
@@ -51,15 +50,34 @@ var _ = Describe("Volume [Volume Tests]", func() {
 	Describe("Volume Create", func() {
 
 		var (
-			volumerequest *api.VolumeCreateRequest
-			volumeID      string
+			volumeID         string
+			numVolumesBefore int
+			numVolumesAfter  int
 		)
 
 		BeforeEach(func() {
+			var err error
+			volumes, err := volumedriver.Enumerate(&api.VolumeLocator{}, nil)
+			numVolumesBefore = len(volumes)
+			Expect(err).NotTo(HaveOccurred())
+		})
 
+		AfterEach(func() {
+			var err error
+			err = volumedriver.Delete(volumeID)
+			Expect(err).ToNot(HaveOccurred())
+
+			volumes, err := volumedriver.Enumerate(&api.VolumeLocator{}, nil)
+			Expect(err).ToNot(HaveOccurred())
+			numVolumesAfter = len(volumes)
+		})
+
+		It("Should create a volume successfully", func() {
+
+			By("Creating the volume")
 			var err error
 
-			volumerequest = &api.VolumeCreateRequest{
+			vr := &api.VolumeCreateRequest{
 				Locator: &api.VolumeLocator{
 					Name: "vol-osd-sanity-cd",
 					VolumeLabels: map[string]string{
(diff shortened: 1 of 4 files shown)


pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+	BeforeEach(func() {
+		var err error
+		restClient, err = volumeclient.NewDriverClient(osdAddress, "nfs", volume.APIVersion, "")
+
+		Expect(err).ToNot(HaveOccurred())
+		//volumedriver = clusterclient.ClusterManager(restClient)
+		volumedriver = volumeclient.VolumeDriver(restClient)
+	})
+
+	AfterEach(func() {
```
lpabon: Remove if not needed.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+
+var _ = Describe("Volume [Volume Tests]", func() {
+	var (
+		restClient   *client.Client
+		volumedriver volume.VolumeDriver
+	)
+
+	BeforeEach(func() {
+		var err error
+		restClient, err = volumeclient.NewDriverClient(osdAddress, "nfs", volume.APIVersion, "")
```
lpabon: You cannot specify here the driver. It must be specified by the caller. If a driver needs to be passed in, adjust sanity.go to take in a driver name.

The reason is that this would make this program only test NFS, while instead we want it to test the specific implementation running.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+					},
+				},
+				Source: &api.Source{},
+				Spec: &api.VolumeSpec{
+					Size:   1,
+					Format: api.FSType_FS_TYPE_NFS,
+				},
+			}
+
+			volumeID, err = volumedriver.Create(volumerequest.GetLocator(), volumerequest.GetSource(), volumerequest.GetSpec())
```
lpabon: Normally the action to be tested is to be included in the It() test. For example, sometimes you may want to pass in correct arguments, and sometimes you may want to pass in incorrect arguments expecting failure.

But it is ok to do this _if_ you there is no negative tests.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+		AfterEach(func() {
+			err := volumedriver.Delete(volumeID)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		It("Volume should be present in inspect", func() {
+
+			inspectVolumes := []string{volumeID}
+
+			By("only one volume to be returned and having the same name as the created volume in beforeeach.")
```
lpabon: For this to be true, the test should confirm that there are no volumes already created, which could be complicated. This is why creating the volume in the It() would be easier, specially for this algorithm:

* Save the number of volumes already in the cluster.
* Create volume
* Expect that now there is 1 more volume
* Get information on volume
* Delete volume

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+			volumesList, err := volumedriver.Inspect(inspectVolumes)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(volumesList).NotTo(BeEmpty())
+			Expect(len(volumesList)).Should(BeEquivalentTo(1))
```
lpabon: Instead of using "Should", use "To(Equal(1+numberofvolumes))"

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+			inspectVolumes := []string{volumeID}
+
+			By("inspecting the details of the volume")
+			volumesList, err := volumedriver.Inspect(inspectVolumes)
+
+			volume := volumesList[0]
+
+			Expect(err).NotTo(HaveOccurred())
+			Expect(volume.GetSpec().GetSize()).To(BeEquivalentTo(1))
+			Expect(volume.GetSpec().GetFormat()).To(BeEquivalentTo(api.FSType_FS_TYPE_NFS))
```
lpabon: This api.FS_TYPE value must be adjusted to be the value of the specified driver implementation running.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+
+		})
+
+		It("should say volume already exists", func() {
+			By("Creating the volume with same name again")
+
+			_, err := volumedriver.Create(volumerequest.GetLocator(), volumerequest.GetSource(), volumerequest.GetSpec())
```
lpabon: This line seems to be too long.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).Should(ContainSubstring("already exists"))
```
lpabon: I'm not sure what "Should" does or how it is different to `To`, but could you change it to `To`?

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+						Name: "vol-osd-sanity-ei-" + strconv.Itoa(i),
+						VolumeLabels: map[string]string{
+							"class": "sanity-test-class",
+						},
+					},
+					Source: &api.Source{},
+					Spec:   &api.VolumeSpec{},
+				}
+
+				volumeID, err := volumedriver.Create(volumerequest.GetLocator(), volumerequest.GetSource(), volumerequest.GetSpec())
```
lpabon: Line is too long.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+
+		It("Should enumerate all created volumes", func() {
+
+			By("All five volumes to be listed")
+
+			volumesList, err := volumedriver.Enumerate(&api.VolumeLocator{}, make(map[string]string))
+			Expect(err).NotTo(HaveOccurred())
+
+			Expect(volumesList).NotTo(BeEmpty())
+			Expect(len(volumesList)).Should(BeEquivalentTo(5))
```
lpabon: Cannot be guaranteed. Just like above, the number of volume already in the cluster must first be saved, which could be done in the BeforeEach.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,194 @@
...
+		})
+
+		It("Should enumeate only one volume", func() {
+			By("Passing volume name in Volume Locator")
```
lpabon: Just one? This should be tested for all 5 volumes.

pkg/sanity/volume.go
lpabon: You can also save the current number of volumes in the cluster on BeforeEach before calling volume create.

pkg/sanity/volume.go
```diff
@@ -0,0 +1,341 @@
...
+
+		Expect(err).ToNot(HaveOccurred())
+		volumedriver = volumeclient.VolumeDriver(restClient)
+	})
+
+	AfterEach(func() {
+
+	})
+
+	Describe("Volume Create", func() {
```
lpabon: Nice

pkg/sanity/volume.go
```diff
@@ -0,0 +1,341 @@
...
+				Expect(err).ToNot(HaveOccurred())
+			}
+
+		})
+
+		It("Should Enumerate all volumes ", func() {
+
+			By("Creating a list of volumes")
+
+			for i := 0; i < 3; i++ {
```
lpabon: Make `3` a variable

pkg/sanity/volume.go
```diff
@@ -329,43 +510,104 @@ var _ = Describe("Volume [Volume Tests]", func() {
...
+			testIfVolumeCreatedSuccessfully(volumedriver, volumeID, numVolumesBefore, vr)
 
-			volumes, err = volumedriver.Inspect([]string{volumeID})
+			By("Updating the volume spec. ")
+			newSize := 5
 
+			set := &api.VolumeSetRequest{
+				Locator: vr.GetLocator(),
+				Spec: &api.VolumeSpec{
+					Size:   uint64(newSize),
```
adityadani: not all drivers are going to support setting the size field. We might need to check for capabilities first and then do it. May be start with just setting the labels. We can have a separate test that does a resize.

PR #302, opened 2018-02-05
osdconfig library

sdeoras:
This PR defines an interface for osdconfig management and implements it

* interface.go defines the interface
* new.go creates a new instance
    * It derives local contexts to work with (these contexts are passed throughout the code)
    * It spawns a watcher that watches on kvdb changes
* api.go implements the interface

Pl. see tests on usage patterns of this library

**Util Package**
A util function has been added in the osdconfig library to help build a func literal that can be registered as a callback
Furthermore, `configutil`, a util package, is provided as a higher level abstraction for easier use

**Pending Tasks**
struct.go may have fields that need to be removed or rearranged.


osdconfig/constants.go
```diff
@@ -0,0 +1,15 @@
+package osdconfig
+
+// kvdb keys
+const (
+	base_key    = "osdconfig"   // root of the tree in kvdb
+	cluster_key = "clusterConf" // cluster data is managed behind this key
+	nodes_key   = "nodesConf"   // all nodeID's exist behind this key and all node data behind respective node ID
+)
+
+// log levels
```
lpabon: Not needed

osdconfig/struct.go
```diff
@@ -0,0 +1,98 @@
+package osdconfig
```
lpabon: Nice

osdconfig/test_execCB_abort_test.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+Loop1:
+	for i := 0; i < 10; i++ {
+		t := time.Now()
+		select {
+		case <-time.After(time.Millisecond * 100):
+			manager.Wait()
+			if time.Since(t) > time.Second { // done waiting for callback execution
+				break Loop1
```
lpabon: Maybe use `continue` here instead of loop?

sdeoras:
Thanks Luis for so quickly reviewing it... will look into changes you recommended.

osdconfig/util.go
```diff
@@ -0,0 +1,85 @@
+package osdconfig
```
harsh-px: Let's not create a util.go file. It's too generic and eventually becomes a dump ground for all miscellaneous routines and the kitchen sink.

osdconfig/interface.go
```diff
@@ -0,0 +1,39 @@
+package osdconfig
+
+import "context"
+
+// A config manager interface allows management of osdconfig parameters
+// It defines setters, getters and callback management functions
+type ConfigManager interface {
+	// Fetch cluster configuration data
+	GetClusterConf() (*ClusterConfig, error)
```
harsh-px: All comments for public types and functions need to be start with function/type name, else golint will fail. For e.g

```
// GetClusterConf fetches cluster configuration data...
GetClusterConf() (*ClusterConfig, error)
```

sdeoras: sounds good... will do.

osdconfig/util.go
sdeoras: ok, will move those util functions around and delete util.go

osdconfig/api.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// get json from kvdb and unmarshal into config
+	kvPair, err := manager.cc.Get(getNodeKeyFromNodeID(nodeID))
+	if err != nil {
+		return nil, err
+	}
+
+	if kvPair == nil {
+		return nil, errors.New("no data fetched from kvdb store")
```
harsh-px: create error constants instead of creating a new error instance each time. Having known errors will also allow users to handle errors if needed.

osdconfig/api.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	config := new(NodeConfig)
+	if err := json.Unmarshal(kvPair.Value, config); err != nil {
```
harsh-px: don't need `err :=` since err is already initialized above

osdconfig/api.go
```diff
@@ -0,0 +1,275 @@
...
+func (manager *configManager) Run(wd *DataToCallback) <-chan error {
+	manager.Lock()
+	// re-establish the context
+	// derive local contexts from parent context
+	manager.localCtx, manager.localCancel = context.WithCancel(manager.ctx)
+	manager.Unlock()
+
+	// declare a channel on which callers of Run can wait on
+	wait := make(chan error)
+	go func(waitChan chan error, dataToCallback *DataToCallback) {
```
sangleganesh: - i think the channel can go 
- prefer running callbacks in order, since there are no guarantees on running them via go routines - so no advantage really

osdconfig/api.go
```diff
@@ -0,0 +1,275 @@
...
+		return errors.New("callback already registered with this name: " + name)
+	}
+
+	return nil
+}
+
+// Run loops over all registered callbacks and executes them.
+// It returns a channel on which you only read and it contains the errors.
+func (manager *configManager) Run(wd *DataToCallback) <-chan error {
+	manager.Lock()
```
sangleganesh: its probably better to have lock for members instead of the entire struct

osdconfig/api.go
```diff
@@ -0,0 +1,275 @@
...
+	// derive local contexts from parent context
+	manager.localCtx, manager.localCancel = context.WithCancel(manager.ctx)
+	manager.Unlock()
+
+	// declare a channel on which callers of Run can wait on
+	wait := make(chan error)
+	go func(waitChan chan error, dataToCallback *DataToCallback) {
+		manager.Lock()
+		defer manager.Unlock()
+
```
sangleganesh: imo not a good idea to lock during callback. you can just copy out the slice and iterate on it.

sdeoras: lock is there simply to prevent people from registering new callbacks during callback execution. we are also storing the state data for callbacks along with their func literals. also this prevents more than one callback executions in situations where someone made multiple kvdb updates.

sangleganesh: in which case it will deadlock .. isn't it ?

osdconfig/api.go
sdeoras: nit? :) yes, we can use previously declared err... i was just in a habit to write code this way since it keeps scope local to that `if` statement and is more idiomatic.

osdconfig/api.go
sdeoras: oh yes, this is something i should have done... recall you mentioning that earlier as well... will do.

osdconfig/api.go
sdeoras: i thought deadlock was where no goroutine was able to proceed, that is not the case here. i want just one to proceed, which sync.Mutex allows me to do. Let's discuss this tomorrow... I ran several tests, it did not deadlock, but then I might not have created the conditions for it to do so... would like to understand better.

osdconfig/api.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// wait for manager to release lock
+func (manager *configManager) Wait() {
+	manager.Lock()
+	manager.Unlock()
+}
+
+// wrap logrus error
+func (manager *configManager) error(args ...interface{}) {
```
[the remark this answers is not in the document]
sdeoras: done. osdconfig library will use logrus for logging and assume logrus configuration to be done outside.

osdconfig/constants.go
sdeoras: done

osdconfig/interface.go
sdeoras: done

osdconfig/util.go
sdeoras: done

osdconfig/api.go
sdeoras: done

osdconfig/api.go
sdeoras: wondering if this needs to be done for every err message. Most are meant to be observed in logs and not managed programmatically, but some should definitely have error constants.

sdeoras: done, several of them, where an error category is evident.

osdconfig/api.go
sdeoras: wait channel is gone now. It was possible to work directly with the contexts.

osdconfig/api.go
sdeoras: i would like to protect both the data state and execution state associated with an instance. A single lock is easier to manage.

osdconfig/constants.go
```diff
@@ -0,0 +1,23 @@
...
+	INPUT_ERR    osdconfigError = base_err + "input is nil"
+	REGISTER_ERR osdconfigError = base_err + "callback name already registered"
+	DATA_ERR     osdconfigError = base_err + "no data fetched from kvdb"
+	EXEC_ERR     osdconfigError = base_err + "callback exec error"
+)
+
+// reg types
+const (
+	CLUSTER_WATCHER Watcher = "osdconfig/clusterConf"
+	NODES_WATHCER   Watcher = "osdconfig/nodesConf"
```
disrani-px: Should use the consts declared above (base_key, etc) to create these

osdconfig/constants.go
```diff
@@ -0,0 +1,23 @@
+package osdconfig
+
+// kvdb keys
+const (
+	base_key    = "osdconfig"   // root of the tree in kvdb
```
disrani-px: According to golang naming standards, variable names should be in camel case. No underscores or all caps.

osdconfig/interface.go
```diff
@@ -0,0 +1,63 @@
...
+	SetNodeConf(config *NodeConfig) error
+
+	// Register callback functions that will get triggered on a change to either
+	// cluster configuration or node configuration data in the backend.
+	// The callback function returns two channels: A channel to write to and a
+	// channel to read on. Context is provided to the callback function in order
+	// to manage cleanup during context cancellations
+	Register(name string, watcherType Watcher,
+		opt interface{},
+		cb func(ctx context.Context, opt interface{}) (chan<- *DataToCallback,
```
disrani-px: Maybe clearer to name them readChannel and writeChannel?

osdconfig/new.go
```diff
@@ -0,0 +1,119 @@
...
+	// create in memory kvdb
+	if kv, err := kvdb.New(mem.Name, "", []string{}, nil, nil); err != nil {
+		return nil, err
+	} else {
+		return kv, nil
+	}
+}
+
+// helper function to obtain kvdb key for node based on nodeID
+// the check for empty nodeID needs to be done elsewhere
```
disrani-px: Maybe assert here if nodeID is never expected to be empty to catch bugs from callers of this function.

osdconfig/osdutil/io.go
```diff
@@ -0,0 +1,168 @@
...
+package osdutil
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/pkg/errors"
+	"github.com/portworx/kvdb"
+	"github.com/sdeoras/openstorage/osdconfig"
```
disrani-px: import should be from libopenstorage. Same in other places in the change.

osdconfig/constants.go
sdeoras: @disrani-px : glad you had a chance to review. will fix the items you noted shortly.

osdconfig/interface.go
sdeoras: i like that. I used send/recv elsewhere in this code but read/write is much nicer.

osdconfig/new.go
sdeoras: ok, let me find out the package that provides this util. @harsh-px had also requested this change.

osdconfig/osdutil/io.go
sdeoras: yes, I realized it bit too late yesterday when CI was failing. Since `osdconfig` is not yet pushed to master, I only had it in the forked repo and goimports picked it up.

osdconfig/constants.go
sdeoras: done

osdconfig/constants.go
sdeoras: done

osdconfig/interface.go
sdeoras: done

osdconfig/new.go
sdeoras: which package do you normally use for this? this one I got from Harsh is for testing "github.com/stretchr/testify/require".

osdconfig/interface.go
```diff
@@ -0,0 +1,63 @@
...
+	// channel to read on. Context is provided to the callback function in order
+	// to manage cleanup during context cancellations
+	Register(name string, watcherType Watcher,
+		opt interface{},
+		cb func(ctx context.Context, opt interface{}) (chan<- *DataWrite,
+			<-chan *DataRead)) error
+
+	// Run executes callback functions, however, it is expected that users never
+	// call this function directly. Run is called behind the scenes using scheduler
+	Run(wd *DataWrite)
```
sangleganesh: this should not be exposed to clients, should be removed from the interface

osdconfig/interface.go
```diff
@@ -0,0 +1,63 @@
...
+	// This function should return an error during occurred previous callback execution cycle
+	Error() error
+
+	// Abort sends context cancellation to execution of callback functions
+	// This function should return immediately and not block
+	Abort()
+
+	// Wait waits while scheduler is busy processing callbacks
+	// This function should block and only return when (say) internal locks are released
+	Wait()
```
sangleganesh: not sure why clients need to know about this.

osdconfig/new.go
sangleganesh: dbg.Assert

osdconfig/api.go
```diff
@@ -0,0 +1,337 @@
...
+	manager.cc = nil
+	manager.cb = nil
+	manager.opt = nil
+}
+
+// Abort sends context cancellation to running callbacks if any
+func (manager *configManager) Abort() {
+	logrus.Info("aborting via context cancellation")
+	manager.runCancel()
+	manager.Lock()
```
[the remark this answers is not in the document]
sdeoras: lock was there to wait, but I am also thinking of removing last three lines in that func, so abort only sends cancellation signal and does not wait/block.

osdconfig/api.go
```diff
@@ -0,0 +1,337 @@
...
+	logrus.Info("abort successful")
+	manager.Unlock()
+}
+
+// Wait waits for running callbacks to finish
+func (manager *configManager) Wait() {
+	manager.Lock()
```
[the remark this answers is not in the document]
sdeoras: main intent was to provide a blocking operation... yes, not a convenient way to infer status. If inferring status is needed, i would propose we create a lookup bool.

osdconfig/api.go
```diff
@@ -0,0 +1,337 @@
...
+	defer manager.Unlock()
+
+	if config == nil {
+		return InputErr
+	}
+
+	// push into kvdb
+	_, err := manager.cc.Put(filepath.Join(baseKey, clusterKey), config, 0)
+	if err != nil {
```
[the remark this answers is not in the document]
sdeoras: done :)

osdconfig/api.go
```diff
@@ -0,0 +1,337 @@
...
+		return nil, InputErr
+	}
+
+	// get json from kvdb and unmarshal into config
+	kvPair, err := manager.cc.Get(getNodeKeyFromNodeID(nodeID))
+	if err != nil {
+		return nil, err
+	}
+
+	if kvPair == nil {
```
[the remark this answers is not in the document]
sdeoras: ok, i would like to learn more about kvdb returns. as a safe bet i put in that pointer check.

osdconfig/api.go
```diff
@@ -0,0 +1,337 @@
...
+	if err = json.Unmarshal(kvPair.Value, config); err != nil {
+		return nil, err
+	}
+
+	return config, nil
+}
+
+// SetNodeConf sets node config data in kvdb
+func (manager *configManager) SetNodeConf(config *NodeConfig) error {
+	manager.Lock()
```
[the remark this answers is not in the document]
sdeoras: yes these can potentially go, but need more experiments. i actually put those in yesterday because i did not want people to fire multiple kvdb puts without regard for completion of callbacks. lock here only guarantees that either kvdb is being set or callbacks are running, not both at the same time.

osdconfig/interface.go
sdeoras: I agree as you can tell with the comments I wrote. It is, however, very useful for tests and several of my tests currently depend on it.

osdconfig/interface.go
sdeoras: Anyone wanting to set/get values in kvdb through osdconfig are recommended to use it. Also it is needed if you use Run function. I would keep it for now and remove later once use cases and stability is better understood.

sangleganesh: to be frank this is way complicated (because it exposes internal mechanisms), unnecessary (clients have nothing to do with this), and creates interfaces which are not required.

osdconfig/interface.go
sangleganesh: imho test should not expose implementation details in interface.
if you want to test implementation, use casts.

osdconfig/interface.go
sdeoras: per discussions yesterday, I provided a higher level util package, `configutil`. It has just 4 functions, two for creating config managers and two for setting up a watch, that's it... no context, no channels. I agree the osdconfig package is bit raw, but complexity should no longer be an issue.

osdconfig/interface.go
sdeoras: ok, let me take a look. btw, this will also get exposed through REST, where I don't plan to expose most of these functions. For devs, including myself, I anticipate needing these down the lane though.

PR #306, opened 2018-02-07
aws: Fix attached path

### LINKED ISSUE #305: Incorrect path saved onto the aws metadata.
Bug

**What happened**:
When using `storageops/aws`, the path on the system and the path on the instance AWS metadata does not match. This is due to the library returning always `/dev/sd*` to attach to even when the system does not use `/dev/sd*` instead it uses `/dev/xvd*`.

The path where the volume was attached on the system should match the path of on the metadata of the instance.

lpabon:
Uses the root drive to determine if it should use `/dev/sd` or `/dev/xvd` as the prefix.

Closes #305

### DIFF
--- a/pkg/storageops/aws/aws_test.go
+++ b/pkg/storageops/aws/aws_test.go
@@ -1,4 +1,4 @@
-package aws_test
+package aws
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


pkg/storageops/aws/aws_test.go
```diff
@@ -1,4 +1,4 @@
-package aws_test
+package aws
```
harsh-px: having a separate package name `aws_test` allows to test the package as if the tester was an external package. If the test package name is the same as the package being tested, the test package can also use methods and variables not exposed to the eventual user.

lpabon: That is a good thing. Unit tests should access internal structures and functions to be tested. You are describing functional tests.

harsh-px: Discussed offline. The goal is to also test internal functions and hence using the same package name.

PR #308, opened 2018-02-09
aws: Remove debugging code

lpabon:
Debugging code was being printed to stdout and it is not necessary from a library

### DIFF
--- a/pkg/storageops/aws/aws.go
+++ b/pkg/storageops/aws/aws.go
@@ -146,10 +146,10 @@ func (s *ec2Ops) waitAttachmentStatus(
 	request := &ec2.DescribeVolumesInput{VolumeIds: []*string{&id}}
 	actual := ""
 	interval := 2 * time.Second
-	fmt.Printf("Waiting for state transition to %q", desired)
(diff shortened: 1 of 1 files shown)


pkg/storageops/aws/aws.go
```diff
@@ -146,10 +146,9 @@ func (s *ec2Ops) waitAttachmentStatus(
 	request := &ec2.DescribeVolumesInput{VolumeIds: []*string{&id}}
 	actual := ""
 	interval := 2 * time.Second
-	fmt.Printf("Waiting for state transition to %q", desired)
```
adityadani: Any reason why we want to remove this code ? It helps debugging AWS ebs attach issues.

lpabon: If we need it, then let's add it to the log, not fmt.Printf(). Is that ok?

adityadani: +1 for logrus instead of fmt.

lpabon:
Noticed that the `runs` variable was not being updated, so I fixed that.

Issue #310, opened 2018-02-10
Migrate dlog 


jvinod:
dlog is no longer maintained, author recommends to migrate to https://github.com/uber-go/zap:

lpabon:
I prefer not moving to zap since there is no need for speed in openstorage. My suggestion is to use the standard Kubernetes simple logger https://github.com/golang/glog .

We can also discuss changing dlog to logrus, but my vote is for glog.

lpabon:
After some discussion (offline) the idea is to use logrus.

peter-edge:
Glog was never meant to be a standard logger for Golang, and is not recommended, even by Google anymore. Logrus is very slow, and a lot of people are moving away from it. The author also does things like change the import path. Zap has kind of become a new standard.

lpabon:
Cool, then we will change to it when it comes part of golang.

lpabon:
Closed by #363

PR #311, opened 2018-02-10
WIP: Move to zap logging

peter-edge:
This starts the work to move to github.com/uber-go/zap. **Vendor still needs to be done, as well as verification that everything works as expected.**

Zap exposes a `*Logger` object, which works with the primary Zap API, and a `*SugaredLogger` object, which exposes a logger more similar to other Golang loggers. By default, a no-op logger is the global logger, so you have to register a logger at startup.

There's a lot of configuration options, but two loggers are easily created with https://godoc.org/go.uber.org/zap#NewDevelopment and https://godoc.org/go.uber.org/zap#NewProduction. I've set this up to use `NewProduction`, but `NewDevelopment` might be closer to what you would expect.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -20,7 +20,7 @@ IF YOU CODE ON OPENSTORAGE, YOU ARE EXPECTED TO KNOW THIS. Just take the 20 minu
 
 ### Items
 
-* Use [dlog](https://go.pedge.io/dlog) for logging.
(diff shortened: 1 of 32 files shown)


lpabon:
Let's discuss this on #310

lpabon:
Thank you for the change, but we have agreed to continue using logrus. 

Closed by #363

PR #320, opened 2018-02-16
Update cloudbackup interface to make it consistent

disrani-px:


api/api.go
```diff
@@ -216,8 +216,6 @@ type CredCreateRequest struct {
 type CredCreateResponse struct {
 	// UUID of the credential that was just created
 	UUID string
-	// CredErr indicates reasonfor failed CredCreate
-	CredErr string
```
lpabon: Removing this from the API? Is this no longer needed? Please get approval from the creator of CredCreateResponse (unless you are it :) ).

disrani-px:
The error string will be in the err returned from each API.

api/api.go
disrani-px: Yes, not longer needed

PR #324, opened 2018-02-20
add check against empty volume names. change volumeID to use uuidgen

### LINKED ISSUE #321: NFS driver accepts blank volume names when creating volume using REST endpoint
BUG REPORT
**What happened**:
NFS driver allows to create with blank volume name or double-spaces or n-spaces volume names when using the REST endpoint.
ex : "" , " " , "   " , "    " -> all of the mentioned are created when created using REST endpoint
Expect an error to the user that blank names are not valid volume names.
Make a rest call to create endpoint with the volume name set as blank

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

timbouvier:
nfs volume driver doesn't check for empty volume names or spaces in volume name during create. nfs volume driver doesn't use uuid for volume id. This PR fixes those things
https://github.com/libopenstorage/openstorage/issues/321
Closes #321 

### DIFF
--- a/volume/drivers/nfs/nfs.go
+++ b/volume/drivers/nfs/nfs.go
@@ -18,6 +18,7 @@ import (
 	"github.com/libopenstorage/openstorage/pkg/seed"
 	"github.com/libopenstorage/openstorage/volume"
 	"github.com/libopenstorage/openstorage/volume/drivers/common"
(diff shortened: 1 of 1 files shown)


timbouvier:
whoops! wrong branch

PR #326, opened 2018-02-21
nfs - validate volume names

### LINKED ISSUE #321: NFS driver accepts blank volume names when creating volume using REST endpoint

timbouvier:
Signed-off-by: bouviti <timothy.bouvier@verizon.com>

nfs volume driver doesn't check for empty volume names or spaces in volume name during create. nfs volume driver doesn't use uuid for volume id. This PR fixes those things

Closes #321 

### DIFF
--- a/volume/drivers/nfs/nfs.go
+++ b/volume/drivers/nfs/nfs.go
@@ -18,6 +18,7 @@ import (
 	"github.com/libopenstorage/openstorage/pkg/seed"
 	"github.com/libopenstorage/openstorage/volume"
 	"github.com/libopenstorage/openstorage/volume/drivers/common"
+	"github.com/pborman/uuid"
 	"github.com/portworx/kvdb"
 	"math/rand"
 	"strings"
@@ -198,7 +199,7 @@ func (d *driver) getNFSVolumePathById(volumeID string) (string, error) {
 }
 
 //append unix time to volumeID
-func (d *driver) getNewSnapVolID(volumeID string) string {
+func (d *driver) getNewSnapVolName(volumeID string) string {
 	return volumeID + "-" + strconv.FormatUint(uint64(time.Now().Unix()), 10)
 }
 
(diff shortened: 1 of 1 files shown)


volume/drivers/nfs/nfs.go
```diff
@@ -211,14 +212,18 @@ func (d *driver) Create(
...
-		dlog.Infof("Creating snap vol id: %s", volumeID)
+	if locator.Name == "" {
+		return "", errors.New("volume name cannot be empty")
 	}
 
+	if hasSpaces := strings.Contains(locator.Name, " "); hasSpaces {
+		return "", errors.New("volume name cannot contain space characters")
+	}
+
+	volumeID := strings.TrimSuffix(uuid.New(), "\n")
```
lpabon: Why `\n` from the end? I do not think it has one.

volume/drivers/nfs/nfs.go
```diff
@@ -211,14 +212,18 @@ func (d *driver) Create(
 	source *api.Source,
 	spec *api.VolumeSpec) (string, error) {
 
-	volumeID := locator.Name
-	if volumeID == "" && source.Parent != "" {
-		volumeID = d.getNewSnapVolID(source.Parent)
-		dlog.Infof("Creating snap vol id: %s", volumeID)
+	if locator.Name == "" {
```
lpabon: This works, but instead use the style `if len(location.Name) == 0` to check for empty strings.

volume/drivers/nfs/nfs.go
```diff
@@ -211,14 +212,18 @@ func (d *driver) Create(
...
+		return "", errors.New("volume name cannot be empty")
```
lpabon: Use `errors.New` only when creating a contract `ErrXXX` API value. Since these are not saved as contract API variables, just return using `fmt.Errorf(...)`

volume/drivers/nfs/nfs.go
timbouvier: @lpabon I stole this line from buse driver. Maybe there was some reason the buse driver trimmed the newline?

volume/drivers/nfs/nfs.go
timbouvier: @lpabon okay

volume/drivers/nfs/nfs.go
timbouvier: @lpabon okay

volume/drivers/nfs/nfs.go
lpabon: weird, yeah, maybe.

PR #329, opened 2018-02-22
add make docs to travis build

pault84:

Automate doc generation  (if swagger spec is provided)
Validate generated swagger json.

This allows us to not have to worry about doc generation and will error if something was forgotten in the spec.


pault84:
This is ready but the docs won't get pushed anywhere.
We can add a git add/commit/push into master for this. but i'm not sure that's the right way to go.

lpabon:
Let's first figure out how can we have API documentation for different versions. Once we have that defined, then we can investigate how to update them on automatically.

lpabon:
To this add the following to `cmd/osd/main.go`:

1. A description that describes this as the REST api, and for a golang client go to openstorage.
2. Update version to v2.0.0

In api.openstorage.org:

1. index.html to have links to versioned APIs. (simple links).
2. Have a path of /<version>/<api>

We should update these at every release of the OSD API.

PR #330, opened 2018-02-22
Make gossip port configurable

pault84:

Right now gossip is pinned on 9002. This causes issues for some people if that port is already taken.

If using the library it's now configurable. if using straight OSD we use 9002 still.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -260,7 +260,7 @@ type Cluster interface {
 	// It also causes this node to join the cluster.
 	// nodeInitialized indicates if the caller of this method expects the node
 	// to have been in an already-initialized state.
-	Start(clusterSize int, nodeInitialized bool) error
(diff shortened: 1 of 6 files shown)


cluster/cluster.go
```diff
@@ -260,7 +260,7 @@ type Cluster interface {
 	// It also causes this node to join the cluster.
 	// nodeInitialized indicates if the caller of this method expects the node
 	// to have been in an already-initialized state.
-	Start(clusterSize int, nodeInitialized bool) error
+	Start(clusterSize int, nodeInitialized bool, gossipPort string) error
```
sangleganesh: not sure why is port being passed as a string ?

pault84: it's used as a string. why convert from int to string if we don't have to?

sangleganesh: it needs to be validated if its a string.

PR #331, opened 2018-02-23
osd sanity haupdate,snapshot,resize fix.

ipochi:

This PR adds snapshot tests, volume ha-update and resize tests to osd-sanity and fixes the size issue in previous tests.
Closes #

### DIFF
--- a/pkg/sanity/osd_test_util.go
+++ b/pkg/sanity/osd_test_util.go
@@ -93,12 +93,3 @@ func random(min, max int) int {
 	rand.Seed(time.Now().Unix())
 	return rand.Intn(max-min) + min
 }
-
-func getNewHALevel(currentHA int64) (newHALevel int) {
-	if currentHA == 1 || currentHA == 3 {
-		newHALevel = 2
-	} else {
-		newHALevel = []int{1, 3}[random(0, 2)]
-	}
-	return
-}
--- a/pkg/sanity/osd_test_util.go
+++ b/pkg/sanity/osd_test_util.go
@@ -17,11 +17,22 @@ limitations under the License.
 package sanity
 
 import (
+	"math/rand"
+	"time"
+
 	"github.com/libopenstorage/openstorage/api"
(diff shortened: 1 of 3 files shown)


pkg/sanity/osd_test_util.go
```diff
@@ -65,11 +76,29 @@ func testIfVolumeCreatedSuccessfully(
...
+	}
+	rand.Seed(time.Now().Unix())
+	return rand.Intn(max-min) + min
+}
+
+func getNewHALevel(currentHA int64) (newHALevel int) {
+	if currentHA == 1 || currentHA == 3 {
+		newHALevel = 2
+	} else {
+		newHALevel = []int{1, 3}[random(0, 2)]
```
adityadani: I don't understand this. Why only randomize when HA is 2?
You could always randomize within a set of numbers which does not contain the current HA level.

pkg/sanity/snapshot.go
```diff
@@ -0,0 +1,317 @@
...
+			volumes, err := volumedriver.Enumerate(&api.VolumeLocator{}, nil)
+			Expect(err).ToNot(HaveOccurred())
+			numVolumesAfter = len(volumes)
+		})
+
+		It("Should enumerate Volume snapshots", func() {
+
+			By("Creating the volume")
+
+			var err error
```
adityadani: nit: create the volume once in BeforeEach of the snapshot test suite. so that you don't have to create it in every "It"

ipochi: Every Describe block is a new test , so sometimes it could be that there is a different configuration or different number of volumes are to be created

pkg/sanity/osd_test_util.go
ipochi: As discussed ,  I will remove the random HA- update and make it simple increment of 1 for this PR.
Next PR will address this issue and further tests of having more combinations of creating with different replication factor combinations and replification factor updates.

PR #332, opened 2018-02-24
Osd sanity bkp restore

ipochi:
Addtion of Backup and Restore tests to osd-sanity
Closes #

### DIFF
--- a/cmd/osd-sanity/sanity_test.go
+++ b/cmd/osd-sanity/sanity_test.go
@@ -28,15 +28,17 @@ const (
 )
 
 var (
-	VERSION  = "(dev)"
-	endpoint string
-	driver   string
-	version  bool
+	VERSION               = "(dev)"
+	endpoint              string
+	driver                string
+	cloudBackupConfigPath string
+	version               bool
 )
 
 func init() {
 	flag.StringVar(&endpoint, prefix+"endpoint", "", "OSD endpoint")
 	flag.StringVar(&driver, prefix+"driver", "", "Volume Driver")
+	flag.StringVar(&cloudBackupConfigPath, prefix+"cloudbackupconfig", "", "Test Cloud Backup/Restore config file , optional")
 	flag.BoolVar(&version, prefix+"version", false, "Version of this program")
 	flag.Parse()
 }
@@ -52,5 +54,9 @@ func TestSanity(t *testing.T) {
 	if len(driver) == 0 {
 		t.Fatalf("--%s.driver must be provided with a Volume driver", prefix)
 	}
-	sanity.Test(t, endpoint, driver)
+	if len(cloudBackupConfigPath) == 0 {
+		t.Logf("No Cloud backup config file provided , Cloud Backup & Restore Tests will be skipped")
+	}
+
+	sanity.Test(t, endpoint, driver, cloudBackupConfigPath)
 }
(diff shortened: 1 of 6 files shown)


cmd/osd-sanity/sanity_test.go
```diff
@@ -52,5 +54,9 @@ func TestSanity(t *testing.T) {
 	if len(driver) == 0 {
 		t.Fatalf("--%s.driver must be provided with a Volume driver", prefix)
 	}
-	sanity.Test(t, endpoint, driver)
+	if len(cloudBackupConfigPath) == 0 {
+		t.Logf("No Cloud backup config file provided , Cloud Backup & Restore Tests will be skipped")
+	}
+
+	sanity.Test(t, endpoint, driver, cloudBackupConfigPath)
```
lpabon: Not now, but next time, create a new `OsdSanityConfig{}` struct and pass that to `Tests()`. That way we can add any other configuration change without changing the number of parameters.

pkg/sanity/sanity.go
```diff
@@ -17,26 +17,58 @@ limitations under the License.
...
+	CloudProviders map[string]map[string]string
+}
+
 // Test will test the CSI driver at the specified address
-func Test(t *testing.T, address, driver string) {
+func Test(t *testing.T, address, driver, cloudBackupConfigPath string) {
 	lock.Lock()
 	defer lock.Unlock()
 
+	cfg, err := CloudProviderConfigParse(cloudBackupConfigPath)
```
lpabon: You probably can check if `len(cloudBackupConfigPath) == 0` and not make this call at all, right?

pkg/sanity/backup_restore.go
```diff
@@ -0,0 +1,423 @@
...
+		var err error
+		restClient, err = volumeclient.NewDriverClient(osdAddress, volumeDriver, volume.APIVersion, "")
+
+		Expect(err).ToNot(HaveOccurred())
+		volumedriver = volumeclient.VolumeDriver(restClient)
+
+		By("Creating Credentials first")
+
+		if cloudBackupConfig != nil {
+			cloudParamsMap := cloudBackupConfig.CloudProviders["azure"]
```
lpabon: This is an Azure test, but line 33 does not specify that it is an azure test. Either add `Azure` to the description in line 33 and copy it for each aws, and gce, or make the key in 55 configurable.

pkg/sanity/backup_restore.go
```diff
@@ -0,0 +1,423 @@
...
+
+			By("Doing Backup")
+
+			bkpReq := &api.BackupRequest{
+				CredentialUUID: credUUID,
+				Full:           false,
+				VolumeID:       volumeID,
+			}
+
+			time.Sleep(time.Second * 10)
```
lpabon: Why sleep? Open ended feedback is erroneous. Try to find another model for detecting a status or event change.

ipochi: done

pkg/sanity/backup_restore.go
ipochi: Made it such a way that , it will do cloud backup restore tests for all credentials in the config file , provided the credentials get validated

pkg/sanity/sanity.go
ipochi: makes sense , parse the file only if its provided.

cmd/osd-sanity/sanity_test.go
ipochi: i'll keep in mind.

PR #339, opened 2018-03-01
Add metadata field for backup info

disrani-px:


api/api.go
```diff
@@ -291,6 +291,8 @@ type BackupInfo struct {
 	// Timestamp is the timestamp at which the source volume
 	// was backed up to cloud
 	Timestamp time.Time
+	// Metadata associated with the backup
```
lpabon: Why is this necessary? Could you provide some context?

disrani-px: Backups might have extra metadata associated with it. This is used to return that metadata with other backup information.

PR #343, opened 2018-03-05
Node delete added in osdconfig

sdeoras:
Added node delete method in osdconfig interface. It is required for cleanup of data that in etcd

### DIFF
--- a/api/client/cluster/client.go
+++ b/api/client/cluster/client.go
@@ -324,3 +324,11 @@ func (c *clusterClient) SetNodeConf(config *osdconfig.NodeConfig) error {
 	}
 	return nil
 }
+
+func (c *clusterClient) UnsetNodeConf(nodeID string) error {
+	request := c.c.Delete().Resource(clusterPath + UriNode + "/" + nodeID)
+	if err := request.Do().Error(); err != nil {
+		return err
+	}
+	return nil
+}
--- a/osdconfig/api.go
+++ b/osdconfig/api.go
@@ -74,6 +74,20 @@ func (manager *configManager) SetNodeConf(config *NodeConfig) error {
 	return err
 }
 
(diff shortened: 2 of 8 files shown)


api/client/cluster/client.go
```diff
@@ -324,3 +324,11 @@ func (c *clusterClient) SetNodeConf(config *osdconfig.NodeConfig) error {
 	}
 	return nil
 }
+
+func (c *clusterClient) DelNodeConf(nodeID string) error {
```
lpabon: If delete is the opposite of `SetNode`, then it should be `Unset`. Or change `SetNodeConf` to `CreateNodeConf` and then the opposite would be `DeleteNodeConf`

osdconfig/api_test.go
```diff
@@ -75,6 +75,17 @@ func TestSetGetNode(t *testing.T) {
 	if !reflect.DeepEqual(expectedConf, receivedConf) {
 		t.Fatal("expected and received values are not deep equal")
 	}
+
+	// now delete the node
+	if err := manager.DelNodeConf(expectedConf.NodeId); err != nil {
+		t.Fatal("error in deleting node config")
```
lpabon: I see this code uses `t.Fatal()` a lot, but instead start using `testify.assert(...)`

sdeoras:
@lpabon i am merging this PR since need it to close another one elsewhere. I already made the changes you requested.

osdconfig/api.go
```diff
@@ -74,6 +74,20 @@ func (manager *configManager) SetNodeConf(config *NodeConfig) error {
 	return err
 }
 
+// UnsetNodeConf deletes node config data in kvdb
+func (manager *configManager) UnsetNodeConf(nodeID string) error {
+	manager.Lock()
```
sangleganesh: do we need the lock here ?

PR #344, opened 2018-03-05
Fix leaky http client when using HTTPS

pault84:

When doing alot of concurrent Attaches/Detaches i saw a ridiculous increase in go-routines.
After doing some research it seems net/http standard client is pretty leaky in this regard because it is trying to re-use connections and keeps them open.


pault84:
This is not the actual fix. issue is server side.

Issue #342, opened 2018-03-03
Cleanup APIs related to logging url, management url, fluentd support


adityadani:
OSD does not support any of the following
- logging url
- management url
- fluentd url
- tunner endpoint

Cleanup the APIs for getting and setting them.

ipochi:
closing as the branch is merged.

PR #347, opened 2018-03-06
Deprecate unused cluster APIs

adityadani:

- SetLogginURL -> Not used by anyone
- SetManagementURL -> Never saw the light of the day towards completion
- SetTunnelConfig -> Never saw the light of the day towards completion
- SetFluentd -> Not used by anyone.

Closes #
#342


cluster/cluster.go
```diff
@@ -50,14 +50,16 @@ type NodeEntry struct {
...
-	Status        api.Status
-	Id            string
-	NodeEntries   map[string]NodeEntry
-	LoggingURL    string
+	Size        int
+	Status      api.Status
+	Id          string
+	NodeEntries map[string]NodeEntry
+	// LoggingURL - Deprecated
+	LoggingURL string
```
pault84: I think we can get rid of these too. unMarshal should skip fields that don't exist in the object i believe.
If not it's fine to leave these.

PR #349, opened 2018-03-08
Enable disk list, create and delete outside of the instance zone for GCE

harsh-px:

The create, delete and list methods were assuming that the disk was in the same zone as instance.

### DIFF
--- a/pkg/storageops/gce/gce.go
+++ b/pkg/storageops/gce/gce.go
@@ -207,7 +207,7 @@ func (s *gceOps) Delete(id string) error {
 	}
 
 	if !found {
-		return fmt.Errorf("disk: %s as it wasn't found", id)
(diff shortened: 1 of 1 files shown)


pkg/storageops/gce/gce.go
```diff
@@ -187,8 +187,30 @@ func (s *gceOps) Create(
...
+			}
+		}
+		return nil
+	}); err != nil {
+		logrus.Errorf("failed to list disks: %v", err)
+		return err
+	}
+
+	if !found {
+		return fmt.Errorf("disk: %s as it wasn't found", id)
```
adityadani: nit: Failed to delete disk: %s as it wasn't found.

harsh-px: done

PR #350, opened 2018-03-12
Fix for #281 and mock tests for the same.

### LINKED ISSUE #281: HTTP error 500 without the error message from the cluster
Cluster Client wrapper `api/client/cluster/client.go` provides HTTP 500 response from the server on failed scenarios , but the error message provided by the server is lost.

Failed scenarios of below endpoints are currently displaying the above said behaviour

deleteFluentDconfig
setFluentDConfig
setTunnelConfig
setManagementURL
setLoggingURL

ipochi:

This PR fixes the issue of some endpoints returning just 500 error masking the actual error returned by the server.
Closes #281 


api/client/request.go
```diff
@@ -210,31 +210,14 @@ func headerVal(key string, resp *http.Response) (int, bool) {
...
 
-	var (
-		status *Status
-		err    error
-	)
-
 	httpOK := resp.StatusCode >= http.StatusOK && resp.StatusCode <= http.StatusPartialContent
-	hasStatus := false
-	if body != nil {
-		err = json.Unmarshal(body, status)
```
ipochi: This line first of all was wrong.
It should have been `json.Unmarshal(body, &status)`

Secondly regardless of the response from the server , the status instance was always nil [before above correction] or empty instance[ after correction]

After careful observation , it turns out its not needed at all.
Since the status code is enough to tell us if the body contains a json string for the client to unmarshal or an error string.

lpabon:
Looks good to me. Let's get a review from @jvinod who wrote that code 3 years ago.

api/client/request.go
jvinod: status is a pointer, do we need to pass &status?  You're relying on the http error to be returned on error, this may not be true. You need to verify that is before this change goes in.

lpabon:
This requires some coordination between our portworx driver, osd API, and this change. 
1. Document error reporting in the API in as an issue.
2. Make the change in openstorage.
3. Make the change in Px.

I'll create 1, and then 3.

PR #357, opened 2018-03-14
Add Cloudbackup Delete API to delete backups

disrani-px:
The previous Delete has been renamed to DeleteAll to delete all backups for a
given volume

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -296,6 +296,15 @@ type CloudBackupEnumerateResponse struct {
 }
 
 type CloudBackupDeleteRequest struct {
+	// ID is the ID of the cloud backup
+	ID string
+	// CredentialUUID is the credential for cloud to be used for the request
+	CredentialUUID string
+	// Force Delete cloudbackup even if there are dependencies
+	Force bool
+}
+
+type CloudBackupDeleteAllRequest struct {
 	CloudBackupGenericRequest
 }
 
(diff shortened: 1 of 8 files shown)


api/api.go
```diff
@@ -296,6 +296,15 @@ type CloudBackupEnumerateResponse struct {
 }
 
 type CloudBackupDeleteRequest struct {
```
pault84: Someone will need to document these API's (and i'm not suggesting myself :) )

api/server/volume.go
```diff
@@ -951,6 +951,7 @@ func (vd *volAPI) Routes() []*Route {
 		{verb: "POST", path: backupPath("/restore", volume.APIVersion), fn: vd.cloudBackupRestore},
 		{verb: "GET", path: backupPath("", volume.APIVersion), fn: vd.cloudBackupEnumerate},
 		{verb: "DELETE", path: backupPath("", volume.APIVersion), fn: vd.cloudBackupDelete},
+		{verb: "DELETE", path: backupPath("/all", volume.APIVersion), fn: vd.cloudBackupDeleteAll},
```
pault84: old api should be deprecated if we don't intend on using it anymore. Again this should happen through swagger api and then removed from next release.

We should also release not this as a breaking change. (same with all the other cloudsnap api renames.

api/api.go
disrani-px: I'll file a separate bug for that. All the cloudbackup APIs need to be documented.

api/server/volume.go
disrani-px: Old API will still be used.

Changes for undocumented APIs don't need to be release noted as it is expected that they can change. Once we documents them we'll release note breaking changes (we should not be making breaking changes for documented APIs in the first place)

api/api.go
disrani-px: Issue #362 to track that.

PR #360, opened 2018-03-14
Add an identifier to storage device to identify metadata only devices.

prabirpaul:

It's needed to propagate information about device(s) being used as a journal device.


api/api.proto
```diff
@@ -194,6 +194,8 @@ message StorageResource {
   string rotation_speed = 11;
   // Timestamp of last time this device was scanned.
   google.protobuf.Timestamp last_scan = 12;
+  // True if dedicated for metadata.
+  bool metadata = 13;
```
sangleganesh: this may not be right level to set this info, because storage resources are bundled up into higher level abstractions, which is where it resides currently.

lpabon:
Also, we have to be really careful of adding and changing protocol buffers / API from now on. We are trying to harden it so we need to make sure a change is really needed.

api/api.proto
prabirpaul: Thought about adding this information in the Node struct (as another map of Disks). That meant Node having to distinguish between data disks and metadata disks. It's ok for StorageInfo to do the same distinction but in Node struct it seemed like introduction of internal storage specific info in otherwise generic struct. That was my reasoning.
I can change if you think this info would be better placed somewhere else, lmk.

PR #361, opened 2018-03-14
Use sirupsen/logrus and switch to dep

lpabon:
Signed-off-by: Harsh Desai <harsh@portworx.com>
Signed-off-by: Luis Pabón <luis@portworx.com>

**What this PR does / why we need it**:
logrus has changed the github team repo.

**Special notes for your reviewer**:
This has a temporary dependency on kvdb@logrus-lowercase and gossip@logrus branches (see Gopkg.toml). Once those packages are updated, we need to update openstorage dependency.

PR #363, opened 2018-03-16
Update to use sirupsen/logrus

lpabon:
dlog has been removed and is no longer available so we needed to standardize on logrus.

This new model forces vendor go files to use `sirupsen/logrus` instead
of `Sirupsen/logrus` by running the new script
`hack/force-github.com-sirupsen-logrus.sh`.

This change also removes the use of dlog and updates to the latest
logrus.

I found a much simpler method than #361 . After over 16 hours of work by Harsh and I on multiple repos, we were still not done. Not only that, but one of the biggest issues was that changing so much code would have really created havoc on our testing and quality. We had to not only change our own code, but also moby, docker, etc.

Requirements are: 
1. Remove dlog since it no longer exists. 
2. Update logrus to use sirupsen from Sirupsen.

To accomplish this I have a new proposal. We change our own code to sirupsen from Sirupsen just like before, but now we update the vendor to use sirupsen from Sirupsen by running the following command:

```
$ cat hack/force-github.com-sirupsen-logrus.sh 
#!/bin/sh
find . -name '*.go' -exec sed -i "s#github.com/Sirupsen#github.com/sirupsen#g" {} \;
```

This forces all imports to be what we want. This is simple, complete, and easy to maintain.


harsh-px:
I think that's okay compromise for now considering the churn involved. I use the same hack for torpedo: https://github.com/portworx/torpedo/blob/master/Makefile#L61

At some point, when we have cycles, we need to remove the hack and update porx to be able to use the openstorage branch with smaller case sirupsen )https://github.com/libopenstorage/openstorage/tree/sirupsen-lowercase)

@jvinod

adityadani:
+1 for this change.

lpabon:
Talked to @jvinod offline -- gave the +1

PR #364, opened 2018-03-18
PWX-4411: Convert node IP to ID if necessary for API calls.

jrivera-px:
PWX-4411:  API calls which take in node id should be also be able to handle the node ip also.  Check input to see if its an ip and convert it to the corresponding node ID.    This was a customer request and allows for a more flexible API. 

Closes #

Add an IP to uuid lookup to :
restoreCloudSnap,
inspectNode,
deleteNode,
createVolume,
updateVolumeReplication


api/server/volume.go
```diff
@@ -66,6 +67,47 @@ func (vd *volAPI) parseID(r *http.Request) (string, error) {
...
+				return nodeIds, err
+			}
+			nodeIds = append(nodeIds, id)
+		}
+	}
+
+	return nodeIds, err
+}
+
+func (vd *volAPI) replicaSpecIPtoIds(rspecRef *api.ReplicaSet) error {
```
lpabon: Add a comment and change function name to updateXXX

api/server/cluster.go
```diff
@@ -171,6 +171,47 @@ func (c *clusterApi) gossipState(w http.ResponseWriter, r *http.Request) {
...
+//   in: path
+//   description: cluster node ip or id
+//   required: true
+//   type: string
+// responses:
+//   '200':
+//      description: cluster node ID
+//      schema:
+//         type: string
+func (c *clusterApi) getnodeidfromip(w http.ResponseWriter, r *http.Request) {
```
sangleganesh: nit: camel case ?

api/server/volume.go
```diff
@@ -66,6 +67,49 @@ func (vd *volAPI) parseID(r *http.Request) (string, error) {
 	return "", fmt.Errorf("could not parse snap ID")
 }
 
+func (vd *volAPI) nodeIPtoIds(nodes []string) ([]string, error) {
+	nodeIds := make([]string, 0)
```
sangleganesh: nit: nodeIds := make([]string, len(nodes))

cluster/manager.go
```diff
@@ -1398,6 +1424,13 @@ func (c *ClusterManager) Remove(nodes []api.Node, forceRemove bool) error {
 	inQuorum := !(c.selfNode.Status == api.Status_STATUS_NOT_IN_QUORUM)
 
 	for _, n := range nodes {
+
+		if id, cerr := c.GetNodeIdFromIp(n.Id); cerr == nil {
```
sangleganesh: i think "n" is a copy of api.Node, prefer to change nodes[i] if that is referred elsewhere.

api/server/cluster.go
```diff
@@ -171,6 +171,47 @@ func (c *clusterApi) gossipState(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	vars := mux.Vars(r)
+	nodeIP, ok := vars["idip"]
+
+	if !ok || nodeIP == "" {
+		c.sendError(c.name, method, w, "Missing id param", http.StatusBadRequest)
```
adityadani: nit: Missing input param. 
The input can be id as well as ip right?

cluster/manager.go
```diff
@@ -266,6 +268,30 @@ func (c *ClusterManager) GetData() (map[string]*api.Node, error) {
 	return nodes, nil
 }
 
+func (c *ClusterManager) nodeIdFromIp(idIp string) (string, error) {
+	// Caller's responsibility to lock the access to the NodeCache.
+	for _, n := range c.nodeCache {
+		if n.DataIp == idIp || n.MgmtIp == idIp {
+			logrus.Infof("Node IP: " + idIp + " maps to ID: " + n.Id)
```
adityadani: nit: may be you want to remove his log line as it will log for every http request which makes call to this API.

PR #368, opened 2018-03-20
CSI 0.2 support

### LINKED ISSUE #333: CSI Moving to 0.2
FEATURE

**What happened**:
CSI has made incompatible changes from 0.1 to 0.2.

CSI implementation needs to be updated to be 0.2 compatible.

See: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/container-storage-interface-community/3KcxXUkqee8/ejeJX_djBQAJ

lpabon:
Support for CSI 0.2 which will be part of Kubernetes 1.10

Closes #333 

Also update vendored csi-sanity which checks the APIs for 0.2


lpabon:
The second commit has this message:

```
The CSI API NodeGetId was being flagged as an error by golint
because it should be NodeGetID. The issue is we have no control
over the name.

Therefore golint, which only was set for CSI, has been set for
all of openstorage but warnings are no longer errors.

```

lpabon:
Major changes are:

* Sizes are in int64 from uint64
* No more versions
* API NodeGetId changed from GetNodeId

csi/identity.go
```diff
@@ -17,50 +17,47 @@ limitations under the License.
...
 		},
 	}, nil
 }
 
+// Probe is a CSI API
+func (s *OsdCsiServer) Probe(
+	ctx context.Context,
+	req *csi.ProbeRequest,
+) (*csi.ProbeResponse, error) {
+	return &csi.ProbeResponse{}, nil
```
adityadani: It expects an empty response?

adityadani: Oh this is just replacing NodeProbe? If so can you add the TBD comment here.

adityadani: discussed offline. Merging this PR.

PR #371, opened 2018-03-22
osdconfig during node decommission

sdeoras:

When node is decommissioned, it needs to be removed from osdconfig as well. This PR adds that logic

### DIFF
--- a/cluster/manager.go
+++ b/cluster/manager.go
@@ -1490,6 +1490,11 @@ func (c *ClusterManager) Remove(nodes []api.Node, forceRemove bool) error {
 				}
 			}
 		}
+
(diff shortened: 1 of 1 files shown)


cluster/manager.go
```diff
@@ -1516,6 +1516,11 @@ func (c *ClusterManager) NodeRemoveDone(nodeID string, result error) {
 			nodeID, err)
 		logrus.Errorf(msg)
 	}
+
```
sangleganesh: can we move the code before deleteNodeFromDB ? this way even if the node crashes after remove config we can still re-run decommission again ?

sdeoras: done

PR #373, opened 2018-03-24
Add implementation to support paths in mount table

jvinod:

Add support to enumerate deleted mounts based on source path. Offers a way to detect such paths and unmount them.


jvinod:
close this in favor of https://github.com/libopenstorage/openstorage/pull/375 that has the 'signed commit.

PR #375, opened 2018-03-24
Add implementation to support deleted paths in mount table

jvinod:

Support "(deleted)" mount entries in mount table and ability to purge them

replaces https://github.com/libopenstorage/openstorage/pull/373

### DIFF
--- a/pkg/mount/deleted_mount.go
+++ b/pkg/mount/deleted_mount.go
@@ -1,0 +1,128 @@
+// +build linux
+
+package mount
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	AllDevices = "AllDevices"
+)
+
+// deletedMounter loads mounts that show up as deleted in the mount table.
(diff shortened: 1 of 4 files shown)


pkg/mount/deleted_mount.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	AllDevices = "AllDevices"
```
adityadani: nit: comment for AllDevices

pkg/mount/deleted_mount.go
```diff
@@ -0,0 +1,128 @@
...
+
+func (m *deletedMounter) Mount(
+	minor int,
+	device string,
+	path string,
+	fs string,
+	flags uintptr,
+	data string,
+	timeout int,
+	opts map[string]string) error {
```
adityadani: nit:
opts map[string]string,
) error {

PR #377, opened 2018-03-26
minor edit to cluster manager boot sequence

sdeoras:

minor edit to previous PR 376


lpabon:
Just confused as to why in this is using `c.kv` to initialize, but kvdb is not initialized until line https://github.com/libopenstorage/openstorage/pull/377/files#diff-b019ace196f6e18492cea0d691b1185eL1108

sangleganesh:
yes, thats what i pointed to saurabh. we should just use kvdb.Instance() and probably get rid of c.kv (probably in a separate PR).

sdeoras:
@lpabon the kvdb set instance is apparently just making it into an instance. but i agree, this was confusing to me too. to be safe, let's move the set instance code bit up in the sequence

sdeoras:
@sangleganesh and @lpabon I moved it up in sequence and added a line updating c.kv as well

Issue #351, opened 2018-03-12
Feature: Secrets API


ram-infrac:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE

# Reason needed
To create OSD-SDK
Move internal PX API to OSD to create external API SDK. This is first API set  we will move to OSD.
# 1. Manage secret commands
Secret commands allows user to manage secret keys. You can encrypt volumes using given secret key.
# a. Set Cluster Secret Key;
Cluster wide secret key is basically a key value pair where the value part is the secret that is used as a passphrase for encrypting volumes. A cluster wide secret key is a common key that can be used to encrypt all volumes.
* **Method**: POST
* **Endpoints**: /setClusterSecretKey
* **content-type**: application/json
* **Response HTTP Status code**: 200
* **Params**:
  clustersecretkey(string): Cluster wide Secret Key
  override(bool) : override previous secret key [true/false]

**Example Request**:
```POST /v1/setClusterSecretKey```
```sh
Content-type : application/json
{
	clustersecretkey: "testkey1",
	override: true
}
```
**Example Response**:
Cluster Secret Key successfully set.

# b. Login to Secret
 Allow login to secure secret store e.g vault, kvdb,aws etc

* **Method**: Post
* **Endpoints**: /secretsLogin
* **Response HTTP Status code**: 200
* **Params**:
secret(int): type of secret (kvdb - 1, vault - 2, aws - 3, Docker - 4, kubernetes -5, default - 0)
secretConfig(string): config for secret login or null

**Example Request**:

```POST /v1/secretsLogin```
```sh
Content-type : application/json
{
secret:2,
secretConfig: null
}
```
**Example Response**:
Secrets Login Succeeded

# c. Set value for given secret key
stores the data against the given key in the secret store
* **Method**: PUT
* **Endpoints**: /setSecret
* **Content-Type**: application/json
* **Response HTTP Status code**: 200
* **Params**:
secretid(string) : key for secret value.
secretvalue(string) : value for secret passpharase.

**Example Request**:
```POST /v1/setSecret```
```sh
Content-type : application/json
{
secretid : testkey1,
secretvalue :testval1
}
```
**Example Response**:
Secret succuessfully set.
# d. Get value of secret key
retrieves the data for the given key.
* **Method**: GET
* **Endpoints**: /getSecret
* **Content-Type**:application/json
* **Response HTTP Status code**: 200
* **Params**:
  secretid(string) : passphrase whose value to be retrived.

[...]

PR #379, opened 2018-03-30
Fix for #351 feature OSD-SDK

ram-infrac:

This is PR for #351 . This is work in progress which will add SecretAPI endpoints in OSD with default `Not implemented` status for secret endpoints.
Closes #


api/server/server.go
```diff
@@ -137,6 +139,12 @@ func startServer(name string, sockBase string, port uint16, routes []*Route) err
 	for _, v := range routes {
 		router.Methods(v.verb).Path(v.path).HandlerFunc(v.fn)
 	}
+	//Add only to cluster routes
+	clusterApi := newClusterAPI()
+	sm := clusterApi.RegisterSecretManager(fake.New())
+	for _, v := range sm.Routes() {
```
ram-infrac: How to add append secrets.Route() into clusterAPI.Routes()?

api/server/cluster.go
```diff
@@ -711,3 +712,8 @@ func handleResourceType(resource string) (api.ResourceType, error) {
 		return api.ResourceType_RESOURCE_TYPE_NONE, fmt.Errorf("Invalid resource type")
 	}
 }
+
+func (c *clusterApi) RegisterSecretManager(sec secrets.SecretManager) *secrets.Manager {
```
ram-infrac: is this right place to add RegisterSecretManager()?

lpabon: Yes since it is a receiver for (*clusterAPI)

lpabon: Why does this return a secrets manager? For possible chaining?

api/server/server.go
lpabon: Ah, I see the issue. Routes are not a property of the SecretManager(), they are a property of the server. The SecretManager is an implementation of the handlers to these routes, but the routes are not a property of the SecretManager. For this reason, the `routes` in the arguments should have been filled in already by the caller.

secrets/secrets.go
```diff
@@ -0,0 +1,172 @@
...
+	// SetClusterSecretKey sets the cluster wide secret key
+	SetClusterSecretKey(secretKey string, override bool) error
+	//CheckSecretLogin validates session with secret store
+	CheckSecretLogin() error
+	//SetSecret sets secret key against data
+	SetSecret(key string, value string) error
+	// GetSecret retrieves the data for the given feature and key
+	GetSecret(key string) (string, error)
+	//Routes to the secret API
+	Routes() []*Route
```
lpabon: Remove. Routes are specific to a REST service. A SecretManager is an implementation of an interface which does not matter how data gets to it. For example, another day we could change to gRPC or whatever method of communication, but the SecretManager would be the same interface implementation.

secrets/secrets.go
```diff
@@ -0,0 +1,172 @@
...
+	lock   sync.Mutex
+}
+
+func NewSecretManager(sec SecretManager) *Manager {
+	return &Manager{
+		secret: sec,
+	}
+}
+
+func (m *Manager) Routes() []*Route {
```
lpabon: These need to be moved to the server which has the handlers

secrets/secrets.go
```diff
@@ -0,0 +1,172 @@
...
+func secretVersion(route, version string) string {
+	if version == "" {
+		return "/" + route
+	} else {
+		return "/" + version + "/" + route
+	}
+}
+
+//TODO: Add swagger yaml
+func (m *Manager) setClusterSecretKey(w http.ResponseWriter, r *http.Request) {
```
lpabon: This is not the way it should be done. It should be:

```
[ route ] 
    --> [ server handler function (generic) which then calls the interface implementation ] 
          --> [ SecretManager (in this file, it would the fake one you have) ]
```

api/server/server.go
ram-infrac: so should I move routes back into api/cluster/routes.go like this 
```{verb: "GET", path: clusterVersion("checksecretslogin", cluster.APIVersion), fn: c.secretLoginCheck},```
->https://github.com/ram-infrac/openstorage/commit/4078f23de524b46b22ef314df58bb20a1a6d05fe#diff-17b630a1004c86897f2252802622cf29R34.

File: secrets/secrets.go (hunk 5 above)
Comment (ram-infrac) [reply secrets/secrets.go]: To summarise : 
Routes:
```{verb: "GET", path: clusterVersion("checksecretslogin", cluster.APIVersion), fn: c.secretLoginCheck}```
https://github.com/ram-infrac/openstorage/commit/4078f23de524b46b22ef314df58bb20a1a6d05fe#diff-17b630a1004c86897f2252802622cf29R34.
 
Handler:
```func (c *clusterApi) setClusterSecretKey(w http.ResponseWriter, r *http.Request) {```

Interface implementation called from handler as,
```secretValue, err := SecretManager.Secret.GetSecret(secretid)```
https://github.com/ram-infrac/openstorage/commit/4078f23de524b46b22ef314df58bb20a1a6d05fe#diff-b6913eea9e933617095ce95032187c81R109

If anyone wishes to override his own secret implementer he can 
```clusterApi.RegisterSecretManager(fake.New())```
https://github.com/ram-infrac/openstorage/commit/4078f23de524b46b22ef314df58bb20a1a6d05fe#diff-44a527ea72ea9572e1cf71c7f1e2a0f0R121

File: api/server/cluster.go [hunk 6]
```diff
@@ -20,10 +21,18 @@ const (
...
-func newClusterAPI() restServer {
-	return &clusterApi{restBase{version: cluster.APIVersion, name: "Cluster API"}}
+type ClusterServerConfiguration struct {
+	ConfigSecretManager *secrets.Manager
+}
+
+func newClusterAPI(config ClusterServerConfiguration) restServer {
+	return &clusterApi{restBase: restBase{version: cluster.APIVersion,
+		name: "Cluster API"},
+		SecretManager: config.ConfigSecretManager,
```
lpabon: Please format correctly:

```go
return &clusterApi{
    restBase: restBase{
        version: cluster.APIVersion,
	name: "Cluster API"
    },
    SecretManager: config.ConfigSecretManager,
}
```

api/server/routes.go
```diff
@@ -31,5 +31,10 @@ func (c *clusterApi) Routes() []*Route {
 		{verb: "POST", path: clusterPath(client.UriNode, cluster.APIVersion), fn: c.setNodeConf},
 		{verb: "DELETE", path: clusterPath(client.UriNode+"/{id}", cluster.APIVersion), fn: c.delNodeConf},
 		{verb: "GET", path: clusterPath("/getnodeidfromip/{idip}", cluster.APIVersion), fn: c.getNodeIdFromIp},
+		{verb: "GET", path: clusterVersion("checksecretlogin", cluster.APIVersion), fn: c.secretLoginCheck},
```
lpabon: Since these are all secret REST command, place them all under `/secret`. Make a new function called `clusterSecretPath("checklogin", cluster.APIVersion)` where it returns `/v1/cluster/secret/checklogin`. Then do the same for the others.

api/server/secrets.go
```diff
@@ -0,0 +1,125 @@
...
+		return
+	}
+
+	err := c.SecretManager.Secret.SetClusterSecretKey(secretKey, override)
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Write([]byte("Cluster" + secretKeyOkMsg + "\n"))
```
lpabon: This should be returned as a JSON packet. Create a SetClusterSecretKeyResponse (see other examples) and use that struct to return the information.

Do this for all calls.

api/server/secrets.go
```diff
@@ -0,0 +1,125 @@
...
+const (
+	secretKeyOkMsg   = "Secret Key set successfully."
+	secretLoginOkMsg = "Secrets Login Succeeded"
+	secretLoginCheck = "Secrets Login Check Succeeded"
+)
+
+//TODO: Add swagger yaml
+func (c *clusterApi) setClusterSecretKey(w http.ResponseWriter, r *http.Request) {
+
+	params := r.URL.Query()
```
lpabon: Do not use URL Query. Instead either get a JSON packet from the request or use mux.Vars() to get any vars in the URL, like ID.

api/server/cluster.go
ram-infrac: ok, will do. thanks

ram-infrac: also unit test failure is occured since signature of newClusterAPI() is changed, will do necessary changes with help of imran on this

api/server/routes.go
ram-infrac: ok, also should I make them more REST like.
e.g instead of /setclustersecretkey , POST call on /clustersecretkey,
similarly /checksecretlogin,- > GET on /secretlogin -> allow to check secrets  session is vaild.
and POST on /secretlogin will allow you to store secrets in secrets store.

Note: Earlier one was same as PX's

secrets/fake/fake.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/libopenstorage/openstorage/secrets"
+)
+
+var (
+	// ErrNotImplemented default secrets in OSD
+	ErrNotImplemented = errors.New("Not Implemented")
+)
+
+type Fake struct {
+	secrets.SecretManager
```
piyush-nimbalkar: you don't need to add this
it will automatically implement the SecretManager interface if you implement all the methods

secrets/fake/fake.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// New returns a new Fake secret implementation
+func New() *Fake {
```
piyush-nimbalkar: could return secrets.SecretManager here, instead of Fake

secrets/fake/fake.go
```diff
@@ -0,0 +1,40 @@
...
+	"errors"
+	"github.com/libopenstorage/openstorage/secrets"
+)
+
+var (
+	// ErrNotImplemented default secrets in OSD
+	ErrNotImplemented = errors.New("Not Implemented")
+)
+
+type Fake struct {
```
piyush-nimbalkar: nit: can we call this DefaultSecretManager instead Fake?
Fake sounds like a testing object; like stubs and mocks

secrets/secrets.go
```diff
@@ -0,0 +1,27 @@
...
+type SecretManager interface {
+	// Login create session with secret store
+	SecretLogin(secretType int, secretConfig map[string]string) error
+	// SetClusterSecretKey sets the cluster wide secret key
+	SetClusterSecretKey(secretKey string, override bool) error
+	//CheckSecretLogin validates session with secret store
+	CheckSecretLogin() error
+	//SetSecret sets secret key against data
+	SetSecret(key string, value string) error
+	// GetSecret retrieves the data for the given feature and key
```
piyush-nimbalkar: remove "given feature"

secrets/secrets.go
```diff
@@ -0,0 +1,27 @@
+package secrets
+
+import "sync"
+
+type SecretManager interface {
```
piyush-nimbalkar: the "manager" term is over used and is getting very confusing
can we call this interface as Secrets? may be even rename the methods to Login, Set, Get, CheckLogin, SetDefaultKey? what do you think?

secrets/secrets.go
```diff
@@ -0,0 +1,27 @@
...
+	// Login create session with secret store
+	SecretLogin(secretType int, secretConfig map[string]string) error
+	// SetClusterSecretKey sets the cluster wide secret key
+	SetClusterSecretKey(secretKey string, override bool) error
+	//CheckSecretLogin validates session with secret store
```
piyush-nimbalkar: nit: format the comment and other such places (space after // )

secrets/fake/fake.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (f *Fake) SetClusterSecretKey(secretKey string, override bool) error {
+	return ErrNotImplemented
+}
+
+func (f *Fake) CheckSecretLogin() error {
+	return ErrNotImplemented
+}
+
+func (f *Fake) SetSecret(secretKey string, secretValue string) error {
```
piyush-nimbalkar: @adityadani, currently we use a map of key-value pairs because a some secrets stores can accept multiple key value pairs. This makes it more flexible to store any kind of data. Should we keep this interface generic as well?

api/client/cluster/client.go
```diff
@@ -192,3 +193,68 @@ func (c *clusterClient) EraseAlert(resource api.ResourceType, alertID int64) err
...
+	request.QueryOption(secrets.OverrideSecrets, stOverride)
+	resp := request.Do()
+	if resp.Error() != nil {
+		return resp.FormatError()
+	}
+	return nil
+}
+
+func (c *clusterClient) SetSecret(secretID string, secretValue string) error {
+	request := c.c.Put().Resource("/setsecret")
```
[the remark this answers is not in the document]
ram-infrac: agreed, i kept it in line with way PX call it.

api/server/secrets.go
```diff
@@ -0,0 +1,125 @@
...
+	method := "setClusterSecretKey"
+	secretKey := params["clustersecretkey"][0]
```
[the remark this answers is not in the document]
ram-infrac: Should this be passed as query string or as body?

secrets/fake/fake.go
adityadani: may be call it as a NullSecretManager.

secrets/fake/fake.go
adityadani: Yes. Lets use secretValue as an interface{}

secrets/secrets.go
adityadani: If we want to go by the golang idiomatic way, then
this package is already called secrets. So this interface can be called just Manager and the APIs would be Login, Set, Get etc

So the caller of this package would do this
```
secretsManager := secrets.NewManager()
secretsManager.Login
```

secrets/fake/fake.go
adityadani: this struct can be a private one. I don't see any reason why it needs to be public.

secrets/struct.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+// SecretLoginRequest specify secret store and config to initiate
+// secret store session
+// swagger: parameters secret
+type SecretLoginRequest struct {
+	SecretConfig map[string]string
+}
+
+type Route struct {
```
adityadani: Do we need this here? We have it already defined in api/server/routes.go?

secrets/secrets.go
```diff
@@ -0,0 +1,27 @@
...
+	CheckSecretLogin() error
+	//SetSecret sets secret key against data
+	SetSecret(key string, value string) error
+	// GetSecret retrieves the data for the given feature and key
+	GetSecret(key string) (string, error)
+}
+
+type Manager struct {
```
adityadani: nit: Manager can be private?

secrets/fake/fake.go
ram-infrac: so package structure should also change, secrets/fake/fake.go to say secrets/defaultsecretmanager(or nullsecretmanager). How about secrets/manager/defaultmanager.go?

api/server/secrets.go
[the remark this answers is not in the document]
piyush-nimbalkar: I think it should be part of the body, similar to how it should be for SetSecret() interface.

secrets/secrets.go
piyush-nimbalkar: Right, this could be private and you can use the Manager interface in the cluster api struct.

ram-infrac: How can we access Manager struct when called from StartClusterAPI() if we made it private?

api/server/cluster.go
```diff
@@ -686,6 +698,10 @@ func clusterVersion(route, version string) string {
 	return "/" + version + "/" + route
 }
 
+func clusterSecretPath(route, version string) string {
+	return clusterVersion("cluster/secrets"+route, version)
```
piyush-nimbalkar: we can use clusterPath here
`return clusterPath("/secrets"+route, version)`

Also can we use the same path constants (secretPath - /secrets) that we used in client. May be define in server and then client uses that?

api/server/secrets.go
```diff
@@ -0,0 +1,147 @@
...
+	var secResp secrets.SecretStatusResponse
+
+	if err := json.NewDecoder(r.Body).Decode(&secReq); err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	err := c.SecretManager.SetDefaultSecretKey(
+		secReq.DefaultSecretKey,
+		secReq.Override)
```
piyush-nimbalkar: nit:
```
SetDefaultSecretKey(
    req.DefaultSecretKey,
    req.Override,
)

File: api/server/secrets.go [hunk 23]
```diff
@@ -0,0 +1,147 @@
...
+
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	secResp.Status = "Cluster" + secretKeyOkMsg
```
piyush-nimbalkar: *Default instead of Cluster. Add a space after that before the message.

cluster/manager.go
```diff
@@ -1693,3 +1693,36 @@ func (c *ClusterManager) DeleteNodeConf(nodeID string) error {
 func (c *ClusterManager) EnumerateNodeConf() (*osdconfig.NodesConfig, error) {
 	return c.configManager.EnumerateNodeConf()
 }
+
+//This needs to be added since clusterclient is REST wrapper over ClusterManager
```
piyush-nimbalkar: nit: space after //, everywhere in this file

secrets/struct.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// ClusterSecretKeyRequest  specify request to set cluster secret key
+// swagger: parameters clusterKey
+type DefaultSecretKeyRequest struct {
+	DefaultSecretKey string
+	Override         bool
+}
+
+// SetsecretsLogin setsecrets
```
piyush-nimbalkar: add a description as well

secrets/struct.go
```diff
@@ -0,0 +1,56 @@
...
+	DefaultSecretKey = "defaultsecretkey"
+	OverrideSecrets  = "override"
+	SecretKey        = "id"
+	SecretValue      = "secretvalue"
+	SecretType       = "store"
+	APIVersion       = "v1"
+)
+
+const (
+	TypeNone   = "None"
```
piyush-nimbalkar: nit: none could also be small case

secrets/struct.go
```diff
@@ -0,0 +1,56 @@
+package secrets
+
+const (
+	DefaultSecretKey = "defaultsecretkey"
+	OverrideSecrets  = "override"
+	SecretKey        = "id"
+	SecretValue      = "secretvalue"
+	SecretType       = "store"
```
piyush-nimbalkar: Let's not introduce another keyword (store) in the APIs. Anyway every where we are calling it Type, so lets call it "type"

secrets/struct.go
```diff
@@ -0,0 +1,56 @@
...
+
+// SecretLoginRequest specify secret store and config to initiate
+// secret store session
+// swagger: parameters secret
+type SecretLoginRequest struct {
+	SecretType   string
+	SecretConfig map[string]string
+}
+
+// SecretLoginResponse whether login is successful or not
```
piyush-nimbalkar: *SecretStatusResponse
nit: can we call it just SecretResponse?

secrets/struct.go
```diff
@@ -0,0 +1,56 @@
...
+	TypeVault  = "vault"
+	TypeAWS    = "aws-kms"
+	TypeDocker = "docker"
+	TypeK8s    = "k8s"
+	TypeDCOS   = "dcos"
+)
+
+// SecretLoginRequest specify secret store and config to initiate
+// secret store session
+// swagger: parameters secret
```
piyush-nimbalkar: update the parameters as well as all the doc strings in this file.

secrets/secrets.go
```diff
@@ -0,0 +1,69 @@
...
+	lock sync.Mutex
+}
+
+func NewSecretManager(sec Secrets) *Manager {
+	return &Manager{
+		Secrets: sec,
+	}
+}
+
+type nullSecretManager struct {
```
piyush-nimbalkar: nit: can we call this nullSecrets instead of manager, as it is an implementation of Secrets interface?

secrets/secrets.go
```diff
@@ -0,0 +1,69 @@
...
+
+var (
+	// ErrNotImplemented default secrets in OSD
+	ErrNotImplemented = errors.New("Not Implemented")
+)
+
+type Secrets interface {
+	// Login create session with secret store
+	Login(secretType string, secretConfig map[string]string) error
+	// DefaultSecretKey  sets the cluster wide secret key
```
piyush-nimbalkar: *SetDefaultSecretKey
Update the comments everywhere to match the method name

secrets/secrets.go
```diff
@@ -0,0 +1,69 @@
...
+	SetDefaultSecretKey(secretKey string, override bool) error
+	// GetDefaultSecretKey returns cluster wide secret key's value
+	GetDefaultSecretKey() (interface{}, error)
+	// CheckSecretLogin validates session with secret store
+	CheckLogin() error
+	// SetSecret sets secret key against data
```
piyush-nimbalkar: nit: stores the given value/data against the key

api/client/cluster/client.go
```diff
@@ -192,3 +194,78 @@ func (c *clusterClient) EraseAlert(resource api.ResourceType, alertID int64) err
...
+	path := clusterPath + secretPath + "/defaultsecretkey"
+	request := c.c.Get().Resource(path)
+	err := request.Do().Unmarshal(defaultKeyResp)
+	if err != nil {
+		return defaultKeyResp, err
+	}
+	return defaultKeyResp, nil
+}
+
+func (c *clusterClient) Set(secretID string, secretValue interface{}) error {
```
lpabon: Cannot use names `Set` and `Get` for `clusterClient` because it is too generic, when the API is only affecting the secret. I think this is supposed to be `SecretSetValue` and `SecretGetValue`

lpabon: Actually, preface all of these API calls with the `SecretXXX` name... That will group them into that "namespace". So it would look like `CloudBackupXXX` functions in `client/volume/client.go`

api/server/cluster.go
```diff
@@ -20,10 +21,21 @@ const (
 
 type clusterApi struct {
 	restBase
+	SecretManager *secrets.Manager
 }
 
-func newClusterAPI() restServer {
-	return &clusterApi{restBase{version: cluster.APIVersion, name: "Cluster API"}}
+type ClusterServerConfiguration struct {
```
lpabon: Could you add comments to this to why it is used and to its variables so that the callers know how to use it?

api/server/secrets.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/libopenstorage/openstorage/secrets"
+)
+
+const (
+	secretKeyOkMsg   = "Secret Key set successfully"
+	secretLoginOkMsg = "Secrets Login Succeeded"
+	secretLoginCheck = "Secrets Login Check Succeeded"
+)
+
+// TODO: Add swagger yaml
```
lpabon: You can probably start adding this soon.

api/server/secrets.go
```diff
@@ -0,0 +1,148 @@
...
+		secReq.DefaultSecretKey,
+		secReq.Override,
+	)
+
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	secResp.Status = "Default " + secretKeyOkMsg
```
lpabon: Unless this is the way it is supposed to be expected, if there is no returned data expected, a returned HTTP status of 200 is all that is needed. So line 38-39 may not be necessary. Just need to ensure we are returning 200.

Again, that is as long as no data is expected in return, because it looks like to me that line 38 is just returning constant message.

api/server/secrets.go
```diff
@@ -0,0 +1,148 @@
...
+	method := "secretsLogin"
+
+	if err := json.NewDecoder(r.Body).Decode(&secReq); err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	err := c.SecretManager.Login(secReq.SecretType, secReq.SecretConfig)
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
```
lpabon: This is probably not a http.StatusInternalServerError since it may have not been able to Login due to invalid information from the user. Instead you can check if `Login()` returns a known `ErrXXX` value and return to the user something like `http.StatusForbidden` if there credentials did not work.

api/server/secrets.go
```diff
@@ -0,0 +1,148 @@
...
+	json.NewEncoder(w).Encode(secResp)
+}
+
+// TODO: Add swagger yaml
+func (c *clusterApi) setSecret(w http.ResponseWriter, r *http.Request) {
+
+	method := "setSecret"
+	var secReq secrets.SetSecretRequest
+	var secResp secrets.SecretResponse
+	params := r.URL.Query()
```
lpabon: Passing the secret key over the URI? Can we instead pass it in the JSON body? The URI has to be encoded through URL functions, but not when passing through JSON body.

api/server/secrets.go
```diff
@@ -0,0 +1,148 @@
...
+
+	secResp.Status = secretKeyOkMsg
+	json.NewEncoder(w).Encode(secResp)
+}
+
+// TODO: Add swagger yaml
+func (c *clusterApi) getSecret(w http.ResponseWriter, r *http.Request) {
+
+	method := "getSecret"
+	params := r.URL.Query()
```
lpabon: same here.  Or here we can use the Gorilla mux to get the value from the URI as a path but not as a query.

For example instead of:

```go
{verb: "GET", path: clusterSecretPath("", cluster.APIVersion), fn: c.getSecret},
{verb: "PUT", path: clusterSecretPath("", cluster.APIVersion), fn: c.setSecret},
```

it could be 

```go
{verb: "GET", path: clusterSecretPath("/{secretKey}", cluster.APIVersion), fn: c.getSecret},
{verb: "PUT", path: clusterSecretPath("/{secretKey}", cluster.APIVersion), fn: c.setSecret},
```

where in `PUT` it passes a JSON object containing the secret value

secrets/secrets.go
```diff
@@ -0,0 +1,69 @@
+package secrets
+
+import (
+	"errors"
+	"sync"
+)
+
+var (
+	// ErrNotImplemented default secrets in OSD
+	ErrNotImplemented = errors.New("Not Implemented")
```
lpabon: Add something like `ErrAccessDenied` returned by `Login()` when the credentials are not accepted. Then this valued can be checked for in the login function in the REST server to see it `Login()` returned an error because access was denied or some other reason. If any other reason other than ErrAccessDenied, then the code should return http.statusinteralerror.

Also, please coordinate this error with those you implemented the Login() function in Px. it may be they already have an error value we could use here.

api/server/secrets.go
ram-infrac: yes Luis, since PR this PR is getting bulky I was planning to add mock test cases and swagger yaml  in new PR once this get merged.

api/server/secrets.go
ram-infrac: @adityadani In my understanding no return data is expected for setting secrets, is should be ok if we return only HTTP status 200?

api/client/cluster/client.go
ram-infrac: @piyush-nimbalkar , @adityadani should I change interface Set and Get back to SetValue/GetValue?

api/server/secrets.go
ram-infrac: but using mux will contradict with other api's ?
Right now we have API as,
```
GET /cluster/secrets/<secret_id>
PUT /cluster/secrets/<secret_id> Body: {value: }
GET /cluster/secrets/defaultsecretkey
PUT /cluster/secrets/defaultsecretkey Body: {value: }
GET /cluster/secrets/verify
POST /cluster/secrets/login Body: {}
```
so both call to GET /cluster/secrets/myid and GET /cluster/secrets/defaultsecretkey  will ultimately goes to route ```{verb: "GET", path: clusterSecretPath("/{secretKey}", cluster.APIVersion), fn: c.getSecret},```

Am I wrong here?

api/server/secrets.go
piyush-nimbalkar: I think set calls could just return an status code with empty response body. If we return anything it should be the set secret key-value, to be RESTful. But I think, we should go with just status code for now.

secrets/secrets.go
piyush-nimbalkar: I think ErrUnauthorized/ErrNotAuthenticated is more precise than AccessDenied in case of a login. AccessDenied is more like a 403 (Forbidden) which should be thrown on accessing a certain resource.

secrets/secrets.go
```diff
@@ -0,0 +1,79 @@
...
+	// ErrSecretExists returned when a secret for the given secret id already exists
+	ErrSecretExists = errors.New("Secret ID already exists")
+)
+
+type Secrets interface {
+	// SecretLogin create session with secret store
+	SecretLogin(secretType string, secretConfig map[string]string) error
+	// SecretSetDefaultSecretKey  sets the cluster wide secret key
+	SecretSetDefaultSecretKey(secretKey string, override bool) error
+	// SecretGetDefaultSecretKey returns cluster wide secret key's value
```
piyush-nimbalkar: this does not return the key's value, but the cluster wide key itself

secrets/struct.go
```diff
@@ -0,0 +1,53 @@
...
+type SetSecretRequest struct {
+	SecretValue interface{}
+}
+
+// GetSecretResponse gets secret value for given key
+type GetSecretResponse struct {
+	SecretValue interface{}
+}
+
+// GetDefaultSecretResponse gets default secret value for given key
```
piyush-nimbalkar: *gets default secret key (remove 'for given key')

secrets/struct.go
```diff
@@ -0,0 +1,53 @@
...
+	SecretConfig map[string]string
+}
+
+// SecretResponse whethere login is successful or not
+type SecretResponse struct {
+	Status string
+}
+
+// DefaultSecretKeyRequest specify request to set cluster secret key
+type DefaultSecretKeyRequest struct {
```
piyush-nimbalkar: SetDefaultSecretKeyRequest? to be consistent with GetDefaultSecretResponse?
Or remove the GetDefaultSecretResponse. I think it's not being used anymore.

secrets/struct.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+// SecretLoginRequest specify secret store and config to initiate
+// secret store session
+type SecretLoginRequest struct {
+	SecretType   string
+	SecretConfig map[string]string
+}
+
+// SecretResponse whethere login is successful or not
```
piyush-nimbalkar: we don't need this anymore, right?

api/server/secrets.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// TODO: Add swagger yaml
+func (c *clusterApi) secretsLogin(w http.ResponseWriter, r *http.Request) {
+	var secReq secrets.SecretLoginRequest
+	method := "secretsLogin"
+
+	if err := json.NewDecoder(r.Body).Decode(&secReq); err != nil {
+		if err == secrets.ErrNotAuthenticated {
+			c.sendError(c.name, method, w, err.Error(), http.StatusForbidden)
```
piyush-nimbalkar: should return a StatusUnauthorized - 401

ram-infrac: updated

secrets/struct.go
ram-infrac: right, removed

secrets/struct.go
ram-infrac: removed

secrets/struct.go
ram-infrac: Updated

secrets/secrets.go
ram-infrac: updated comment

api/server/secrets.go
lpabon: You are correct here. Ignore my request.

lpabon:
I will review tomorrow. in the meantime, please resolve the conflict.

api/server/secrets.go
ram-infrac: ok, thanks Luis

api/server/secrets.go
```diff
@@ -0,0 +1,138 @@
+package server
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/libopenstorage/openstorage/secrets"
+)
+
+// TODO: Add swagger yaml
```
lpabon: swagger still needs to be done.

api/server/secrets.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		if err == secrets.ErrNotAuthenticated {
+			c.sendError(c.name, method, w, err.Error(), http.StatusUnauthorized)
+			return
+		}
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
lpabon: awesome 👍

api/server/secrets.go
```diff
@@ -0,0 +1,138 @@
...
+// TODO: Add swagger yaml
+func (c *clusterApi) setSecret(w http.ResponseWriter, r *http.Request) {
+
+	method := "setSecret"
+	var secReq secrets.SetSecretRequest
+	params := r.URL.Query()
+	secretID := params[secrets.SecretKey]
+
+	if len(secretID) == 0 || secretID[0] == "" {
+		c.sendError(c.name, method, w, "Missing secret ID", http.StatusInternalServerError)
```
lpabon: This should be a status error of 400 (invalid requets) not a server error of 500 (StatusInternalServerError is a HTTP error) since it was caused by the client. See https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

api/server/secrets.go
```diff
@@ -0,0 +1,138 @@
...
+
+// TODO: Add swagger yaml
+func (c *clusterApi) getSecret(w http.ResponseWriter, r *http.Request) {
+
+	method := "getSecret"
+	params := r.URL.Query()
+	secretID := params[secrets.SecretKey]
+
+	if len(secretID) == 0 || secretID[0] == "" {
+		c.sendError(c.name, method, w, "Missing secret ID", http.StatusInternalServerError)
```
lpabon: same here. Please check all of the input calls to the server making sure they return the correct HTTP status.

ram-infrac: updated

api/server/secrets.go
ram-infrac: updated

api/server/testutils_test.go
```diff
@@ -52,6 +55,11 @@ func newTestCluster(t *testing.T) *testCluster {
 	// Create a new mock cluster
 	tester.c = mockcluster.NewMockCluster(tester.mc)
 
+	// Create a new mock Serets
+	// This doesn't set *mocksecrets.Mocksecrets, thats why needed to
+	// write newTestClusterSecrets() and newTestServerSecrets()
+	//tester.sm = mocksecrets.NewMockSecrets(tester.mc)
```
ram-infrac: @lpabon ,  is there any way to achieve this, thus avoid writing different  newTestClusterSecrets() and newTestServerSecrets()?

lpabon: @ram-infrac Hi, I'm not sure what you mean. Could you explain the pros and cons or the issue? What is the difference between newTestServerSecret and newTestClusterSecret? Is newTestServerSecret needed?

lpabon:
🎉 Congratulations @ram-infrac

ram-infrac:
Thanks @lpabon :smile:

PR #385, opened 2018-04-04
goclient: Get error from HTTP body

### LINKED ISSUE #372: Error information is not correctly retrieved by the golang client
BUG

**What happened**:
See: https://github.com/libopenstorage/openstorage/pull/350/files#diff-7d1e2910f3cef1ae679f6a61d7b7325cR211 in #350 

The issue is that the function parseHTTPstatus is expecting the error string to be encoded in the body as a `Status` Json object. The issue is that some, or most, of the handlers in the server the function sendError() which in turn uses `http.Error(w, err)` to write the error as a raw string in the body.

With the error being raw string, parseHTTPstatus currently does not retreive the error correctly and thus it returns the generic `HTTP xxxx status`.

The client functions to return the error string set by the server handler.

**Proposals**:

1. Support both a status json object, or fall back to reading the string and returning it if length is != 0
2. Standardize on the status type for ALL OSD api calls

My vote is for number 1 now and number 2 for next release.

### LINKED ISSUE #378: Need to standardize model of return errors over REST commands
BUG

**What happened**:
@ipochi found during #350 that the client call `parseHTTPStatus()` expected errors to be encoded in the body of the HTTP payload a s JSON Status object. On the other hand, the server calls use `http.Error()` to save the error raw in the body as it is normally done on REST.

This issue causes that the client calls do not get the correct error from the server. Example:

```go
err := SomeClientCall()
fmt.Println(err)
```

would show

```
HTTP error 500
```

instead of 

```
Some error set by the server as a error type
```

## Proposal

Instead of expecting a status JSON object, expect all HTTP errors (4xx/5xx) to have the error stored as a string in the body

### LINKED ISSUE #384: /cluster/status does not work
BUG

**What happened**:
The client Golang function `cluster.Status()` does not work because it is expecting the server to return a JSON encoded `api/client/requets.go:Status`, when in fact it is returning a number _raw_ on the body because the server is returning the value of `cluster.Status` which is an _int_.

The client needs to return what the server has provided.

Start a server then run a `curl -X GET -i 'http://<address>/v1/cluster/status'`

lpabon:
Golang clients where unable to get the actual error set by the server handlers and implementations.

Closes #372
Closes #350
Closes #384
Closes #378

It has been found that there was no caller of the `client.Status()` function in osd or in px and it has been removed since it did not work anyways.


lpabon:
This replaces the work from #350 and includes those unit tests here.

api/server/cluster_test.go
```diff
@@ -572,6 +573,252 @@ func TestGetNodeIdFromIpFailed(t *testing.T) {
...
+}
+
+func TestSetClusterConfFailed(t *testing.T) {
+
+	// Create a new global test cluster
+	ts, tc := testClusterServer(t)
+	defer ts.Close()
+	defer tc.Finish()
+
+	secretsConfig := &osdconfig.SecretsConfig{
```
sdeoras: recommend to use new(type).Init() syntax available for osdconfig structs, but what you have here should work as well.

lpabon: Interesting, that is normally not idiomatic golang. The method I used is the one used normally. But both work.

Issue #383, opened 2018-04-03
Configuration library be exposed over REST?


lpabon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
Question

**What happened**:
Should the configuration library be exposed over REST? That would mean that anyone with the REST client can manipulate metadata in the configuration. This may lead to many issues.

sdeoras:
@lpabon : FYI, we did discuss this topic a few times already and have implemented both methods to interface with osdconfig, one via REST and other directly through kvdb. I believe you _do_ want users to be able to change a few params, which was the original intent that @harsh-px brought to me a few times. I would like to know which parameters are sensitive.

sdeoras:
closing this per internal discussion (PWX-5154)

PR #390, opened 2018-04-06
Reorganize to create cluster.ClusterClient

### LINKED ISSUE #382: Cluster Client golang implementation should have a subset Interface
FEATURE

**What happened**:
The current golang client implementation does not implement many of the cluster.Cluster interfaces and probably never will. Instead we should create a _ClusterClient_ struct or interface which is specifically tailored for the REST caller.

Here are the functions which are not implemented:

```go
func (c *clusterClient) AddEventListener(cluster.ClusterListener) error {
	return nil
}

func (c *clusterClient) UpdateData(nodeData map[string]interface{}) error {
	return nil
}

func (c *clusterClient) UpdateLabels(nodeLabels map[string]string) error {
	return nil
}

func (c *clusterClient) GetData() (map[string]*api.Node, error) {
	return nil, nil
}
func (c *clusterClient) NodeRemoveDone(nodeID string, result error) {
}

func (c *clusterClient) Shutdown() error {
	return nil
}

func (c *clusterClient) Start(int, bool, string) error {
	return nil
}
```

lpabon:
This change also makes the struct api/client/cluster.ClusterClient
exposed and moves the documentation there. We will be adding examples
and more documentation as time continues.

This reoganization of the interfaces and structs makes it simple for
developers to read the pages generated by GoDoc. When we add examples,
those examples could then be run right from the webpage using
play.golang.org as it is done in the official Golang packages
documentation.

Here is PDF [example of how the documentation looks like from GoDoc](https://github.com/libopenstorage/openstorage/files/1885418/cluster.-.The.Go.Programming.Language.pdf)

Closes #382


cluster/cluster.go
```diff
@@ -226,20 +209,34 @@ type ClusterAlerts interface {
...
 	SetSize(size int) error
 
+	// Remove node(s) from the cluster permanently.
+	Remove(nodes []api.Node, forceRemove bool) error
+
+	ClusterStatus
+	ClusterAlerts
+	ClusterData
+
+	// This may be removed from being exposed over REST: See #383
```
sdeoras: I made a comment on #383 and have created a JIRA 5154 to track decision on this as well.

api/client/cluster/client.go
```diff
@@ -18,20 +18,21 @@ const (
...
 
-type clusterClient struct {
-	c *client.Client
-}
+// Interface check
+var _ cluster.ClusterClient = &ClusterClient{}
 
-func newClusterClient(c *client.Client) cluster.Cluster {
-	return &clusterClient{c: c}
+// ClusterClient a REST of implementation of cluster.ClusterClient
```
piyush-nimbalkar: nit: *REST implementation

sangleganesh:
The reason the cluster client interface was composed from smaller interfaces is because that's how clients are using it. For example, parts of code which are using the cluster client to access and update cluster data do not need to know or access other parts of cluster client. Are there advantages to making it one big interface instead of being composed of smaller interfaces ?

lpabon:
@sangleganesh I'm not sure I understand your question, but I tried not to make too many changes. The large interface was there already, and seems to be the style of the previous authors.

I just tried to organize it so that the documentation is easier and the the interfaces exposed on the client are those we have chosen to have available from a client side. Therefore, those interfaces which as not to be exposed to the client are no longer part of the `ClusterClient` interface.

sangleganesh:
These interfaces are still being used.
For ex. ClusterData.UpdateData/GetData/ClusterRemove are used by portworx.

lpabon:
New change has the following:

```
Instead of deleting them, have ClusterRemove and ClusterData
be composed of exported and non-exported interfaces for compatability
with Portworx code.
```

PR #394, opened 2018-04-10
add consumer type to volume consumers

harsh-px:

It will be useful to identify what is the type of the volume consumer since name alone doesn't provide context. For e.g in k8s, the type would be a Pod.


lpabon:
This is ok for now, but if you find that this is happening too much (meaning you need to add more values to the message), instead make a `map[string]string` and use that for the information. We can then document the keys that are needed.

PR #399, opened 2018-04-17
Cleanup Volume Client REST Golang Interface

lpabon:
Just like #390, the Volume interface needed to be cleaned up.
VolumeDriver is the internal interface and the new VolumeClient
is the external exported interface. VolumeClient removed the unused
IO, GraphXXX, Shutdown, MountedAt, and other functions/interfaces.

Also removed Type, and Name from VolumeClient since they always
returned a constant and provided no real value.

Example of Godoc output attached.
[outputFile.pdf](https://github.com/libopenstorage/openstorage/files/1920729/outputFile.pdf)

### DIFF
--- a/api/client/volume/client.go
+++ b/api/client/volume/client.go
@@ -1,11 +1,7 @@
 package volume
 
 import (
-	"bytes"
 	"errors"
-	"fmt"
-	"io"
-	"io/ioutil"
 	"strconv"
 
 	"github.com/libopenstorage/openstorage/api"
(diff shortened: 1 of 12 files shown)


api/client/volume/client.go
```diff
@@ -135,14 +49,9 @@ func (v *volumeClient) Create(locator *api.VolumeLocator, source *api.Source,
...
 
-// Status diagnostic information
-func (v *volumeClient) Status() [][2]string {
-	return [][2]string{}
-}
-
 // Inspect specified volumes.
 // Errors ErrEnoEnt may be returned.
-func (v *volumeClient) Inspect(ids []string) ([]*api.Volume, error) {
+func (v *VolumeClient) Inspect(ids []string) ([]*api.Volume, error) {
```
piyush-nimbalkar: Not related to this PR, but it seems that the client takes a list of IDs and passes only one to the server (this last on in this slice). The server also expects only one ID. Should we change the implementation of server to take process multiple IDs.. anyway it is returning a list of volume objects.

volume/volume.go
```diff
@@ -84,6 +84,18 @@ type VolumeDriver interface {
 	Enumerator
 }
 
+// VolumeClient contains the exported functions to external clients.
+type VolumeClient interface {
+	StatsDriver
+	BlockDriver
+	Enumerator
+	SnapshotDriver
+	QuiesceDriver
```
harsh-px: In an offline discussion with @jvinod , he mentioned we can remove the quiesce calls from the API.

api/client/volume/client.go
lpabon: I think what it does here is that it runs through a for loop in line 60 adding them to the URI. I wonder what the max would be 😨 . Maybe you are correct, we should probably change this to a single volume. I'll create an issue to track this.

volume/volume.go
lpabon: Sweet!!

api/client/volume/client.go
lpabon: #401 created

volume/volume.go
lpabon: I will do it on the next patch. I created #402

PR #400, opened 2018-04-17
Add GroupSnapshot API

pault84:
Takes in a id (GroupId) and Labels (key:value)
If GroupId is set, will use groupid to find volumes.
If GroupID is not set will use Labels to find volumes (K8s will use this).

Returns a struct which has a list of SnapshotResponses and an error.


api/api.proto
```diff
@@ -579,3 +579,26 @@ message ActiveRequests {
   int64 RequestCount = 1;
   repeated ActiveRequest ActiveRequest = 2;
 }
+
+// GroupSnapCreateRequest specifies a request to create a snapshot of given group.
+// swagger:parameters snapVolumeGroup
+message GroupSnapCreateRequest {
+  // group id
+  string id = 1;
```
sangleganesh: nit: id is lower case here and below

api/client/volume/client.go
```diff
@@ -4,20 +4,22 @@ import (
...
-	volumePath = "/osd-volumes"
-	snapPath   = "/osd-snapshot"
-	credsPath  = "/osd-creds"
-	backupPath = "/osd-backup"
+	graphPath     = "/graph"
+	volumePath    = "/osd-volumes"
+	snapPath      = "/osd-snapshot"
+	credsPath     = "/osd-creds"
+	backupPath    = "/osd-backup"
+	volgroupsPath = "/osd-volgroups"
```
sangleganesh: nit: volgroupsPath not required .. ?

api/client/volume/client.go
```diff
@@ -658,3 +660,23 @@ func (v *volumeClient) CloudBackupSchedEnumerate() (*api.CloudBackupSchedEnumera
...
+
+func (v *volumeClient) SnapshotGroup(groupID string, labels map[string]string) (*api.GroupSnapCreateResponse, error) {
+
+	response := &api.GroupSnapCreateResponse{}
+	request := &api.GroupSnapCreateRequest{
+		Id:     groupID,
+		Labels: labels,
+	}
+
+	req := v.c.Post().Resource(snapPath + "/groupsnap").Body(request)
```
sangleganesh: nit: call it /snapshotgroup (so its consistent with api ?)

api/api.proto
pault84: when doing make Proto it becomes upper case

api/client/volume/client.go
pault84: correct will remove

api/client/volume/client.go
pault84: surething

api/api.proto
lpabon: Also, Could you change it to <group><call> style like CloudBackupDelete does it. So this would be `SnapshotGroupCreateXXX` and so on. That would be more consistent with the rest of the APIs

pault84:
SnapshotGroupCreate sounds weird to me. i'm not creating a snapshot group i'm group snapping.

lpabon:
> SnapshotGroupCreate sounds weird to me. i'm not creating a snapshot group i'm group snapping.
Yeah, I can see that. Hmm. I'm good with either way.

api/client/volume/client.go
```diff
@@ -658,3 +659,23 @@ func (v *volumeClient) CloudBackupSchedEnumerate() (*api.CloudBackupSchedEnumera
...
+
+func (v *volumeClient) SnapshotGroup(groupID string, labels map[string]string) (*api.GroupSnapCreateResponse, error) {
+
+	response := &api.GroupSnapCreateResponse{}
+	request := &api.GroupSnapCreateRequest{
+		Id:     groupID,
+		Labels: labels,
+	}
+
+	req := v.c.Post().Resource(snapPath + "/snapshotgroup").Body(request)
```
disrani-px: Define the paths as constant somewhere and use them in the client and server

api/server/volume.go
```diff
@@ -936,6 +936,48 @@ func (vd *volAPI) unquiesce(w http.ResponseWriter, r *http.Request) {
...
+		vd.sendError(vd.name, method, w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	d, err := vd.getVolDriver(r)
+	if err != nil {
+		notFound(w, r)
+		return
+	}
+
+	snapRes, _ = d.SnapshotGroup(snapReq.Id, snapReq.Labels)
```
disrani-px: Error from SnapshotGroup shouldn't be ignored.

api/api.proto
```diff
@@ -579,3 +579,26 @@ message ActiveRequests {
...
+  // Created snapshots
+  //
+  // in: body
+  // Required: true
+  repeated SnapCreateResponse snapshots = 1;
+  // Error message
+  //
+  // in: body
+  // Required: true
+  string error = 2;
```
disrani-px: It's better to return an error from the REST call instead of putting it in the response. With the current wat the caller has to check 2 things. 
1. Did the rest call return an error
2. Does the response have an error.

volume/volume.go
```diff
@@ -104,6 +104,8 @@ type SnapshotDriver interface {
 	Snapshot(volumeID string, readonly bool, locator *api.VolumeLocator) (string, error)
 	// Restore restores volume to specified snapshot.
 	Restore(volumeID string, snapshotID string) error
+	// GroupSnapshot takes a snapshot of specified volumegroup.
+	SnapshotGroup(groupID string, labels map[string]string) (*api.GroupSnapCreateResponse, error)
```
disrani-px: Comment says GroupSnapshot but API is SnapshotGroup and the response is GroupSnap. Make it consistent.
Also better to define GroupSnapCreateRequest to pass in parameters IMO.

PR #406, opened 2018-04-20
change array to map

pault84:

Change snap responses to a map instead of an array.


pault84:
doc changes are in.

Issue #407, opened 2018-04-23
SDK for OpenStorage


lpabon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE

**Description**:
The SDK for OpenStorage will provide developers with ability to programatically control OpenStorage based storage systems like Portworx. Developers will be provided with documentation, an API, clients for multiple languages, and an mock storage application to test and develop their code.

**Architecture**:
The SDK will run a gRPC server which will communicate with the OpenStorage interfaces, exporting only the objects and API functions to be consumed by external developers. See diagram below:

```
+------------+
|            |        +------------+
|   Current  |        |            |
|   REST     |        |  gRPC SDK  |
|            |        |            |
+------------+--------+------------+
|                                  |
|    OpenStorage Interfaces        |
|                                  |
+----------------------------------+
```

**Tasks**:
- [x] Share gRPC code with current CSI server to minimize duplication
- [x] Update `api.proto` with the RPC and Messages needed for the SDK
- [x] Update https://libopenstorage.github.io with documentation based on `api.proto`.
- [x] Create a server which serves gRPC for the SDK in `api/server/sdk`
- [x] Create an implementation for Cluster calls (including unit tests)
- [x] Create an implementation for Volume calls (including unit tests)
- [x] Update `Makefile` to create clients in Python, Ruby, Golang and place them in `api/client/sdk`
- [x] Create a fake in-memory driver implementation which can be used to create a simple application for developers to communicate with while developing and testing their code.

**Cloud APIs**:
- [x] Cloud Credentials APIs
- [x] Scheduling Policy

**Volume APIs**:
- [ ] Stat?
- [x] Cloud Backup
- [ ] Cloud Backup Scheduling
- [x] Object Storage

**Fake Driver**
- [x] Support Credentials
- [x] Support CloudBackup

Once done, we need a:
- [ ] Finalize documentation
- [ ] Full Review of the API

lpabon:
features tracked by project.

PR #408, opened 2018-04-23
sdk: Create new library for gRPC servers

lpabon:
This changes provides management of gRPC servers which will be
shared between CSI and SDK.

Part of #407

### DIFF
--- a/pkg/grpcserver/grpcserver.go
+++ b/pkg/grpcserver/grpcserver.go
@@ -1,0 +1,139 @@
+/*
+Package grpcserver is a generic gRPC server manager
+Copyright 2018 Portworx
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


pkg/grpcserver/grpcserver.go
```diff
@@ -0,0 +1,139 @@
+/*
+Package grpcserver is a generic gRPC server manager
+Copyright 2018 Portworx
```
pault84: Can we put this here?

pkg/grpcserver/grpcserver.go
```diff
@@ -0,0 +1,139 @@
...
+	running  bool
+	lock     sync.Mutex
+}
+
+// New creates a gRPC server on the specified port and transport.
+func New(config *GrpcServerConfig) (*GrpcServer, error) {
+	if nil == config {
+		return nil, fmt.Errorf("Configuration must be provided")
+	}
+	if len(config.Name) == 0 {
```
pault84: NIT: make this else if to compact the code a bit?

pkg/grpcserver/server.go
```diff
@@ -1,6 +1,6 @@
 /*
-Package csi is CSI driver interface for OSD
-Copyright 2017 Portworx
+Package grpcserver is a generic gRPC server manager
+Copyright 2018 Portworx
```
pault84: Same question as above can we have this in the osd codebase?

pkg/grpcserver/grpcserver.go
lpabon: yep

PR #409, opened 2018-04-23
Stat the attached EBS volume before returning the devicePath in Attach.

### LINKED ISSUE #305: Incorrect path saved onto the aws metadata.

adityadani:

AWS API returns a devicePath which was requested in the Attach call or
the one which it was able to attach the EBS volume on. For ex.

If the requested devicePath was /dev/sdb, then it will return in its
Attach API /dev/sdb, however based on the virtualization type of the
instance it will actually attach the volume at /dev/xvdb which can be
visible in lsblk output.

In the DevicePath API always os.Stat the device path and then return.

Closes #305


adityadani:
@lpabon I had modified the unit test for the DevicePath API for all storageops implementation to do a stat on the path returned by the API, so that we will catch such regressions.

PR #410, opened 2018-04-24
sdk: Cluster Inspect and Enumerate

lpabon:
Implementation for SDK Cluster Inspect and Enumerate

Part of #407 

This is the first of many gRPC calls for the Cluster API (Volume will be done after Cluster). Please read the proto file. I placed some questions there asking if it is needed to export some of the values.

Also, the unit tests use the generated Golang client 😀


api/api.proto
```diff
@@ -600,4 +600,82 @@ message GroupSnapCreateResponse {
...
+  // Used memory of the node
+  uint64 mem_used = 4;
+  // Free memory of the node
+  uint64 mem_free = 5;
+  // Average load (percentage)
+  int64 avg_load = 6;
+  // Node status
+  Status status = 7;
+  // Generated number of the node (XXX Do we need this exposed?)
+  uint64 gen_number = 8;
```
adityadani: this does not need to be exposed.

api/api.proto
```diff
@@ -600,4 +600,82 @@ message GroupSnapCreateResponse {
...
+  // List of disks on the node 
+  map<string, StorageResource> disks = 9;
+  // List of storage pools this node supports
+  repeated StoragePool pools = 10;
+  // Management IP
+  string mgmt_ip = 11;
+  // Data IP
+  string data_ip = 12;
+  // Timestamp (XXX What is this?)
+  google.protobuf.Timestamp timestamp = 13;
```
adityadani: we can skip this field as well.

api/api.proto
```diff
@@ -600,4 +600,82 @@ message GroupSnapCreateResponse {
...
+  // Start time of this node (XXX Time the node started or time that px started?)
+  google.protobuf.Timestamp start_time = 14;
```
adityadani: There is not really a need of start time. But uptime of a node does make sense. May be we can add that later as it will require some changes in ClusterManager.

api/api.proto
```diff
@@ -600,4 +600,82 @@ message GroupSnapCreateResponse {
...
+  // Status of the cluster
+  Status status = 1;
+  // Id of the cluster
+  string id = 2;
+  // NodeId is the id of the node servicing these requests
+  string node_id = 3;
+  // Nodes are a list of all the nodes on the cluster
+  repeated StorageNode nodes = 4;
+
+  // XXX Should we expose logging_url, management_url, fluentdconfig, and tunnelconfig?
```
adityadani: all of these are deprecated. They are not removed from the actual Cluster object for backward compatibility. We can discuss how we can get rid of them from the api.Cluster object.

PR #415, opened 2018-04-24
Volgrouprework

pault84:
Need the error in the response because ctl should not show http errors.


api/client/volume/client.go
```diff
@@ -670,12 +670,10 @@ func (v *volumeClient) SnapshotGroup(groupID string, labels map[string]string) (
 
 	req := v.c.Post().Resource(snapPath + "/snapshotgroup").Body(request)
 	res := req.Do()
-	if res.Error() != nil {
```
disrani-px: Why is this being changed? This doesn't make sense as to how REST APIs are supposed to work.

1. You are ignoring HTTP errors
2. You are returning HTTP OK for requests that actually fail. Which is the wrong behavior for REST calls. If a REST API fails the HTTP error code should be set.
3. In golang this API is inconsistent and awkward. All other APIs return an error code but this one doesn't.

sangleganesh: *the api inconsistency is all over the place, with the REST returning *Response objects and the API not returning it. It makes more sense to move towards returning a single Response object with errors. HTTP errors must be returned only if we have HTTP errors, API errors should not translate to HTTP errors.

PR #416, opened 2018-04-25
Add typedef for OpType for Cloudbackup

disrani-px:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 *.swn
 .starscope.db
 cscope.out
+cmd/osd-sanity/osd-sanity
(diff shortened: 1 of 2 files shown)


lpabon:
Why is this needed? Doesn't seem to provide any benefits.

Keep in mind that we will be moving some structs (if we can) from the api.go to api.proto as part of the SDK project.

disrani-px:
No benefit in openstorage because there is no implementation of CloudBackup here. The benefit is for the users of the API. There are fixed types of operations now that are defined for anyone checking the status.
These are used by the code in Portworx.

PR #417, opened 2018-04-25
Revert (#413)

pault84:
This reverts commit 8597d9ca590c1051f0f80fbed3b8ad1e1062c93a.
We need the external error because http could throw an error.
Also, we need the error in the response object so we can show a friendly error in the cli.


api/server/volume.go
```diff
@@ -974,12 +974,7 @@ func (vd *volAPI) snapGroup(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	snapRes, err = d.SnapshotGroup(snapReq.Id, snapReq.Labels)
-	if err != nil {
-		vd.sendError(vd.name, method, w, err.Error(), http.StatusBadRequest)
-		return
-	}
-
+	snapRes, _ = d.SnapshotGroup(snapReq.Id, snapReq.Labels)
```
sangleganesh: expecting drivers to not return error seems odd. @lpabon what is convention followed generally ?

lpabon: AFAIK, the interface should return an error and leave it up to the server (REST, gRPC, etc) to decide how to interpret and return the error accordingly.

So I am a little bit confused as to why we are removing #413. I think 413 is doing the right thing. Calling `sendError()` will appropriately setup the Body with the error message, and interpreted correctly by the REST client.

There is a difference between the http client throwing an error and the server returning an error. We should be careful to identify this in the client, and, afaik, it is done correctly.

@pault84 Could you provide a more complete example on to why this PR is needed?

pault84: That is supposed to be there in my branch it's there not sure why i don't see it in this commit.

pault84:
Figured out why that change was gone... vscode had my code cached somehow.

lpabon:
@pault84 Still don't get this, why return the error over a struct instead of using `err`?

pault84:
Because in pxctl we print the err straight from the function in this case. and that means printing a http-400 error. which looks bad. we just want to show the friendly error from the response object.

lpabon:
@pault84 I thought that was fixed by #372 ? Does the client pxctl use have that fix?

pault84:
```
snapRes, err = d.SnapshotGroup(snapReq.Id, snapReq.Labels)
	if err != nil {
		vd.sendError(vd.name, method, w, err.Error(), http.StatusBadRequest)
		return
	}

	json.NewEncoder(w).Encode(&snapRes)
```

only returns an http error i can't parse that in pxctl. pxctl doesn't know it's using a http call.

also check email discussion with ganesh

PR #419, opened 2018-04-25
sdk: gRPC client builds

lpabon:
This patch adds support to build Python, Ruby, and NodeJS gRPC
clients from `api/api.proto`. This is done by `make proto` from
the root openstorage.

Part of #407 

This will require developers to install the tools needed to build the client files for each of the languages. The Makefile checks for them and tells you if what to do and the README's also.

Also, no need to review the generated files for each of the languages. Concentrate mainly on the tree structure and Makefiles.


api/client/sdk/python/Makefile
```diff
@@ -0,0 +1,19 @@
...
+grpc:
+ifndef HAS_VIRTUALENV
+	$(error "Please install virtualenv for python: pip install virtualenv")
+endif
+	virtualenv $(VEDIR)
+	bash -c "source $(VEDIR)/bin/activate && \
+	pip install grpcio grpcio-tools && \
+	python -m grpc_tools.protoc \
+		-I$(APIDIR) \
+		--python_out=. \
```
ram-infrac: tree-structre : can we move autogenerated files into separate sub directory?

lpabon: Not needed in python to have an `include` directory like C.

PR #420, opened 2018-04-25
sdk: Cluster Alert Functions

lpabon:
Provides Alert functions for the SDK

Part of #407 

Changed SDK API functions only to ClusterAlertXXX to keep them all using the same style as the rest of the API calls.


api/api.proto
```diff
@@ -654,19 +644,76 @@ message StorageCluster {
   repeated StorageNode nodes = 4;
 }
 
-// Enumerate
+service OpenStorageCluster {
+	// Enumerate lists all the nodes in the cluster.
+	rpc Enumerate(ClusterEnumerateRequest)
```
sangleganesh: nit: indentation

api/server/sdk/cluster_test.go
```diff
@@ -291,3 +293,105 @@ func TestSdkInspectIdNotPassed(t *testing.T) {
 	assert.Equal(t, serverError.Code(), codes.InvalidArgument)
 	assert.Contains(t, serverError.Message(), "Node")
 }
+
+func TestSdkAlertEnumerate(t *testing.T) {
```
sangleganesh: what is this test testing ?

lpabon: It is testing that the grpc connection goes to the server, then the server calls the interface, and returns the expected value.

api/api.proto
lpabon: Weird, I'm not sure what is going on there with the tabs. I'll check it.

api/api.proto
```diff
@@ -654,19 +644,76 @@ message StorageCluster {
...
+}
+
+message ClusterAlertClearResponse {
+}
+
+message ClusterAlertEraseRequest {
+  // Type of resource (required)
+  ResourceType resource = 1;
+  // Id of alert as returned by ClusterEnumerateAlertResponse (required) 
+	int64 alert_id = 2;
```
adityadani: nit: indentation
There are multiple indentation issues in this file.

api/api.proto
```diff
@@ -654,19 +644,76 @@ message StorageCluster {
...
+	// Enumerate lists all the nodes in the cluster.
+	rpc Enumerate(ClusterEnumerateRequest)
+		returns (ClusterEnumerateResponse) {}
+
+	// Inspect the node given a UUID.
+	rpc Inspect(ClusterInspectRequest)
+    returns (ClusterInspectResponse) {}
+    
+  // Get a list of alerts from the storage cluster
+  rpc AlertEnumerate(ClusterAlertEnumerateRequest)
```
adityadani: Is this a RPC standard to name it as AlertEnumerate? IMO EnumerateAlert sounds more intuitive.

api/api.proto
```diff
@@ -654,19 +644,76 @@ message StorageCluster {
...
 }
 
-// Enumerate
+service OpenStorageCluster {
+	// Enumerate lists all the nodes in the cluster.
+	rpc Enumerate(ClusterEnumerateRequest)
+		returns (ClusterEnumerateResponse) {}
+
+	// Inspect the node given a UUID.
+	rpc Inspect(ClusterInspectRequest)
```
adityadani: nit: maybe rename it to InspectNode

api/api.proto
lpabon: Trying to standardize on reverse object naming style ---> direction. LIke Cluster->Alert->Enumerate.. Or AAA->BBB->CCCC for the API

api/api.proto
lpabon: Good one. I will make that change.

PR #421, opened 2018-04-26
sdk: SDK Developers Test Container

lpabon:
This adds support to create a small (35MB) container that can be used by developers to test their clients.

To build the container type: `make docker-build-sdk`

How to use:

* Start the docker container

```
$ docker run -d -p 9100:9100 openstorage/sdk
```

* Tested using Polyglot using the following command:

```
$ echo {} | java -jar polyglot.jar \
    --command=call \
    --endpoint=localhost:9100 \
    --full_method=openstorage.api.OpenStorageCluster/Enumerate
```

Part of #407 

The values in the Makefile to build `osd` static is so that we can use Alpine container which makes the total size 35MB! (Super fast downloads for developers!)


lpabon:
FYI, I'm not sold yet on the container name as `openstorage/sdk`. I welcome a better name now, or after it merges. I was thinking of something like `sdk-emulator`, or `sdk-mock`, but I can't decide.

Makefile
```diff
@@ -112,6 +112,16 @@ generate-mockfiles:
...
 
+docker-build-sdk:
+	rm -rf _tmp
+	mkdir -p _tmp
+	CGO_ENABLED=0 GOOS=linux go build \
+				-a -ldflags '-extldflags "-static"' \
+				-tags "$(TAGS)" \
+				-o ./_tmp/osd \
+				./cmd/osd
+	docker build -t openstorage/sdk -f Dockerfile.sdk .
```
sangleganesh: probably cleanup _tmp after docker build ?

cmd/osd/main.go
```diff
@@ -235,20 +236,34 @@ func start(c *cli.Context) error {
...
 		})
 		if err != nil {
 			return fmt.Errorf("Failed to start CSI server for driver %s: %v", d, err)
 		}
 		csiServer.Start()
+
+		// Start SDK Server for this driver
+		sdkServer, err := sdk.New(&sdk.ServerConfig{
+			Net:        "tcp",
+			Address:    ":9100",
```
sangleganesh: suggest making port an input param, with default to 9100 if no input provided.

PR #422, opened 2018-04-26
Adding schedule API into OSD(#387)

### LINKED ISSUE #387: Feature: Schedule Policy API
FEATURE

# Reason Needed
To create OSD-SDK
Move internal PX API to OSD to create external API SDK.
# Manage Schedule Policies

Schedule policy allows user to create snapshot schedule policy. Schedule policies can be addded to the volume either during volume create or after volume create.
 # a. List Schedule Policy
 List all scheduly policy created. If policy name is specified then get single policy details.
* **Method** : GET
* **Endpoint** : /schedpolicy
* **Content-type** : application/json
* **Response status** : 200
* **Params** :
    name (string) name of schedule policy

**Example Request** :
```GET /v1/schedpolicy```
```sh
Content-type : application/json
{
    name : "testpolicy"
}
```
# b. Create Schedule Policy
Create new schedule policy.
* **Method** : POST
* **Content-type** : /schedpolicy
* **Response status** : 200
* **Params** :
   name (string) : name of policy to create
   schedule (string) : policy intervals  mins, daily, monthly ,weekly etc

**Example Request** :
```POST /v1/schedpolicy```
```sh
Content-type : application/json
{
    name : "policy1"
    schedule : "12:00"
}
```
**Example Response** :
200

# c.  Update schedule policy
  Update schedule policy
* **Method** : PUT
* **Endpoint** : /schedpolicy
* **Content-type** : application/json
* **Response status** : 200
* **Params** :
   name (string) : name of policy to update
   schedule(string) : schedule for policy mins, daily, weekly etc.

**Example Request**
```PUT /v1/schedpolicy```
```sh
Content-type : application/json
{
    name : "demopolicy1"
    schedule : "12:00"
}
```
**Example Response** :
200

# d. Delete schedule policy
Delete schedule policy
* **Method** : DELETE
* **Endpoint** :  /schedpolicy
* **Content-type** : application/json
* **Response status** : 200
* **Params** :
    name(string): name of policy to delete

**Example Request** :
```DELETE /v1/schedpolicy```
```sh
Content-type : application/json
{
    name : "policy1"
}
```
**Example Response** :
200

ram-infrac:

This is PR for #387, which will add Schedule Policy API endpoints in OSD with default Not implemented status for schedule policy endpoints. Client can override this default manager implementation with their own.
Closes #387


schedpolicy/sched_policy.go
```diff
@@ -0,0 +1,52 @@
...
+func NewSchedulePolicyManager(sched SchedulePolicy) *Manager {
+	return &Manager{
+		SchedulePolicy: sched,
+	}
+}
+
+type nullSchedMgr struct {
+}
+
+func New() *nullSchedMgr {
```
lpabon: This one (and the Secret one) should probably have their name name from `New` to `NewNullSchedManager` because if not, when reading the server.go:120 looks like a real manager is being allocated.

schedpolicy/sched_policy.go
```diff
@@ -0,0 +1,52 @@
...
+	SchedPolicyDelete(name string) error
+	// SchedPolicyEnumerate enumerates all configured policies or the ones specified.
+	SchedPolicyEnumerate([]string) ([]*SchedPolicy, error)
+}
+
+type Manager struct {
+	SchedulePolicy
+}
+
+func NewSchedulePolicyManager(sched SchedulePolicy) *Manager {
```
lpabon: All of the following is a suggestion. What you have done works, but it is not normal golang so it is not as easy to read:

I'm not sure I understand the point of this function (and for Secret also). NullSchedManager is an implementation of `SchedulePolicy`.  server/cluster.go:36 can be `schedpolicy.SchedulePolicy` the interface instead of a pointer. Then the function below which returns a pointer to a schedule manager struct can be used since it satisfies the interface.

You can also add the following line to the top of the file to make sure that nullSchedMgr always satisfies the interface:

```go
var _ SchedPolicy = &nullSchedMgr{}
```

This will have the compiler immediately check if this struct satisfies the interface.

PR #423, opened 2018-04-27
sdk: Volume Operations

### LINKED ISSUE #401: Inspect may be needed to be changed to only take a single volume id
Possible BUG

**What happened**:
There could be a situation where too many volumes are passed to Inspect in api/client/volume/client.go. It creates a URI with all the ids and send it to the server. It would be more reliable and predictable to only ask for a single volume.

lpabon:
This patch adds support for Create, Delete, Inspect, and Enumerate Volumes

Part of #407 

* Do not review any of the generated client files (api.pb.go, or any files that ends in js, py, or rb)
* Changed `Inspect()` to only take a _single_ volume id. This fixes #401


api/api.proto
```diff
@@ -645,27 +645,87 @@ message StorageCluster {
...
     returns (ClusterAlertEraseResponse) {}
 }
 
+service OpenStorageVolume {
+  // Creates a volume
+  rpc Create(VolumeCreateRequest)
+    returns (VolumeCreateResponse) {}
+
+  // CreateSimpleVolume provides a simple API to create a volume
+  rpc CreateSimpleVolume(VolumeCreateSimpleVolumeRequest)
```
adityadani: IMO Providing two Create APIs would unnecessarily create confusion for the developers. Also exposing this CreateSimpleVolume API can be error prone as the developers would have to pass all the volume attributes as name value pairs and they can make mistakes. The existing Create API is well defined with an Object that can be passed over grpc/http and hence less error prone.

For docker/kubernetes it is OSD who is making the conversion from map to an object and hence it is under our control and less error prone.

api/server/sdk/volume.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/libopenstorage/openstorage/api/spec"
+	"github.com/libopenstorage/openstorage/cluster"
+	"github.com/libopenstorage/openstorage/volume"
+)
+
+// VolumeServer is an implementation of the gRPC OpenStorageVolume interface
+type VolumeServer struct {
+	specHandler spec.SpecHandler
+	driver      volume.VolumeDriver
```
adityadani: Current OSD http server handles multiple implementations of drivers simultaneously. It looks at the http request and chooses the correct driver. Are we planning to support that in grpc ?

api/server/sdk/volume_ops.go
```diff
@@ -0,0 +1,226 @@
...
+
+		// Return information on existing volume
+		return &api.VolumeCreateResponse{
+			Id: v.GetId(),
+		}, nil
+	}
+
+	// Check if the caller is asking to create a snapshot or for a new volume
+	var id string
+	if source != nil && len(source.GetParent()) != 0 {
```
adityadani: nit: source!=nil has already been checked above so you can skip it in this if

api/server/sdk/volume_ops.go
```diff
@@ -0,0 +1,226 @@
...
+	if err == nil {
+		// Check the requested arguments match that of the existing volume
+		if spec.Size != v.GetSpec().GetSize() {
+			return nil, status.Errorf(
+				codes.AlreadyExists,
+				"Existing volume has a size of %v which differs from requested size of %v",
+				v.GetSpec().GetSize(),
+				spec.Size)
+		}
+		if v.GetSpec().GetShared() != req.GetSpec().GetShared() {
```
adityadani: You might just want to add a new function that compares two specs and then returns an error if there is a discrepancy. It will compare all the fields and also handle empty values.

api/server/sdk/volume_ops.go
```diff
@@ -0,0 +1,226 @@
...
+) (*api.VolumeDeleteResponse, error) {
+
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply volume id")
+	}
+
+	// If the volume is not found, return OK to be idempotent
+	volumes, err := s.driver.Inspect([]string{req.GetVolumeId()})
+	if (err == nil && len(volumes) == 0) ||
+		(err != nil && err == kvdb.ErrNotFound) {
```
adityadani: why check for kvdb.ErrNotFound? I don't think all drivers return that error. volume.ErrEnoEnt would be the correct error

api/server/sdk/volume.go
lpabon: I have not seen a use case for this in openstorage, but if it turns out to be one, we will need can discuss it. Right now, just like CSI, the gRPC SDK will support only one driver.

api/api.proto
lpabon: Here is a more concrete proposal of the proto message:

sangleganesh: I think we should use the same API as we have currently (and use the same VolumeSpec and other proto definitions).

lpabon: I really would prefer for the users of this API not to use the three objects. It is extremely confusing and some of the values in api.VolumeSpec are not even used. I would like to provide a consistent clear API for them to consume. This API does not hinder the current one we use on REST at all, but when we expose these APIs to users which are new to OpenStorage, they would prefer a simpler, clearer API call. I know, because, everything I see the Create call, it takes me a while to understand it, and is only by looking into the source code as examples. I do not want to put this burden on the developers.

PR #426, opened 2018-05-03
sdk: Support for gRPC REST Gateway

lpabon:
This patch provides OpenStorage with support for the gRPC REST
Gateway based on https://github.com/grpc-ecosystem/grpc-gateway.

It also enables the REST gateway to provide Swagger-UI from the
'/swagger-ui' endpoint. This is done by taking the swagger-ui
content files and converting them to Golang code using packr from
https://github.com/gobuffalo/packr.

Developers will be able to go to `<endpoint>/swagger-ui` with their browser
and not only see the browser REST UI, but also execute it calls from it!

Lastly, the protobuf vendoring was updated due to missing support
from the files generated by the Gateway.

Part of #407

* No need to review generated files
* No need to review Swagger UI files added to `api/server/sdk/swagger-ui`.
* packr looks at the code looking for `packr.NewBox` like in `api/server/sdk/server.go`. Once it finds it, it then looks at the directory, which is mentioned in the API function, then it creates a file called `<package>-packr.go` with all the content. In this patch, it would create a file called `sdk-packr.go` in the same directory. It is highly discouraged to commit this file to git, and that is why it is created on every build (it is incredibly fast), and also why `.gitignore` has been updated to make sure it is not added.


api/api.proto
```diff
@@ -647,45 +648,95 @@ message StorageCluster {
 service OpenStorageCluster {
   // Enumerate lists all the nodes in the cluster.
   rpc Enumerate(ClusterEnumerateRequest)
-    returns (ClusterEnumerateResponse) {}
+    returns (ClusterEnumerateResponse) {
+      option(google.api.http) = {
```
sdeoras: i am curious, this option is part of proto spec as in protoc will take care of generating required code or requires specific plugins other than grpc plugin for go?

PR #437, opened 2018-05-16
PWX-5422: added raise singleton method

sdeoras:

added a new method to raise an alert as a singleton based on resource id and alert type

### DIFF
--- a/alert/alert.go
+++ b/alert/alert.go
@@ -52,6 +52,9 @@ type Alert interface {
 	// GetKvdbInstance.
 	GetKvdbInstance() kvdb.Kvdb
 
+	// RaiseSingleton raises a singleton alert.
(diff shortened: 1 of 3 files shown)


alert/alert.go
```diff
@@ -52,6 +52,9 @@ type Alert interface {
 	// GetKvdbInstance.
 	GetKvdbInstance() kvdb.Kvdb
 
+	// RaiseSingleton raises a singleton alert.
+	RaiseSingleton(alert *api.Alert) error
```
lpabon: This doesn't make sense to me. What does RaiseSingleton mean? To me it sounds like a single object is being.. raised? Not sure.

sdeoras:
new api is not required at the moment. closing for now.

PR #438, opened 2018-05-21
Add object store API's into OSD(#436)

### LINKED ISSUE #436: Feature: Object Store API's [WIP]
FEATURE

**Reason Needed**:
To create OSD-SDK
Move internal PX API to OSD to create external API SDK.

# Manage Object Store
Allow you to create and manage S3 compliant Objectstore.
 # a. Create Object Store
 Creates volume required to run object store.
* **Method** : POST
* **Endpoint** : /objectstore
* **Content-type** : application/json
* **Response status** : 200
* **Params** :
    volumeid (string): ID of volume

**Example Request** :
```POST /v1/objectstore```
```sh
Content-type : application/json
{
    volumeid : "1"
}
```
# b. Show object store
Show status of current object store
* **Method** : GET
* **Content-type** : /objectstore
* **Response status** : 200

**Example Request** :
```GET /objectstore```

**Example Response** :
200

# c.  Delete object store
   Use delete object store api to delete the volumes created for object store
* **Method** : DELETE
* **Endpoint** : /objectstore
* **Content-type** : application/json
* **Response status** : 200

**Example Response** :
200

# d. Enable/Disable object store
  Enable/Disable object store on server
* **Method** : PUT
* **Endpoint** :  /objectstore
* **Content-type** : application/json
* **Response status** : 200
* **Params** :
    enable(bool): Enable or disable object store flag

**Example Request** :
```PUT /v1/objectstore```
```sh
Content-type : application/json
{
    enable : true
}
```
**Example Response** :
200

ram-infrac:

This PR is for #436, which will add ObjectStore API's into OSD with default ```Not Imlemented``` status.

Closes #436 


api/client/cluster/objectstore.go
```diff
@@ -0,0 +1,51 @@
...
+	request := c.c.Get().Resource(clusterPath + ObjectStorePath)
+	if err := request.Do().Unmarshal(objectstoreInfo); err != nil {
+		return nil, err
+	}
+	return objectstoreInfo, nil
+}
+
+func (c *clusterClient) ObjectStoreCreate(volume string) error {
+	req := c.c.Post().Resource(clusterPath + ObjectStorePath)
+	req.QueryOption(objectstore.VolumeName, volume)
```
ram-infrac: @lpabon , I remember you advice on not to use "querystring", but our rest client causes /path variable to automatically convert into lower case characters. I don't know whether this behaviour is expected or not, but volume name can be case sensitive so keeping this as query parameter instead of including in path.

lpabon: great point. Thanks for the explanation. Please add this as a comment to the code.

objectstore/struct.go
```diff
@@ -0,0 +1,22 @@
+package objectstore
+
+const (
+	Enable     = "enable"
+	VolumeName = "name"
+)
+
+// Objectstore info returns current object store status
+// swagger:model
+type ObjectstoreInfo struct {
```
lpabon: Each of these values should have comments since they are to be consumed by the REST client.

Also, what do you think of making this a message in `api.proto`. Do you think this struct can be used for the SDK? Looks like it could be.

api/server/objectstore.go
```diff
@@ -0,0 +1,140 @@
...
+	method := "objectStoreCreate"
+	params := r.URL.Query()
+	volumeName := params[objectstore.VolumeName]
+
+	if len(volumeName) == 0 || volumeName[0] == "" {
+		c.sendError(c.name, method, w, "Missing secret ID", http.StatusBadRequest)
+		return
+	}
+
+	err := c.ObjectStoreManager.ObjectStoreCreate(volumeName[0])
```
lpabon: I would assume that when ObjectStorageCreate is called with a volume ID, that it returns the object storage information on how to access it. Why does it not return that information?

api/server/objectstore.go
```diff
@@ -0,0 +1,140 @@
...
+// - application/json
+// responses:
+//   '200':
+//     description: success
+//     schema:
+//      $ref: '#/definitions/ObjectstoreInfo'
+func (c *clusterApi) objectStoreInspect(w http.ResponseWriter, r *http.Request) {
+	method := "objectStoreInspect"
+
+	objInfo, err := c.ObjectStoreManager.ObjectStoreInspect()
```
lpabon: Doesn't Inspect need to take the volume ID just as Create() did?

api/server/objectstore.go
```diff
@@ -0,0 +1,140 @@
...
+		return
+	}
+
+	enable, err := strconv.ParseBool(strEnable[0])
+	if err != nil {
+		c.sendError(c.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	err = c.ObjectStoreManager.ObjectStoreUpdate(enable)
```
lpabon: Same here, volume id _should_ be used right?

api/server/objectstore.go
```diff
@@ -0,0 +1,140 @@
...
+// produces:
+// - application/json
+// parameters:
+// responses:
+//   '200':
+//     description: success
+func (c *clusterApi) objectStoreDelete(w http.ResponseWriter, r *http.Request) {
+	method := "objectStoreDelete"
+
+	err := c.ObjectStoreManager.ObjectStoreDelete()
```
lpabon: Same here, volume id _should_ be used right?

api/server/objectstore.go
ram-infrac: There is Inspect call for retrieving information about objectstore. Objectstorecreate creates object store, you have to call Objectstoreupdate(true) to enable it.

Current PX Create Object Store API does not send back info about objectstore in create call, if I voluntarily add it here I have to add Inspect() call inside Create() call while integration

api/server/objectstore.go
ram-infrac: No it doesn't. There is only one objectstore can be running on PX cluster with one volume as backend, I guess that's reason Inspect does not take volumeID. Same for Update and Delete.

@disrani-px  Please let me know If I'm missing something here

api/client/cluster/objectstore.go
ram-infrac: sure, thanks

objectstore/struct.go
ram-infrac: >>"Each of these values should have comments since they are to be consumed by the REST client."
Ok

are we going to export all of API i.e in OSD by SDK? If that's true we should move this to api.proto. In fact, I should start working on other API set i.e Schedule Policy, Secrets etc. What do you think?

api/server/objectstore.go
lpabon: It should return some type of information. We should do the right thing and return the requested data information to the caller. Please work with others to make sure we can change the interfaces also (yes, we should change the interfaces) to make sure that they return the appropriate information.

api/server/objectstore.go
lpabon: This is a limitation of the implementation of Px, but not of OpenStorage. The interface should not assume this limitation. There should be a setting in the VolumeSpec also showing that this volume has enabled ObjectStorage with a link to the objectstorage info object.

disrani-px: Yes, right now only one objectstore can be created in PX.

api/server/objectstore.go
```diff
@@ -0,0 +1,140 @@
...
+// responses:
+//   '200':
+//     description: success
+func (c *clusterApi) objectStoreCreate(w http.ResponseWriter, r *http.Request) {
+	method := "objectStoreCreate"
+	params := r.URL.Query()
+	volumeName := params[objectstore.VolumeName]
+
+	if len(volumeName) == 0 || volumeName[0] == "" {
+		c.sendError(c.name, method, w, "Missing secret ID", http.StatusBadRequest)
```
disrani-px: Wrong error message

api/server/objectstore.go
disrani-px: Creating an objectstore just created a binding between the objectstore and the volume to be used. The credentials are created when the objectstore is first enabled, so the only thing you can return here is the objectstore UUID, which isn't very helpful right now since there can only be one objectstore anyways.

objectstore/objectstore.go
```diff
@@ -0,0 +1,42 @@
...
+var (
+	ErrNotImplemented = errors.New("Not Implemented")
+)
+
+type ObjectStore interface {
+	// ObjectStoreInspect returns status of objectstore
+	ObjectStoreInspect() (*ObjectstoreInfo, error)
+	// ObjectStoreCreate objectstore on specified volume
+	ObjectStoreCreate(volume string) error
+	// ObjectStoreDelete objectstore on node
```
disrani-px: Deletes it from the cluster not from the node

api/server/objectstore.go
disrani-px: I disagree, the volume should not know what it is being used for.

disrani-px: The Create API can always return the same object as Inspect API.

disrani-px: There is no mapping from volumeID to objectstore. In the future, there might be a volume for each bucket, so that mapping doesn't make sense. You can pass in the objectstore ID to update/delete/inspect.

ram-infrac: sounds good

api/server/objectstore.go
```diff
@@ -0,0 +1,175 @@
...
+//   '200':
+//     description: success
+//     schema:
+//      $ref: '#/definitions/ObjectstoreInfo'
+func (c *clusterApi) objectStoreInspect(w http.ResponseWriter, r *http.Request) {
+	method := "objectStoreInspect"
+	params := r.URL.Query()
+	objstoreID := params[objectstore.ObjectStoreID]
+
+	if len(objstoreID) == 0 || objstoreID[0] == "" {
```
disrani-px: This should be optional. If there is only one objectstore configured there shouldn't be a need to pass in the objectstore ID. Better to leave this up to the implementer of ObjectStoreInspect()

objectstore/struct.go
```diff
@@ -0,0 +1,34 @@
...
+
+const (
+	Enable        = "enable"
+	VolumeName    = "name"
+	ObjectStoreID = "id"
+)
+
+// Objectstoreinfo returns current objectstore details
+// swagger:model
+type ObjectstoreInfo struct {
```
lpabon: Shouldn't this be in `api.proto`?

ram-infrac: missed this in offline meeting discussion. I have added this into api.proto

api/api.proto
```diff
@@ -419,6 +419,33 @@ message Alerts {
   repeated Alert alert = 1;
 }
 
+// ObjectstoreInfo is a structure that has current objectstore info
+// swagger:model
+message ObjectstoreInfo {
+    // UUID of objectstore                                                      
+    string UUID = 1;
```
lpabon: This looks great, but for consistency make all variables lower case. The output of the api.pb.go will make them upper case.

PR #439, opened 2018-05-21
added new simple alerts interface

sdeoras:

This PR defines a new interface for managing alerts.

This is intended to close the issues associated with non-rate-limited alerts flooding the backend storage such as etcd

This PR is only to settle on the interface. New PR's will be raised for implementation using kvdb

### DIFF
--- a/alert/interface.go
+++ b/alert/interface.go
@@ -1,6 +1,9 @@
 package alert
 
-import "time"
+import (
+	"io"
+	"time"
+)
 
 const (
 	NoAlertFound Error = "no alert found for the input key"
@@ -14,13 +17,16 @@ func (e Error) Error() string {
 	return string(e)
 }
 
-// Altertable defines an interface that can be managed using Alerter interface.
+// AlertType uniquely identifies type of an alert using a number
+type AlertType int64
+
+// Altertable defines an interface that can be managed using Manager interface.
 // An instance of Alertable is an alert object
 type Alertable interface {
-	// Type indicates alert type
-	Type() int64
-	// ID is the resource ID of alert generator
-	ID() string
+	// AlertType fetches alert type
+	AlertType() AlertType
+	// ResourceID fetches resource ID of alert generator
+	ResourceID() string
 	// Reason reports a brief comment on why alert was raised
 	Reason() string
 	// FirstSeen is the time stamp of first alert of this type and ID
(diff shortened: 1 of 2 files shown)


alert/interface.go
```diff
@@ -0,0 +1,48 @@
...
+type Error string
+
+// Error reports underlying error as a string
+func (e Error) Error() string {
+	return string(e)
+}
+
+// Altertable defines an interface that can be managed using Alerter interface.
+// An instance of Alertable is an alert object
+type Alertable interface {
```
adityadani: Isn't this interface already implemented in api.pb.go as the Alert struct is a proto definition?

alert/interface.go
```diff
@@ -0,0 +1,48 @@
...
+	// LastSeen is the time stamp of last alert of this type and ID
+	LastSeen() time.Time
+	// Marshal defines serialization of implementing object
+	Marshal() ([]byte, error)
+	// Unmarshal defines deserialization of implementing object
+	Unmarshal(b []byte) error
+}
+
+// Alerter defines an interface to manage alerts
+type Alerter interface {
```
adityadani: nit: I would vote for AlertManager as the name instead of Alerter. Still your call.

alert/interface.go
```diff
@@ -0,0 +1,48 @@
...
+	// Raise takes an input of Alertable type, serializes it and stores it in a backend service such as etcd
+	Raise(alertable Alertable) error
+	// Retrieve takes a key and returns an instance of Alertable object. It returns NoAlertFound if no entry is found
+	Retrieve(key string) (Alertable, error)
```
adityadani: Instead of a key can we take in definite constructs : (AlertType, ResourceID) that will uniquely identify the Alert and also easily query kvdb.

sdeoras:
@adityadani: thanks for taking a look. per recent discussions one of the paths was to redesign alerts interface. we are also evaluating other paths which extends previously defined interface. let's converge on which path we want to take.

alert/interface.go
sdeoras: considering this is going in libopenstorage, it would be easier to base it on interface instead of concrete structs. Other (non px) implementors may want to do have different definitions in their struct but as long as they comply with interface, we should be good... so was my thinking.

alert/interface.go
sdeoras: yeah, let's make this another manager :)

alert/interface.go
sdeoras: if altert type is not a px specific notion, yes, we can make it that way and provide a type in this pkg for others to use as well.

sdeoras:
not required. we decided to work with existing api. closing for now.

PR #442, opened 2018-05-23
 Add vsphere storage ops implementation

harsh-px:


pkg/storageops/vsphere/vsphere_util.go
```diff
@@ -0,0 +1,157 @@
...
+	if err != nil {
+		return false
+	}
+
+	_, err = storageops.GetEnvValueStrict("VSPHERE_DATASTORE")
+	return err == nil
+}
+
+// Get canonical volume path for volume Path.
+// Borowwed from https://github.com/kubernetes/kubernetes/blob/release-1.10/pkg/cloudprovider/providers/vsphere/vsphere_util.go#L312
```
adityadani: nit: borrowed

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+// Tags will list the existing labels/tags on the given volume
+func (ops *vsphereOps) Tags(volumeID string) (map[string]string, error) {
+	return nil, storageops.ErrNotSupported
+}
+
+func GetVMObject(ctx context.Context, conn *vclib.VSphereConnection, uuid string) (*vclib.VirtualMachine, error) {
+	// TODO change impl below using multiple goroutines and sync.WaitGroup to make it faster
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	err := conn.Connect(ctx)
```
adityadani: nit: if err := Connect(); err != nil

pkg/storageops/vsphere/vsphere_util.go
```diff
@@ -0,0 +1,157 @@
...
+	insecure, err := storageops.GetEnvValueStrict("VSPHERE_INSECURE")
+	if err == nil && strings.ToLower(insecure) == "true" {
+		cfg.InsecureFlag = true
+	}
+	return &cfg, nil
+}
+
+// IsDevMode checks if requirement env variables are set to run the pkg outside vsphere in dev mode
+func IsDevMode() bool {
+	_, err := storageops.GetEnvValueStrict("VSPHERE_VM_UUID")
```
adityadani: Why is VMSPHERE_VM_UUID not a part of the above ReadVSphereConfigFromEnv function?

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+	if err != nil {
+		return nil, err
+	}
+
```
adityadani: check for uuid is empty before getting all datacenters.

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+	return "vsphere"
+}
+
+func (ops *vsphereOps) Create(opts interface{}, labels map[string]string) (interface{}, error) {
+	volumeOptions, ok := opts.(*vclib.VolumeOptions)
+	if !ok {
+		return nil, fmt.Errorf("invalid volume options specified to create: %v", opts)
+	}
+
+	if volumeOptions.Tags == nil || len(volumeOptions.Tags) == 0 {
```
adityadani: nit: just len(volumeOptions.Tags) == 0 should suffice

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+		volumeOptions.Tags = labels
+	} else {
+		for k, v := range labels {
+			volumeOptions.Tags[k] = v
```
adityadani: create the Tags map if it is nil

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+
+	if len(datastores) == 0 {
+		return nil, fmt.Errorf("failed to find any available datastores to vm: %s for given prefix: %s", vmObj.Name(), cfg.Datastore)
+	}
+
+	logrus.Infof("Using following configuration for vsphere:")
+	logrus.Infof("  vCenter: %s:%s", cfg.VCenterIP, cfg.VCenterPort)
+	logrus.Infof("  Datacenter: %s", vmObj.Datacenter.Name())
+	logrus.Infof("  VMUUID: %s", cfg.VMUUID)
+	logrus.Infof("  Datastores: %v", datastores)
```
adityadani: Can the caller of Create API provide a datastore from this list or is it always going to be cfg.Datastore? If it is the latter case we should default to using cfg.Datastore instead of passing it in Create API.

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+
+	// Get the canonical path for the volume path.
+	canonicalVolumePath, err := getcanonicalVolumePath(ctx, ops.vm.Datacenter, diskPath)
+	if err != nil {
+		logrus.Errorf("Failed to get canonical vsphere disk path for: %s with "+
+			"volumeOptions: %+v on datastore: %s. err: %+v", diskPath, volumeOptions, datastore, err)
+		return nil, err
+	}
+
+	if filepath.Base(datastore) != datastore {
```
adityadani: nit: should we just move this check into the function getCanonicalVolumePath ?

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+
+	return nil, storageops.ErrNotSupported
+}
+
+// DeviceMappings returns map[local_attached_volume_path]->volume ID/NAME
+func (ops *vsphereOps) DeviceMappings() (map[string]string, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err := ops.renewVM(ctx, ops.vm)
```
adityadani: nit: if err := ops.renerVM(); err != nil {}

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	vmDevices, err := ops.vm.Device(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get devices for vm: %s", ops.vm.Name())
+	}
+
+	dsMatcher := regexp.MustCompile(fmt.Sprintf(".*\\[%s.+", ops.cfg.Datastore))
```
adityadani: nit: make this regex a part of ops itself so that we do not compile it every time.

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,397 @@
...
+func (ops *vsphereOps) DeviceMappings() (map[string]string, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	err := ops.renewVM(ctx, ops.vm)
+	if err != nil {
+		return nil, err
+	}
+
+	vmDevices, err := ops.vm.Device(ctx)
```
adityadani: Does this API differentiate between root partition devices and the ones which we created?

pkg/storageops/test/storageops.go
```diff
@@ -20,19 +19,28 @@ var diskLabels = map[string]string{
...
-			tags(t, d, diskName)
-			enumerate(t, d, diskName)
-			inspect(t, d, diskName)
-			attach(t, d, diskName)
-			devicePath(t, d, diskName)
-			teardown(t, d, diskName)
+			fmt.Printf("Created disk: %v\n", disk)
+			var diskID string
+			var ok bool
+			if driverName == "vsphere" {
```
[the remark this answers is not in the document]
harsh-px: done

pkg/storageops/vsphere/vsphere.go
harsh-px: done

pkg/storageops/vsphere/vsphere_util.go
harsh-px: added it back. It was needed

pkg/storageops/vsphere/vsphere.go
harsh-px: Caller always need to provide it in the create call. I've removed Datastore from cfg as it wasn't needed. The concept of datastores is only required by the consumers of this interface which is unit tests and portworx.

harsh-px:
@lpabon and @adityadani : addressed your comments. Please review once you get a chance.

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,376 @@
+package vsphere
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/kubernetes/kubernetes/pkg/cloudprovider/providers/vsphere/vclib/diskmanagers"
```
adityadani: nit: we could just stick to either github.com/kubernetes or k8s.io/kubernetes (line no. 14)

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,376 @@
...
+			logrus.Infof("vm: %s uuid: %s is in datacenter: %s running on host: %v", vm.Name(), vmUUID, dc.Name(), host.Reference())
+			return vm, nil
+		}
+	}
+
+	return nil, fmt.Errorf("failed to find vm with uuid: %s in any datacenter for vc: %s", vmUUID, conn.Hostname)
+}
+
+func (ops *vsphereOps) renewVM(ctx context.Context, vm *vclib.VirtualMachine) (*vclib.VirtualMachine, error) {
+	err := ops.conn.Connect(ctx)
```
adityadani: Connect implementation refreshes the Client object or creates a new one. If it does not return an error, then we can pass in ops.conn.GoVmomiClient to RenewVM.

If it returns an error then we can call the NewClient. Otherwise we are unnecessarily creating new clients for every renew call.

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,376 @@
...
+	// Go over all the devices attached on this vm and create a map of just the virtual disks and where
+	// they are attached on the vm
+	m := make(map[string]string)
+	for _, device := range vmDevices {
+		if vmDevices.TypeName(device) == "VirtualDisk" {
+			virtualDevice := device.GetVirtualDevice()
+			backing, ok := virtualDevice.Backing.(*types.VirtualDiskFlatVer2BackingInfo)
+			if ok {
+				devicePath, err := ops.DevicePath(backing.FileName)
+				if err == nil && len(devicePath) != 0 { // TODO can ignore errors?
```
adityadani: nit: we are calling renewVM twice, once in this function and once in DevicePath.

harsh-px: That's because `DevicePath` is also an interface function by itself and hence needs to ensure it has a renewed vm. With the updated renewVM implementation now, it will be less expensive.

pkg/storageops/vsphere/vsphere.go
harsh-px: done

pkg/storageops/vsphere/vsphere.go
harsh-px: good point. Fixed.

harsh-px:
@lpabon could you take a look when you get a chance?

pkg/storageops/utils.go
```diff
@@ -24,3 +28,12 @@ func AddElementToMap(
...
 	}
 }
+
+// GetEnvValueStrict fetches value for env variable "key". Returns error if not found or empty
+func GetEnvValueStrict(key string) (string, error) {
+	if val := os.Getenv(key); len(val) != 0 {
+		return val, nil
+	}
+
+	return "", fmt.Errorf("env variable %s is not set", key)
```
lpabon: Q: Is environment variables the only method of getting this information? Would a kv store or file not have them also?

pkg/storageops/vsphere/vsphere.go
```diff
@@ -0,0 +1,378 @@
...
+	if !ok {
+		return "", fmt.Errorf("invalid input: %v to GetDeviceID", diskPath)
+	}
+
+	return id, nil
+}
+
+// Attach takes in the path of the vmdk file and returns where it is attached inside the vm instance
+func (ops *vsphereOps) Attach(diskPath string) (string, error) {
+	ctx, cancel := context.WithCancel(context.Background())
```
lpabon: Q: Why run WithCancel() we are not cancelling or have a chain of ctx's. Maybe use WithTimeout() instead?

pkg/storageops/utils.go
harsh-px: Yeah, if we want, we could feed the providers with their configurations through various means (env, file, kv store). To start with, I had to pick one and env variables seemed most natural and easy to use. Also providers like GCE and AWS are written to accept known env variables (just like KUBECONFIG for k8s). Hence went ahead with env variables as the means for setting up configs.

pkg/storageops/vsphere/vsphere.go
harsh-px: We discussed this offline a bit. I'm going to leave this like this for now. The official examples for WithCancel() also use background (https://golang.org/pkg/context/#WithCancel)

PR #446, opened 2018-05-25
Add APIs for CloudMigrate

disrani-px:


api/api.go
```diff
@@ -417,6 +421,62 @@ type CloudBackupSchedEnumerateResponse struct {
...
+	Status          MigrateStatus
+	LastUpdate      time.Time
+	LastSuccess     time.Time
+}
+
+type CloudMigrateStartRequest struct {
+	Operation Operation
+	ClusterId string
+	// Depending on Operation type this can be VolumeID or VolumeGroupID
+	TargetId string
```
talakad: for migrating volume group, should we allow labels as well ( along with group iD)? Since group snapshot api allows both groupID and labels.

api/api.go
```diff
@@ -417,6 +421,62 @@ type CloudBackupSchedEnumerateResponse struct {
...
+}
+
```
talakad: might be a good idea to add migrateStartResponse structure that returns a unique ID? this unique ID can then be used to get status and also for cancelling the request?  With volume group migration( using volume labels), both status and cancel requests becomes easier with unique id.

disrani-px: The way it is right now, you would need to give it the same targetID for starting and canceling the operation. Migrating the group would take in labels, and you would pass in labels to both. We can update these to return task IDs once we have corresponding task IDs for cloudsnaps.

lpabon:
If these APIs are to be used by an external (non-px) caller, then we should add support also in the SDK.

disrani-px:
This isn't required in the SDK

lpabon:
Ok, I'll review this tonight

api/api.go
```diff
@@ -105,11 +105,15 @@ const (
...
-	OsdSnapshotPath = "osd-snapshot"
-	OsdCredsPath    = "osd-creds"
-	OsdBackupPath   = "osd-backup"
-	TimeLayout      = "Jan 2 15:04:05 UTC 2006"
+	OsdVolumePath        = "osd-volumes"
+	OsdSnapshotPath      = "osd-snapshot"
+	OsdCredsPath         = "osd-creds"
+	OsdBackupPath        = "osd-backup"
+	OsdMigratePath       = "osd-migrate"
+	OsdMigrateStartPath  = OsdMigratePath + "/start"
```
lpabon: Why have these defined here? They should be defined in the routes instead as the others are.

api/api.go
```diff
@@ -437,6 +441,62 @@ type CloudBackupSchedEnumerateResponse struct {
 	Schedules map[string]CloudBackupScheduleInfo
 }
 
+type Operation string
+
+const (
```
lpabon: Define all of these in api.proto to make it easier when we add this support to the SDK. They will need to be enums. Check [SDK Requirements](https://github.com/libopenstorage/openstorage/blob/master/docs/dev-sdk.md#sdk-requirements) for enums requirements

api/api.go
```diff
@@ -437,6 +441,62 @@ type CloudBackupSchedEnumerateResponse struct {
...
+	LocalVolumeName string
+	RemoteVolumeId  string
+	CloudSnapId     string
+	CurrentStage    MigrateStage
+	Status          MigrateStatus
+	LastUpdate      time.Time
+	LastSuccess     time.Time
+}
+
+type CloudMigrateStartRequest struct {
```
lpabon: Move structs as messages in the api.proto file.

api/api.go
```diff
@@ -437,6 +441,62 @@ type CloudBackupSchedEnumerateResponse struct {
...
+	Operation Operation
+	ClusterId string
+	// Depending on Operation type this can be VolumeID or VolumeGroupID
+	TargetId string
```
lpabon: Change this to be volume_target_id

api/api.go
```diff
@@ -437,6 +441,62 @@ type CloudBackupSchedEnumerateResponse struct {
...
+	CloudSnapId     string
+	CurrentStage    MigrateStage
+	Status          MigrateStatus
+	LastUpdate      time.Time
+	LastSuccess     time.Time
+}
+
+type CloudMigrateStartRequest struct {
+	Operation Operation
+	ClusterId string
```
lpabon: Is this the destination cluster id or the source?

api/api.go
```diff
@@ -437,6 +441,62 @@ type CloudBackupSchedEnumerateResponse struct {
...
+	TargetId string
+}
+
+type CloudMigrateCancelRequest struct {
+	Operation Operation
+	// Depending on Operation type this can be VolumeID or VolumeGroupID
+	TargetId string
+}
+
+type CloudMigrateStatusResponse struct {
```
lpabon: What is the key in this map? Please document.

api/client/volume/client.go
```diff
@@ -441,7 +439,7 @@ func (v *volumeClient) Unquiesce(volumeID string) error {
 // CredsEnumerate enumerates configured credentials in the cluster
 func (v *volumeClient) CredsEnumerate() (map[string]interface{}, error) {
 	creds := make(map[string]interface{}, 0)
-	err := v.c.Get().Resource(credsPath).Do().Unmarshal(&creds)
+	err := v.c.Get().Resource(api.OsdCredsPath).Do().Unmarshal(&creds)
```
lpabon: Why is this change needed? These are the types of things that would be good to document to the reviewer on the PR.

volume/drivers/fake/fake.go
```diff
@@ -51,6 +51,7 @@ type driver struct {
 	volume.QuiesceDriver
 	volume.CredsDriver
 	volume.CloudBackupDriver
+	volume.CloudMigrateDriver
```
lpabon: Implement an in-memory implementation of this in the fake driver.

api/api.go
disrani-px: Don't want to define them as enum, want to use strings like the cloudbackup API types. Can't define enums in the proto file so keeping them here.

api/api.go
disrani-px: They depend on the typed consts above, so can't move them there

api/api.go
disrani-px: This can be volume of group id. That's why it is called target id

api/client/volume/client.go
disrani-px: "OsdCredsPath" is already defined in api.go, no reason to define them locally here too.

api/api.go
disrani-px: The way it is defined right now for the other paths is incorrect. The server and client either redefine them or hard code the paths. The server and client need to refer to them from a common place, which is this file

volume/drivers/fake/fake.go
disrani-px: I don't see any value in adding that here. The implementation should test it.

disrani-px:
Not sure what more info you want, it's just the APIs that are defined in the PR.

I'll add the Unit tests, but they are kind of useless and don't really provide any value or catch any issues.

api/api.go
lpabon: Oh I see. Great point. In other words, there should be a common location for both to share. Please create an issue to fix it so that this style can be applied to all.

api/api.go
lpabon: I understand what you mean, but these values must be defined in api.proto and not here for when the SDK does call this.

api/api.go
lpabon: They need to move to have them used by the SDK in the future.

api/api.go
lpabon: As the author of the code this makes sense, but as a reader it is difficult to get that information from `TargetId`. As a reader, it would make it easier to have variables be self-documenting.

api/client/volume/client.go
lpabon: Cool

volume/drivers/fake/fake.go
lpabon: Need the implementation in fake driver so that when this is added to the SDK, the user of the SDK can use the mock-sdk-server container and test their code. If this is not added at the time of creation, it will must likely never get added by the original author.

lpabon:
> I'll add the Unit tests, but they are kind of useless and don't really provide any value or catch any issues.

This is absolutely incorrect. Unit tests provide an amazing method to make sure that the code you have added works now and most importantly in future changes by others. That is where unit tests provide the most quality assurance.

disrani-px:
I agree with unit tests, but the interface tests that are added for these REST APIs don't really test anything.

api/api.go
disrani-px: @jvinod mentioned we can still get the string of the enum from the protobuf defintion so moved it to the proto file

api/api.go
disrani-px: Migration is always to a remote cluster, we don't ever need to pass in the local cluster ID for anything

api/api.go
disrani-px: There is a comment to clarify it. volume_target_id doesn't make sense since it can also be a group id.

api/server/sdk/api/api.swagger.json
```diff
@@ -17,7 +17,7 @@
   "paths": {
     "/v1/cloudbackup": {
       "get": {
-        "summary": "Enumerate",
+        "summary": "Erases an alert for a given resource",
```
disrani-px: `make docs` seems to have messed this up.

disrani-px: `make proto` made these changes, I reverted them. But `make proto` still produces this wrong output.

api/api.proto
```diff
@@ -1585,3 +1585,81 @@ message SdkCloudBackupStateChangeRequest {
...
+    // Migrate all volumes in the cluster
+    MigrateCluster = 0;
+    // Migrate a single volume
+    MigrateVolume = 1;
+    // Migrate a group of volumes
+    MigrateVolumeGroup = 2;
+}
+
+enum CloudMigrateStage {
+    Backup = 0;
```
lpabon: All enums must be unique across the entire file (according to protoc error messages). Please prefix with the enum name.

api/api.proto
```diff
@@ -1585,3 +1585,81 @@ message SdkCloudBackupStateChangeRequest {
 
 message SdkCloudBackupStateChangeResponse {
 }
+
+enum CloudMigrateOperationType {
+    // Migrate all volumes in the cluster
+    MigrateCluster = 0;
```
lpabon: Highly suggest that 0 is left as an unknown value on all enums. That way you can check for errors if it was not defined.

volume/drivers/fake/fake.go
lpabon: This still needs to be done.

api/api.proto
disrani-px: Scoped them inside a message which is easier to read then prefixing with the enum name

api/api.proto
disrani-px: Added invalid states

PR #449, opened 2018-06-01
Credentials API osd-sdk

ipochi:
This PR adds the Credentials API functionality to osd-sdk
This PR includes work from me and @ram-infrac 
Part of #407 


api/api.proto
```diff
@@ -803,6 +803,206 @@ service OpenStorageVolume {
     }
 }
 
+service OpenStorageCredentials {
+
+  // Provide credentials to OpenStorage and if valid,
+  // it will return an identifier to the credentials
+  rpc ProvideForAWS(ProvideCredentialsForAWSRequest)
```
lpabon: Either use all "ProvideCredentialsXXX" or "CreateCredentialsXXX". Same for the REST urls. either `credentials/provide` or `credentials/create`.

My vote is not to use Create since we are not creating credentials. Either the verb could be Provide, or Submit, or Register, or something like that, but they all need to be consistent.

api/server/sdk/credentials.go
```diff
@@ -0,0 +1,404 @@
...
+		return nil, status.Error(codes.InvalidArgument, "Must supply Secret Key")
+	}
+
+	if len(req.GetRegion()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply Region Key")
+	}
+
+	if len(req.GetEndpoint()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply Endpoint Key")
+	}
```
lpabon: Good checks

api/server/sdk/credentials.go
```diff
@@ -0,0 +1,404 @@
...
+			"failed to create Azure credentials: %v",
+			err.Error())
+	}
+
+	// Validate if the credentials provided were correct or not
+	validateReq := &api.CredentialsValidateRequest{CredentialId: uuid}
+
+	validateErr := s.driver.CredsValidate(validateReq.GetCredentialId())
+
+	if validateErr != nil {
```
lpabon: Seems like this delete section is a repetition. Consolidate the processing into one function they can all use.

api/api.proto
disrani-px: It should be Create since it is creating the Credentials object and also keeps it inline with the other operations (volume, snapshot, etc)

api/api.proto
```diff
@@ -803,6 +803,206 @@ service OpenStorageVolume {
...
+      option(google.api.http) = {
+        post: "/v1/credentials/create/google"
+        body: "*"
+      };
+    }
+  // EnumerateForAWS lists the configured AWS credentials                      
+   rpc EnumerateForAWS(EnumerateCredentialsForAWSRequest)                       
+     returns (EnumerateCredentialsForAWSResponse) {                             
+     option(google.api.http) = {                                                
+       post: "/v1/credentials/enumerate/aws"
```
disrani-px: These endpoints look wrong, it should look like the current endpoints and you shouldn't need a different type of endpoint for each type of cloud.
For creating POST to /v1/credential (The input object should have the type and corresponding config)
For listing GET from /v1/credential
For getting a particular credential GET from /v1/credential/{id}
For deleting DELETE from /v1/credential/{id}

Also preferably it should be "credential" and not "credentials"

lpabon: @disrani-px Do not confuse these with the REST endpoint. This is the REST Gateway for the gRPC. And all are POSTs since they are all pushing objects.

api/api.proto
lpabon: That's fine. Then all should be Create.

api/api.proto
disrani-px: I didn't say they were the same, but the endpoints should be similar. Also POST doesn't mean you are pushing objects, you can always send objects with other types of requests too. POST should be used for create operations and the other request types should be used as mentioned above for other operations.

lpabon: I think this discussion is outside the scope of this PR. We can take this offline. @ipochi For now continue with all POSTs as the others, until further notice. @disrani-px and I will discuss it offline and propose a PR change if needed.

ipochi:
Review comments implemented.

api/api.proto
```diff
@@ -830,6 +830,193 @@ service OpenStorageVolume {
     }
 }
 
+service OpenStorageCredentials {
```
lpabon: Spacing needs to be fixed in the file. Looks like some are tabs and some are spaces. Looks like the original file is created with tabs as spaces and tabs set to 2

api/api.proto
```diff
@@ -830,6 +830,193 @@ service OpenStorageVolume {
...
+      post: "/v1/credentials/validate"
+      body: "*"
+    };
+  }
+}
+
+message CredentialCreateAzureRequest {
+  // Azure Credential
+  AzureCredential credential = 1;
+
```
lpabon: extra line here should be deleted

api/api.proto
```diff
@@ -830,6 +830,193 @@ service OpenStorageVolume {
...
+message CredentialCreateGoogleResponse {
+  // Id of the credentials
+  string credential_id = 1;
+}
+
+message CredentialCreateAWSRequest {
+  // AWS S3 Credential
+  S3Credential credential = 1;
+
```
lpabon: here

api/api.proto
```diff
@@ -830,6 +830,193 @@ service OpenStorageVolume {
...
+   // EnumerateForAzure lists the configured Azure credentials                  
+   rpc EnumerateForAzure(CredentialEnumerateAzureRequest)                   
+     returns (CredentialEnumerateAzureResponse) {                           
+     option(google.api.http) = {                                                
+       post: "/v1/credentials/enumerate/azure"                                  
+       body: "*"                                                                
+     };                                                                         
+   }                                                                            
+                                                                                
+
```
lpabon: Just one empty line in between

api/api.proto
```diff
@@ -830,6 +830,193 @@ service OpenStorageVolume {
...
+  // Create credential for Google and if valid ,
+  // returns a unique identifier
+  rpc CreateForGoogle(CredentialCreateGoogleRequest)
+    returns (CredentialCreateGoogleResponse) {
+      option(google.api.http) = {
+        post: "/v1/credentials/create/google"
+        body: "*"
+      };
+    }
+  // EnumerateForAWS lists the configured AWS credentials
```
lpabon: Add an empty line in between

api/server/sdk/credentials.go
```diff
@@ -0,0 +1,383 @@
...
+
+		if err != nil {
+			return status.Errorf(
+				codes.Internal,
+				"failed to delete invalid Google credentials: %v",
+				err.Error())
+		}
+
+		return status.Errorf(
+			codes.Internal,
```
lpabon: If the provided credentials did not validate, we shouldn't use `codes.Internal` as an error. I think we should use [PermissionDenied](https://github.com/grpc/grpc-go/blob/master/codes/codes.go#L72).

api/api.proto
ipochi: fixed

PR #450, opened 2018-06-01
Add helper API to wait for completion of cloudbackup operations

disrani-px:
Cloudbackup is being used in multiple plugins which can use
this API to wait for operations to complete

### DIFF
--- a/volume/cloudbackup.go
+++ b/volume/cloudbackup.go
@@ -1,0 +1,64 @@
+package volume
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/cenkalti/backoff"
+	"github.com/libopenstorage/openstorage/api"
+)
+
+var (
+	cloudsnapBackoffInitialInterval   = 5 * time.Second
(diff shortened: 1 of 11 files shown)


volume/cloudbackup.go
```diff
@@ -0,0 +1,64 @@
...
+
+		err = fmt.Errorf("CloudBackup operation %v for %v in state %v", opType, id, csStatus.Status)
+		switch csStatus.Status {
+		case api.CloudBackupStatusFailed, api.CloudBackupStatusAborted:
+			opError = err
+			return nil
+		case api.CloudBackupStatusDone:
+			opError = nil
+			return nil
+		case api.CloudBackupStatusNotStarted, api.CloudBackupStatusActive, api.CloudBackupStatusPaused, api.CloudBackupStatusStopped:
```
talakad: api. CloudBackupStatusStopped should move to the same case as api.CloudBackupStatusDone, as that state is meant for user initiated cloudbackup stop and cloudbackup is expected to be inactive at that point.

PR #451, opened 2018-06-04
sdk: Move CloudBackup structures to api.proto

lpabon:
The CloudBackup API is required by the SDK project described in #407 and, like any other call, requires the structures to be defined in the proto file. The issue is that most of the stuctures were defined in `api.go` . This change moves the definitions of the structures to `api.proto` laying down the foundation to the a future CloudBackup SDK API patch.

Part of #407

This required the following changes: 

1. Type `CloudBackupOpType` was redefined as an enum in the proto file, since protocol buffers cannot define constant strings.
1. Type `CloudBackupStatusType` was redefined as an enum.

I have also separated the generated code into its own commit, for easier reviewing.


lpabon:
Fixing travis build issues (which didn't show up on my tests, weird)..

lpabon:
@talakad and @pault84 that these changes would require the REST server to maintain compatibility with the older objects which used const strings instead of enums.

lpabon:
After some investigation, the new objects created from the proto file would have different JSON tags from those previously defined in `api.go`, therefore they would create incompatibilities in the REST interface.

Closing this PR for one where the SDK translates the objects as done in other calls.

Issue #447, opened 2018-05-26
Expose Objectstore API over gRPC SDK


ram-infrac:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE

**Description**
This will add objectstore API into openstorage SDK so that user can able to access Objectstore functionality with SDK. This should also provide documentation, client for multiple languages.

Functions needs to be exposed are -

1. ObjectStoreInspect(objectstoreID string) (ObjectstoreInfo, error)
           ObjectStoreInspect return status of provided objectstore in ObjectstoreInfo struct.
2. ObjectStoreCreate(volume string) (*api.ObjectstoreInfo, error)
       ObjectStoreCreate creates objectstore on given volume, return objectstoreInfo struct
	
3. ObjectStoreDelete(objectstoreID string) error
       ObjectStoreDelete deletes objectstore from cluster.

4. ObjectStoreUpdate(objectstoreID string, enable bool) error
        ObjectStoreUpdate updates objectstore status (enable/disable)

Where ObjectstoreInfo is -
``` 
type ObjectstoreInfo struct {
	// UUID of objectstore
	UUID string
	// VolumeID of volume used by object store
	VolumeID string
	// Enable/Disable created objectstore
	Enabled bool
	// Status of objectstore running/failed
	Status string
	// Action being taken on this objectstore
	Action int
	// AccessKey for login into objectstore
	AccessKey string
	// SecretKey for login into objectstore
	SecretKey string
	// Endpoints for accessing objectstore
	Endpoints []string
	// CurrentEndpoint on which objectstore server is accessible
	CurrentEndpoint string
	// AccessPort is objectstore server port
	AccessPort int
	// Region for this objectstore
	Region string
}
```

PR #456, opened 2018-06-07
Objectstore API OSD-SDK

ram-infrac:
This PR add Objectstore API over gRPC SDK, so developer/user can use objectstore functionality with SDK

Part of #407 
Closes #
#447 


api/server/sdk/objectstore.go
```diff
@@ -68,6 +68,14 @@ func (s *ClusterServer) UpdateObjectstore(
 	req *api.SdkObjectstoreUpdateRequest,
 ) (*api.SdkObjectstoreUpdateResponse, error) {
 
+	err := s.cluster.ObjectStoreUpdate(req.GetObjectstoreId(), req.GetEnable())
```
ram-infrac: Since ObjectstoreId is optional here avoiding check for it. 
Also relying on default value provided by protobuf
- for string it's  empty string 
- for bool it's false by default
@lpabon Let me know is it ok to rely on default value generated by api.proto keeping in mind of future sdk release

api/server/sdk/objectstore_test.go
```diff
@@ -0,0 +1,221 @@
...
+	// Setup client
+	c := api.NewOpenStorageObjectstoreClient(s.Conn())
+
+	// Get info
+	resp, err := c.InspectObjectstore(context.Background(), &api.SdkObjectstoreInspectRequest{ObjectstoreId: objResp.Uuid})
+	assert.Error(t, err)
+	assert.Nil(t, resp)
+
+	// Verify
+	assert.Contains(t, err.Error(), "some error")
```
ipochi: for failed scenarios verify internal codes like below
https://github.com/libopenstorage/openstorage/blob/055b88ca745e1bf046f0cb8d8ba96899f3bf66e8/api/server/sdk/credentials_test.go#L232

Add bad argument tests

ram-infrac: thanks Imran , Added mock test cases for bad argument with "InvalidArgument " error code check

ipochi:
Looks good to me.

PR #457, opened 2018-06-08
sdk: CloudBackup Support

lpabon:
Provides CloudBackup calls to the SDK.

Part of #407

* Do not worry too much about the docs. There will be a full review of the docs next week.
* Take care reviewing the translations from SDK types to Golang interface types.
* Also, the generated files have been moved to a separate commit for convenience.


api/api.proto
```diff
@@ -1373,3 +1457,185 @@ message SdkObjectstoreUpdateRequest {
...
+}
+
+message SdkCloudBackupDeleteResponse {
+}
+
+message SdkCloudBackupDeleteAllRequest {
+  // id of the volume for the request
+  string src_volume_id = 1;
+  // CredentialUUID is the credential for cloud to be used for the request
+  string credential_uuid = 3;
```
ram-infrac: is there any specific reason to jump for parameter number here?

api/server/sdk/cloud_backup.go
```diff
@@ -0,0 +1,252 @@
...
+	if err := s.driver.CloudBackupCreate(&api.CloudBackupCreateRequest{
+		VolumeID:       req.GetVolumeId(),
+		CredentialUUID: req.GetCredentialUuid(),
+		Full:           req.GetFull(),
+	}); err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to create backup: %v", err)
+	}
+
+	return &api.SdkCloudBackupCreateResponse{
+		// THIS WILL BE FIXED BY #### where Create will return an ID
```
ram-infrac: Just curious about this, here should current ```CloudBackupCreate``` api should be modified to return backup ID right? Alternatively we can enumerate and match volume ID with newly created cloudbackup and send that back again

api/server/sdk/cloud_backup.go
```diff
@@ -0,0 +1,252 @@
...
+
+// Restore a backup
+func (s *CloudBackupServer) Restore(
+	ctx context.Context,
+	req *api.SdkCloudBackupRestoreRequest,
+) (*api.SdkCloudBackupRestoreResponse, error) {
+
+	if len(req.GetBackupId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must provide backup id")
+	} else if len(req.GetCredentialUuid()) == 0 {
```
ipochi: nit: do we really need else statements 
couple of if's would do , idiomatic go.

api/server/sdk/cloud_backup.go
```diff
@@ -0,0 +1,252 @@
...
+
+// Delete deletes a backup
+func (s *CloudBackupServer) Delete(
+	ctx context.Context,
+	req *api.SdkCloudBackupDeleteRequest,
+) (*api.SdkCloudBackupDeleteResponse, error) {
+
+	if len(req.GetBackupId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must provide backup id")
+	} else if len(req.GetCredentialUuid()) == 0 {
```
ipochi: Correct me if i am wrong
pxctl doesnt place a mandatory requirement on providing cred uuid if there is only one credential present in the cluster.

Shouldn't it be the same for SDK ?

api/server/sdk/cloud_backup.go
```diff
@@ -0,0 +1,252 @@
...
+		Contents: r.Contents,
+	}, nil
+}
+
+// History returns ??
+func (s *CloudBackupServer) History(
+	ctx context.Context,
+	req *api.SdkCloudBackupHistoryRequest,
+) (*api.SdkCloudBackupHistoryResponse, error) {
+
```
ipochi: check if volume id is provided or not ?

api/server/sdk/cloud_backup.go
```diff
@@ -0,0 +1,252 @@
...
+	return r.ToSdkCloudBackupEnumerateResponse(), nil
+}
+
+// Status provides status on a backup
+func (s *CloudBackupServer) Status(
+	ctx context.Context,
+	req *api.SdkCloudBackupStatusRequest,
+) (*api.SdkCloudBackupStatusResponse, error) {
+
+	r, err := s.driver.CloudBackupStatus(&api.CloudBackupStatusRequest{
```
ipochi: check for request params local and volumeId

api/server/sdk/cloud_backup.go
ipochi: on second thoughts , I might disagree with my own statement.
but food for thought.

api/api.proto
lpabon: No, my mistake. I'll fix that.

api/server/sdk/cloud_backup.go
lpabon: yeah, good point.

api/server/sdk/cloud_backup.go
lpabon: Yeah, I understand what you mean, but it is ok.

api/server/sdk/cloud_backup.go
lpabon: yeah, I missed that. Thanks.

api/server/sdk/cloud_backup.go
lpabon: The object comments say that the srcvolumeid is optional. But those may be incorrect. I will check with @talakad .

talakad: srcvolumeid is optional. if not mentioned, it will return status of all instead of requested src volumeid.

PR #458, opened 2018-06-08
SDK Schedule Policy

ipochi:
This PR Adds SchedulePolicy api's to osd-sdk
Part of #407

Generated Files are in a separate commit.
SpecHandler code in type SchedulePolicyServer is commented out for clarification.

OSD Rest server has the below interface
https://github.com/libopenstorage/openstorage/blob/055b88ca745e1bf046f0cb8d8ba96899f3bf66e8/schedpolicy/sched_policy.go#L10

but the same Enumerate and Get calls for OSD-SDK  , it has been clubbed as one.


api/server/sdk/schedule_policy.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, status.Errorf(
+			codes.Internal,
+			"failed to create schedule policy: %v",
+			err.Error())
+	}
+
+	return &api.SdkSchedulePolicyCreateResponse{}, err
+}
+
+// Update method creates schedule policy
```
ram-infrac: nit: updates schedule policy.
Similar changes for rest of comments

api/api.proto
```diff
@@ -911,6 +911,86 @@ service OpenStorageCredentials {
...
+}
+
+message SdkSchedulePolicyDeleteResponse {
+}
+
+message SdkSchedulePolicy {
+  // Name of the schedule policy
+  string name = 1;
+  // Schedule string in yaml
+  string schedule = 2;
```
lpabon: This should probably not be a string. Looks like multiple values are provided encoded here. Not part of this PR, but something that needs to be addressed before release.

api/server/sdk/schedule_policy.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/libopenstorage/openstorage/api"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// SchedulePolicyServer is an implementation of the gRPC OpenStorageSchedulePolicy interface
+type SchedulePolicyServer struct {
+	api.OpenStorageSchedulePolicyServer
+	//specHandler spec.SpecHandler
```
lpabon: If not needed, please remove.

api/server/sdk/schedule_policy.go
```diff
@@ -0,0 +1,151 @@
...
+	cluster cluster.Cluster
+}
+
+// Create method creates schedule policy
+func (s *SchedulePolicyServer) Create(
+	ctx context.Context,
+	req *api.SdkSchedulePolicyCreateRequest,
+) (*api.SdkSchedulePolicyCreateResponse, error) {
+
+	if len(req.GetSchedulePolicy().GetName()) == 0 {
```
lpabon: Need check first if `req.GetSchedulePolicy() != nil`

lpabon: You will need to do this in all the functions that use this object.

api/server/sdk/schedule_policy.go
```diff
@@ -0,0 +1,151 @@
...
+			codes.Internal,
+			"failed to delete schedule policy: %v",
+			err.Error())
+	}
+
+	return &api.SdkSchedulePolicyDeleteResponse{}, err
+}
+
+// Enumerate method enumerates schedule policy
+func (s *SchedulePolicyServer) Enumerate(
```
lpabon: This API does not make sense to me. The Golang interface 

https://github.com/libopenstorage/openstorage/blob/055b88ca745e1bf046f0cb8d8ba96899f3bf66e8/schedpolicy/sched_policy.go#L17-L18

How can it get `the ones specified` if it takes no arguments?

I think that is incorrect. Enumerate should just return them all since there are no "flags" to match. No need to take `Name` as an argument.

What do you guys think @ipochi @ram-infrac ?

api/api.proto
ipochi: from what  i recall , this is a yaml , which is passed as a string.

lpabon: Created https://github.com/libopenstorage/openstorage/issues/459 to track

Issue #467, opened 2018-06-13
sdk: Change AlertErase to AlertDelete


lpabon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG

**What happened**:
AlertErase is actually deleting the alert, so the name should change.

Part of #407

Issue #466, opened 2018-06-13
sdk: Change CreateFromVolumeId to Clone


lpabon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
BUG

**What happened**:
CreateFromVolumeId can be easier explained if we change the name to Clone

Part of #407

PR #468, opened 2018-06-14
sdk: Fix url and methods in the REST Gateway

### LINKED ISSUE #452: SDK Rest endpoints look wrong
DISCUSSION

**What happened**:
As part of #449 @disrani-px brought up the following:

> These endpoints look wrong, it should look like the current endpoints and you shouldn't need a different type of endpoint for each type of cloud.
> For creating POST to /v1/credential (The input object should have the type and corresponding config)
> For listing GET from /v1/credential
> For getting a particular credential GET from /v1/credential/{id}
> For deleting DELETE from /v1/credential/{id}

Placing it here for an open discussion

lpabon:
The REST Gateway generator could be better utilized by being more specific on which methods and urls to use.

Closes #452

The generated files are on their own commit.


lpabon:
While working on this with @disrani-px we found two issues: #467, and #466

api/api.proto
```diff
@@ -907,35 +900,33 @@ service OpenStorageCredentials {
   rpc EnumerateForAWS(SdkCredentialEnumerateAWSRequest)
     returns (SdkCredentialEnumerateAWSResponse) {
       option(google.api.http) = {
-        post: "/v1/credentials/enumerate/aws"
-        body: "*"
+        get: "/v1/credentials/aws"
```
disrani-px: Same here, probably need one API to enumerate all credentials instead of having separate ones.

lpabon: That is a great point. Created #469 to fix this.

PR #485, opened 2018-06-21
Sdk enumerate

### LINKED ISSUE #475: sdk: All Enumerates should return the IDs only

lpabon:
Allow clients to have more flexibility by only returning ids in the Enumerates.

Closes #475 
Part of #407

Do not review the generated files commit.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -698,9 +698,9 @@ func (c *Cluster) ToStorageCluster() *StorageCluster {
 		NodeId: c.NodeId,
 	}
 
-	cluster.Nodes = make([]*StorageNode, len(c.Nodes))
+	cluster.NodeIds = make([]string, len(c.Nodes))
 	for i, v := range c.Nodes {
-		cluster.Nodes[i] = v.ToStorageNode()
+		cluster.NodeIds[i] = v.Id
 	}
 
 	return cluster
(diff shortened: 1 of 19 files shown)


api/api.go
```diff
@@ -698,9 +698,9 @@ func (c *Cluster) ToStorageCluster() *StorageCluster {
 		NodeId: c.NodeId,
 	}
 
-	cluster.Nodes = make([]*StorageNode, len(c.Nodes))
+	cluster.NodeIds = make([]string, len(c.Nodes))
```
pault84: This is used only by the sdk? if so i'm ok with this change if not we can't do this.

api/api.pb.go
```diff
@@ -1062,53 +1062,53 @@ func (m *Group) GetId() string {
 // swagger:model
 type VolumeSpec struct {
 	// Ephemeral storage
-	Ephemeral bool `protobuf:"varint,1,opt,name=ephemeral,proto3" json:"ephemeral,omitempty"`
```
pault84: Why are all the proto3 references deleted? did we change versions? does it not matter?

lpabon: Great question. It all depends who generated them before and after. Right before we 'release', we should agree which protoc version to use and generate using that.

api/api.go
lpabon: Yeah, the StorageCluster is an SDK-only structure.

PR #486, opened 2018-06-25
sdk: Rework credentials apis

### LINKED ISSUE #469: sdk: Instead of having Credentials.CreateForXXX functions, use enums
BUG

**What happened**:
As pointed to @disrani-px we need to change the credentials CreateForXXX and EnumerateForXXX and have them take an enum of the cloud provider as one of the values in the request.

lpabon:
Instead of having an API per credential type, we now use the OneOf tags in proto files, making it much easier to use and implement.

Closes #469 
Part of #407 

* Added Inspect call.
* Generated files moved to their own commit.


api/server/sdk/credentials_test.go
```diff
@@ -598,90 +620,11 @@ func TestSdkCredentialEnumerateAzureSuccess(t *testing.T) {
 		// Setup client
 	c := api.NewOpenStorageCredentialsClient(s.Conn())
 
-	// Enumerate Azure Credentials
-	resp, err := c.EnumerateForAzure(context.Background(), req)
+	// Inspect
+	resp, err := c.Inspect(context.Background(), req)
```
ram-infrac: can you please add example of how to use Inspect resp here?

lpabon: Yeah, great idea.

api/server/sdk/credentials_test.go
```diff
@@ -241,19 +249,19 @@ func TestSdkAzureCredentialCreateBadArgument(t *testing.T) {
...
 	defer s.Stop()
 
-	req := &api.SdkCredentialCreateAzureRequest{}
-
-	params := make(map[string]string)
-
-	params[api.OptCredType] = "azure"
-	params[api.OptCredAzureAccountKey] = req.GetCredential().GetAccountKey()
-	params[api.OptCredAzureAccountName] = req.GetCredential().GetAccountName()
+	req := &api.SdkCredentialCreateRequest{
```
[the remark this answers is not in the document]
lpabon: Doesn't look like an issue to me.

api/server/sdk/credentials_test.go
lpabon: Actually, there is an example on line 627, but I'll add more info to the proto file docs.

PR #489, opened 2018-06-27
Add in-memory schedule policy to cluster

### LINKED ISSUE #487: Cluster.Schedule functions are pointing to nil and panic
BUG

**What happened**:
The cluster manager object has no implementations of the schedule functions since they are added to the rest point instead of the cluster object.

See: https://github.com/libopenstorage/openstorage/blob/master/api/server/server.go#L134

When another Golang function calls the cluster.SchedXXX calls, they panic, because even though they compile, there is no implementation, and therefore a nil pointer is tried to be called.

lpabon:
The issue is that even though it compiled, the current cluster.Cluster
instance in osd did not have an implementation of the schedulerpolicy.
This would cause panics when the SDK would try to use the functions.
Instead it was in the REST server.

Closes #487

* Move the configuration struct to cluster/ from api/server
* Add an in-memory implementation of the SchedulePolicy implementation
* Change the interface name from SchedulePolicy to
SchedulePolicyProvider so that it does not conflict with the struct in
schedpolicy/ of the same name.

### DIFF
--- a/cluster/manager.go
+++ b/cluster/manager.go
@@ -50,7 +50,6 @@ var (
 // ClusterManager implements the cluster interface
 type ClusterManager struct {
 	secrets.Secrets
-	sched.SchedulePolicy
 	objectstore.ObjectStore
 
 	size          int
@@ -69,6 +68,7 @@ type ClusterManager struct {
 	selfNodeLock  sync.Mutex // Lock that guards data and label of selfNode
 	system        systemutils.System
 	configManager osdconfig.ConfigManager
+	schedManager  sched.SchedulePolicyProvider
 }
 
 type checkFunc func(ClusterInfo) error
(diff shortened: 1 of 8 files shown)


schedpolicy/sched_policy.go
```diff
@@ -19,30 +19,3 @@ type SchedulePolicy interface {
 	// SchedPolicyGet returns schedule policy matching given name.
 	SchedPolicyGet(name string) (*SchedPolicy, error)
 }
-
-func NewDefaultSchedulePolicy() SchedulePolicy {
```
ram-infrac: If we removed default ```Not Implemented```  implementation, caller will be working with in-memory fake implementation of schedule policy. Is that expected behaviour?

cluster/manager.go
```diff
@@ -1080,6 +1093,9 @@ func (c *ClusterManager) Start(
 	if err != nil {
 		return err
 	}
+	if config.ConfigSchedManager == nil {
```
ram-infrac: can we use ```CheckNullClusterServerConfiguration()``` here instead of checking each of Manager objects?

lpabon: Yeah, I think that is a good idea. I will make the change.

schedpolicy/sched_policy.go
lpabon: before, the caller was getting a panic, because there was no implementation. The one you see above which was removed was the one in the REST server, but no in the ClusterManager.

schedpolicy/fake.go
```diff
@@ -0,0 +1,94 @@
...
+			Schedule: string(kv.Value),
+		}
+	}
+
+	return list, nil
+}
+
+func (s *fakeSchedMgr) SchedPolicyGet(name string) (*SchedPolicy, error) {
+	var schedule string
+	_, err := s.kv.GetVal(fakeSchedPrefix, &schedule)
```
ram-infrac: I may be wrong, but we query schedule against name right? did name is missed here

lpabon: Yes, That is a bug. Great find :+1:

PR #494, opened 2018-06-27
Remove silent handler from cluster object

lpabon:
PR #489 fixed the issue where any Golang function request to the
cluster object would be handled by an in-memory implementation instead
of having the system panic due to having no implementation available.

This could introduce another issue in that it could have a 'silent'
handler which could be unexpected. Instead, this fix makes it so
that the cluster object has a handler which always returns error
on default, removing the silent handler.

### DIFF
--- a/schedpolicy/sched_policy.go
+++ b/schedpolicy/sched_policy.go
@@ -19,3 +19,30 @@ type SchedulePolicyProvider interface {
 	// SchedPolicyGet returns schedule policy matching given name.
 	SchedPolicyGet(name string) (*SchedPolicy, error)
 }
+
+func NewDefaultSchedulePolicy() SchedulePolicyProvider {
+	return &nullSchedMgr{}
+}
+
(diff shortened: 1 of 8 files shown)


schedpolicy/sched_policy.go
```diff
@@ -19,3 +19,30 @@ type SchedulePolicyProvider interface {
...
+
+func NewDefaultSchedulePolicy() SchedulePolicyProvider {
+	return &nullSchedMgr{}
+}
+
+type nullSchedMgr struct {
+}
+
+func (sp *nullSchedMgr) SchedPolicyCreate(name, sched string) error {
+	return ErrNotImplemented
```
sdeoras: i prefer these returned errors to be constants and not vars

sdeoras: also error strings should not start with capital letters

lpabon:
thanks @sdeoras .

PR #497, opened 2018-06-28
sdk: SdkSchedulePolicy Schedule structure

### LINKED ISSUE #484: sdk: SdkSchedulePolicy.schedule should be a structured value not a string.

lpabon:
SdkSchedulePolicy Schedule was an unstructured string making it difficult for developers to understand how to use it. This patch provides a structure to the schedule making it easier to use.
 
Closes #484

Provides structure to SdkSchedulePolicy by creating three objects (Daily, Weekly, Monthly) and placing them in the proto file message using `oneof`.


api/server/sdk/utils.go
```diff
@@ -0,0 +1,210 @@
...
+
+func sdkSchedToRetainInternalSpecYamlByte(req *api.SdkSchedulePolicyInterval) ([]byte, error) {
+	sched, err := sdkSchedToRetainInternalSpec(req)
+	if err != nil {
+		return nil, err
+	}
+
+	out, err := yaml.Marshal(sched)
+	if err != nil {
+		return nil, status.Error(codes.Internal, "Failed to create schedule")
```
ram-infrac: can you please also append err so that user will know it's parsing error

ram-infrac: Not blocking , but good to have this

PR #513, opened 2018-07-03
sdk: Volume Set API

### LINKED ISSUE #482: sdk: Change OpenStorageCredentials and OpenStorageSchedulePolicy REST points
```diff
@@ -934,8 +938,7 @@ service OpenStorageCredentials {
   rpc Validate(SdkCredentialValidateRequest)
     returns (SdkCredentialValidateResponse) {
       option(google.api.http) = {
-        post: "/v1/credentials/validate"
-        body: "*"
+        get: "/v1/credentials/validate/{credential_id}"
       };
     }
 }
@@ -955,7 +958,7 @@ service OpenStorageSchedulePolicy {
   rpc Update(SdkSchedulePolicyUpdateRequest)
     returns (SdkSchedulePolicyUpdateResponse) {
       option(google.api.http) = {
-        post: "/v1/schedulepolicy/update"
+        put: "/v1/schedulepolicy"
         body: "*"
       };
     }
```

### LINKED ISSUE #499: sdk: Finalize documentation in the proto file.

### LINKED ISSUE #505: sdk: Volume Set() API
The Set() API is used for many operations. We may have to break this into multiple API calls or use OneOf{}.

lpabon:
Adds support for Volume Set(). Also:

* Lots of documentation updates
* CloudBackupEnumerate now returns objects instead of just IDs
there is no Inspect call in CloudBackup interface for the caller
to get information

Closes #482
Closes #505
Closes #499
Closes #482

* Created new `VolumeSpecSet` proto message which uses `oneof` to be clear if a value was set or not. If a value is not set, the value of the `oneof` variable will be `nil`. If it is, then the value will not be nil, and it will contain the appropriate type.
* Generated files are in their own commit


api/api.proto
```diff
@@ -299,6 +309,58 @@ message VolumeSpec {
...
+  oneof passphrase_opt { string passphrase = 15; }
+  // SnapshotSchedule a well known string that specifies when snapshots should be taken.
+  oneof snapshot_schedule_opt { string snapshot_schedule = 16; }
+  // Scale allows autocreation of volumes.
+  oneof scale_opt { uint32 scale = 17; }
+  // Sticky volumes cannot be deleted until the flag is removed.
+  oneof sticky_opt { bool sticky = 18; }
+  // Group identifies a consistency group
+  Group group = 21;
+  // GroupEnforced is true if consistency group creation is enforced.
```
lpabon: This should be a oneof if it can be set

api/api.proto
```diff
@@ -847,15 +948,17 @@ service OpenStorageObjectstore {
...
   rpc Delete(SdkObjectstoreDeleteRequest)
     returns(SdkObjectstoreDeleteResponse){
       option(google.api.http) = {
         delete: "/v1/objectstore/{objectstore_id}"
       };
     }
 
-  // Updates provided objectstore status
+  // Updates provided objectstore status.
+  // This call can be used to stop the server while maintaining the same
```
ram-infrac: nit: start/stop server

api/server/sdk/volume_ops.go
```diff
@@ -248,3 +248,96 @@ func (s *VolumeServer) Enumerate(
...
+	if req.GetIoProfileOpt() != nil {
+		spec.IoProfile = req.GetIoProfile()
+	} else {
+		spec.IoProfile = vol.GetIoProfile()
+	}
+
+	// FSType
+	if req.GetFormatOpt() != nil {
+		spec.Format = req.GetFormat()
+	}
```
lpabon: If not a pointer make sure to always set to current value.

PR #515, opened 2018-07-03
APIs to create cloudbackup and cloudbackup schedule for a volumegroup.

talakad:
CloudBackupGroupCreate API creates group snapshot of a volume group( specified
either through a group id or set of labels), and uploads these snapshots
to a cloud.

CloudBackupGroupSchedCreate API creates schedule for creating and uploading
group snapshot of a volume group.

Closes #


api/client/volume/client.go
```diff
@@ -632,6 +644,24 @@ func (v *volumeClient) CloudBackupSchedCreate(
 	return createResponse, nil
 }
 
+// CloudBackupGroupSchedCreate for a volume group creates a schedule to backup
+// volume group to the cloud
+func (v *volumeClient) CloudBackupGroupSchedCreate(
+	input *api.CloudBackupGroupSchedCreateRequest,
+) (*api.CloudBackupSchedCreateResponse, error) {
+	createResponse := &api.CloudBackupSchedCreateResponse{}
+	req := v.c.Post().Resource(api.OsdBackupPath + "/schedgroup").Body(input)
```
disrani-px: Define the paths as a constant in api.go and use them here and api/server/volume.go

api/api.go
```diff
@@ -420,12 +431,32 @@ type CloudBackupScheduleInfo struct {
...
+	GroupID string
+	// Labels indicates a volume group for this cloudsnap schedule
+	Labels map[string]string
 }
 
 type CloudBackupSchedCreateRequest struct {
 	CloudBackupScheduleInfo
 }
 
+type CloudBackupGroupSchedCreateRequest struct {
```
disrani-px: The schedules will be deleted with the same API right? Maybe use CloudBackupSchedCreateRequest (and path) for creating both too?

api/api.go
```diff
@@ -420,12 +431,32 @@ type CloudBackupScheduleInfo struct {
...
+	// GroupID indicates the group of volumes for which cloudbackup schedule is
+	// being created
+	GroupID string
+	// Labels indicates a volume group for which this group cloudsnap schedule is
+	// being created. If this is provided GroupId is not needed and vice-versa.
+	Labels map[string]string
```
disrani-px: We should document here what happens when both are provided.

api/api.go
```diff
@@ -420,12 +431,32 @@ type CloudBackupScheduleInfo struct {
...
+	// CredentialUUID is cloud credential to be used with this schedule
+	CredentialUUID string
+	// Schedule is the frequence of backup
```
disrani-px: typo: frequency

api/api.go
```diff
@@ -247,6 +247,17 @@ type CloudBackupCreateRequest struct {
 	Full bool
 }
 
+type CloudBackupGroupCreateRequest struct {
+	// GroupID indicates backup request for a volumegroup with this group id
```
lpabon: This is either GroupID or Labels. What happens if they are both set?

api/api.go
```diff
@@ -247,6 +247,17 @@ type CloudBackupCreateRequest struct {
 	Full bool
 }
 
+type CloudBackupGroupCreateRequest struct {
```
lpabon: Move all new structs to api.proto

volume/volume.go
```diff
@@ -136,6 +136,8 @@ type QuiesceDriver interface {
 type CloudBackupDriver interface {
 	// CloudBackupCreate uploads snapshot of a volume to the cloud
 	CloudBackupCreate(input *api.CloudBackupCreateRequest) error
+	// CloudBackupGroupCreate creates and then uploads volumegroup snapshots
+	CloudBackupGroupCreate(input *api.CloudBackupGroupCreateRequest) error
```
lpabon: How is status determined? When the user calls CloudBackupCreate( src_volume_id ) they can then call CloudBackupStatus( src_volume_id )

Is there something similar for this new API?

api/api.go
lpabon: When this is moved to api.proto use `oneof` instead, and it will be clear.

api/api.go
```diff
@@ -247,6 +247,17 @@ type CloudBackupCreateRequest struct {
...
+	// GroupID indicates backup request for a volumegroup with this group id
+	GroupID string
+	// Labels indicates backup request for a volume group with these labels
+	Labels map[string]string
```
lpabon: Use `oneof` to let the caller know only GroupID or Labels are used.

api/api.go
talakad: Create a new ticket to track this effort separately #516

api/api.go
talakad: Thanks. Will fix that

api/api.go
talakad: Yes. I will add. A volume list that matches both criteria will be used.

api/api.go
talakad: Yes. Same API is used to delete the schedule as both are recognized by UUID. thought of re-suaing the same API, but input parameters could get confusing. Hence the separate API.

api/client/volume/client.go
talakad: rest of the operations in the same class don' have them defined. so leaving it as it is.

disrani-px: Actually the new APIs I added do this. We'll never fix this if we keep adding string everywhere

PR #529, opened 2018-07-11
sdk: Fix Inspect Expectation

### LINKED ISSUE #528: fake driver returns nil when calling Inspect
BUG

**What happened**:
Calling driver.Inspect() with a value known not to exist returns an `err` value of `nil`. This causes callers to panic if they look at `err.Error()`.

Return a non-nil err value for a key that is not found

lpabon:
Closes #528

Incorrect expectation of error on Inspect

Closes #528


volume/drivers/fake/fake.go
```diff
@@ -126,6 +126,17 @@ func (d *driver) Status() [][2]string {
 	return [][2]string{}
 }
 
+func (d *driver) Inspect(volumeIDs []string) ([]*api.Volume, error) {
+	volumes, err := d.StoreEnumerator.Inspect(volumeIDs)
+	if err != nil {
+		return nil, nil
```
lpabon: Bug, should return nil, err

PR #530, opened 2018-07-12
 sdk: Create new Node Service and Clean up docs

### LINKED ISSUE #527: sdk: Cluster calls need to be adjusted
* cluster.Enumerate() -> Returns id of clusters paired to the current one. Currently just one
* cluster.Inspect(id) -> Returns cluster info and its node ids cluster.Enumerate().
    * Id may be a parameter that we add when pairing is added
* node.Inspect(id) -> Return node information, which is what cluster.Inspect() today.
* node.Enumerate() -> Return all the ids of the nodes attached to the current cluster.

lpabon:

* Major updates to documentation
* Updated gRPC REST Gateway entry points
* Cleaned up Cluster APIs by creating new Node Service
* Added InspectCurrent() to both Cluster and Node services
* Changed name of OpenStorageVolume.Set() to Update() to match others
* Cleaned up errors to return NotFound when the key is not found.
* Removed VolumeSpec from Clone() since it was not needed.

Closes #527

### DIFF
--- a/api/api.proto
+++ b/api/api.proto
@@ -309,8 +309,8 @@ message VolumeSpec {
   bool sharedv4 = 26;
 }
 
-// VolumeSpecSet provides a method to set any of the VolumeSpec of an existing volume
-message VolumeSpecSet {
+// VolumeSpecUpdate provides a method to set any of the VolumeSpec of an existing volume
+message VolumeSpecUpdate {
   // Ephemeral storage
   oneof ephemeral_opt { bool ephemeral = 1; }
   // Size specifies the thin provisioned volume size in bytes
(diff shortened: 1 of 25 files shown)


api/api.proto
```diff
@@ -805,7 +820,7 @@ service OpenStorageVolume {
   rpc Create(SdkVolumeCreateRequest)
     returns (SdkVolumeCreateResponse) {
       option(google.api.http) = {
-        post: "/v1/volume"
+        post: "/v1/volumes"
```
ram-infrac: why **volumes** -> volume create/delete etc will always be for single operation right?

api/api.proto
```diff
@@ -849,17 +864,18 @@ service OpenStorageVolume {
   rpc Enumerate(SdkVolumeEnumerateRequest)
     returns (SdkVolumeEnumerateResponse) {
       option(google.api.http) = {
-        post: "/v1/volume/enumerate"
+        post: "/v1/volumes/enumerate"
```
ram-infrac: should be GET call

api/api.proto
```diff
@@ -1412,82 +1476,94 @@ message SdkVolumeUnmountRequest {
...
   // Unique name of the volume. This will be used for idempotency.
   string name = 1;
-  // Parent volume id, if specified will create a new volume as a clone of the parent.
+  // Parent volume id or snapshot id will create a new volume as a clone of the parent.
   string parent_id = 2;
-  // Volume specification
-  VolumeSpec spec = 3;
 }
 
+// Defines the resonpse when creating a clone from a volume or a snapshot
```
ram-infrac: nit : response

api/api.proto
```diff
@@ -1796,70 +1950,91 @@ message SdkCloudBackupHistoryItem {
...
 
+// Empty response
 message SdkCloudBackupSchedDeleteResponse{
 }
 
+// Empty request
 message SdkCloudBackupSchedEnumerateRequest{
 }
 
+// Defies a response containing a map listing the schedules for volume
```
ram-infrac: nit: Defines

api/api.proto
lpabon: It was decided in the review to make them all plural

api/api.proto
lpabon: I tried to make this one a get, but it sends data to the server (the labels in the locator), so it will not allow a GET

ram-infrac: ahh, I know this issue

PR #535, opened 2018-07-13
PWX-5813: Remove whitespace from .stack & .heap filenames.

jrivera-px:

Remove spaces from the profile dump file names. 

Closes # PWX-5813

### DIFF
--- a/pkg/dbg/profile_dump.go
+++ b/pkg/dbg/profile_dump.go
@@ -28,11 +28,11 @@ func DumpGoMemoryTrace() {
 func DumpGoProfile() error {
 	trace := make([]byte, 5120*1024)
 	len := runtime.Stack(trace, true)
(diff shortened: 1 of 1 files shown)


pkg/dbg/profile_dump.go
```diff
@@ -28,11 +28,11 @@ func DumpGoMemoryTrace() {
 func DumpGoProfile() error {
 	trace := make([]byte, 5120*1024)
 	len := runtime.Stack(trace, true)
-	return ioutil.WriteFile(path+time.Now().String()+".stack", trace[:len], 0644)
+	return ioutil.WriteFile(path+time.Now().Format("2006-01-02T15:04:05.999999-0700MST")+".stack", trace[:len], 0644)
```
disrani-px: Use constants instead of hardcoding formats. Also time.RFC3339 is a similar format.

PR #542, opened 2018-07-16
Add field to volume to track if FS resize is required

disrani-px:

Adds a flag to track if fs resize should be run when a volume is mounted

Closes #


api/api.proto
```diff
@@ -307,6 +307,8 @@ message VolumeSpec {
   bool journal = 25;
   // Sharedv4 is true if this volume can be accessed via sharedv4.
   bool sharedv4 = 26;
+  // FsResizeRequired if an FS resize is required on the volume.
+  bool fs_resize_required = 27;
```
lpabon: Can this value be set in Volume.Set(). If it can, then this value needs to be added to VolumeSpecUpdate and to the SDK accordingly. If not, then just add a comment on this value saying that this value cannot be set on a volume after creation.

disrani-px: Doesn't need to be in the SDK. There is no enforcement for the flag here, it depends on the implementation, so a comment here, either way, would be wrong.

disrani-px: There is nothing preventing it from being changed. Some implementations could allow it to be changed. There shouldn't be a comment here for that.

lpabon: Ok, so if it can be changed, then it needs to add it to VolumeSpecUpdate and SDK server handler. Then the driver can decide if it can be changed.

PR #543, opened 2018-07-17
 sdk: Reorganize Enumerate and REST calls

### LINKED ISSUE #539: sdk: Enumerate calls should be GET calls
BUG

**What happened**:
The SDK Enumerate calls should be GETs. For this to correctly occur, we need the following:

* Enumerate calls changed to GET
* new APIs called EnumerateWithFilters() to use POST

Some SDK Enumerate calls are PUT/POST because they can pass in maps. The gRPC REST Gateway code generator does not allow GETs to pass data.

lpabon:
Make Enumerate calls be a simple GET with no parameters. If parameters are to be passed as filters, use the `EnumerateWithFilters` functions. This makes the REST interface much cleaner. Learning from this request, I also fixed some of the other REST paths.

Closes #539 


api/api.proto
```diff
@@ -1159,31 +1176,31 @@ service OpenStorageCloudBackup {
...
+        get: "/v1/cloudbackups/volume"
       };
     }
 
   // Status returns the status of any cloud backups of a volume
   rpc Status(SdkCloudBackupStatusRequest)
     returns (SdkCloudBackupStatusResponse) {
       option(google.api.http) = {
-        get: "/v1/cloudbackups/volume/{volume_id}/status"
+        get: "/v1/cloudbackups/volume/status"
```
ram-infrac: I didn't get part of avoiding volume_id here, get request will be ```"/v1/cloudbackups/volume/status``` for which volumes or all of them

PR #552, opened 2018-07-25
Do not filter decommissioned nodes in cluster.Enumerate

sangleganesh:
- cluster.Enumerate must return all nodes, even those decommissioned
- nodeCache is init when node joins cluster
- it is kept updated by the watch

Closes #

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -434,12 +434,13 @@ func (c *ClusterManager) watchDB(key string, opaque interface{},
 
 	c.size = db.Size
 
-	peers := c.getNonDecommisionedPeers(db)
-	c.gossip.UpdateCluster(peers)
+	c.gossip.UpdateCluster(c.getNonDecommisionedPeers(db))
+
+	// update the nodeCache and remove any nodes not present in cluster database
 	c.nodeCacheLock.Lock()
 	defer c.nodeCacheLock.Unlock()
 	for _, n := range c.nodeCache {
-		_, found := peers[types.NodeId(n.Id)]
+		_, found := db.NodeEntries[n.Id]
 		if !found {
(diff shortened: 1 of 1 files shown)


cluster/manager/manager.go
```diff
@@ -434,12 +434,13 @@ func (c *ClusterManager) watchDB(key string, opaque interface{},
...
-	peers := c.getNonDecommisionedPeers(db)
-	c.gossip.UpdateCluster(peers)
+	c.gossip.UpdateCluster(c.getNonDecommisionedPeers(db))
+
+	// update the nodeCache and remove any nodes not present in cluster database
 	c.nodeCacheLock.Lock()
 	defer c.nodeCacheLock.Unlock()
 	for _, n := range c.nodeCache {
-		_, found := peers[types.NodeId(n.Id)]
+		_, found := db.NodeEntries[n.Id]
```
adityadani: If we are keeping decommissioned nodes in cache then we need to modify getNodeEntry() at line 238 
to fill in node info from DB if the status is decommissioned.
```
else if n.Status == api.Status_STATUS_OFFLINE &&
		(n.DataIp == "" || n.MgmtIp == "") {
		// cached info unstable, read from DB
		if clustDBRef.Id == "" {
			// We've been passed "empty" struct, lazy-init before use
			clusterDB, _, _ := readClusterInfo()
			*clustDBRef = clusterDB
		}
		// Gossip does not have essential information of
		// an offline node. Provide the essential data
		// that we have in the cluster db
```

sangleganesh: the entry in nodeCache is already from clusterdb, and its gossip wouldn't update it since it does not talk to decommissioned peers. (see the PX PR which displays this info for your reference).

PR #554, opened 2018-07-25
Add field to snapshot request to avoid snapshot retries.

talakad:
Users can indicate if snapshot must continue to retrysnapshot creation if it
fails to create one during snapshot create command.

Closes #


lpabon:
@talakad and I discussed it offline and determined that the change to the SDK implemented in the PR is ok for now. If SDK clients need this and/or we see that more options are required, an `options map[string]string` may be added to the SnapshotCreateRequest and all its options defined. This would be too heavy for this PR, so we will wait until we need it. It is much easier to add, and impossible to remove things in the SDK.

PR #556, opened 2018-07-25
Add spec option requested_nodes_options

sangleganesh:

add a flag that goes along with zones, nodes, racks which says
that these inputs are only optional.

Signed-off-by: sangleganesh <ganesh@portworx.com>

Closes #

you can suggest some other name too.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -46,6 +46,9 @@ const (
 	SpecLabels               = "labels"
 	SpecPriorityAlias        = "priority_io"
 	SpecIoProfile            = "io_profile"
+	// SpecBestEffortLocationProvisioning default is false. If set provisioning request will succeed
+	// even if specified data location parameters could not be satisfied.
+	SpecBestEffortLocationProvisioning = "best_effort_location_provisioning"
 )
 
 // OptionKey specifies a set of recognized query params.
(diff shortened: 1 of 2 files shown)


api/api.go
```diff
@@ -14,38 +14,39 @@ import (
...
+	SpecAggregationLevel       = "aggregation_level"
+	SpecDedupe                 = "dedupe"
+	SpecPassphrase             = "secret_key"
+	SpecAutoAggregationValue   = "auto"
+	SpecGroup                  = "group"
+	SpecGroupEnforce           = "fg"
+	SpecZones                  = "zones"
+	SpecRacks                  = "racks"
+	SpecRack                   = "rack"
+	SpecRequestedNodesOptional = "requested_nodes_optional"
```
jvinod: SpecBestEffort="best_effort" // By default this is false. If set provisioning request will succeed even if specified data location parameters could not be satisfied.

api/api.go
```diff
@@ -42,10 +42,13 @@ const (
 	SpecZones                = "zones"
 	SpecRacks                = "racks"
 	SpecRack                 = "rack"
-	SpecRegions              = "regions"
-	SpecLabels               = "labels"
-	SpecPriorityAlias        = "priority_io"
-	SpecIoProfile            = "io_profile"
+	// SpecBestEffort default is false. If set provisioning request will succeed
+	// even if specified data location parameters could not be satisfied.
+	SpecBestEffort    = "best_effort"
```
adityadani: nit: can we rename it to "best_effort_provisioning". This way it makes it clear to the user as to what this spec is doing. The term best_effort can be applied to many of the spec parameters here. For ex. if this spec does not apply to io_priority then we can name it specifically as "best_effort_location_provisioning" or something similar.

sangleganesh:
@adityadani / @jv updated diff.

PR #561, opened 2018-08-06
sdk: Support for Versions

### LINKED ISSUE #560: Adding versioning to the SDK
FEATURE

**What happened**:
Clients and servers should be able to determine the version of their generated code.

## Proposal

To do this, I am proposing to add the following to the proto file (ignore the values, they are just examples):

```diff
diff --git a/api/api.proto b/api/api.proto
index 9318efb3..ddb724e3 100644
--- a/api/api.proto
+++ b/api/api.proto
@@ -19,6 +19,20 @@ option go_package = "api";
 option java_multiple_files = true;
 option java_package = "com.openstorage.api";
 
+// SDK Verion is Major.Minor.Patch
+enum SdkVersion {
+  // Must be set to allow enums to have the same value
+  option allow_alias = true;
+  // This value must be here always set to zero
+  MUST_HAVE_ZERO_VALUE = 0;
+
+  // Major Version
+  SdkVersionMajor = 0;
+  // Minor Version
+  SdkVersionMinor = 14;
+  // Patch Version
+  SdkVersionPatch = 0;
+}
+
 enum Status {
   STATUS_NONE = 0;
   STATUS_INIT = 1;
```

This allows clients and servers to determine their SDK RPC version. It also provides the possibility to interrogate a server and ask it what version it is using.

### Example output in Go

```go
// SDK Verion is Major.Minor.Patch
type SdkVersion int32

const (
       // This value must be here always set to zero
       SdkVersion_MUST_HAVE_ZERO_VALUE SdkVersion = 0
       // Major Version
       SdkVersion_SdkVersionMajor SdkVersion = 0
       // Minor Version
       SdkVersion_SdkVersionMinor SdkVersion = 14
       // Patch Version
       SdkVersion_SdkVersionPatch SdkVersion = 0
)
```

lpabon:
Needed a method to retrieve version information from the server.

Closes #560 

Adds a new SDK function to determine both the SDK version used and the version of the server. A new Version() call was added to the volume interface to provide a model for drivers to return version information.


lpabon:
> Lets set the driver versions to 1.0.0 instead of 0.0.0
Sounds good! 👍

PR #563, opened 2018-08-08
travis: Add SDK tests to travis

lpabon:
Provides tests for the SDK api as expected by [sdk-test](https://github.com/libopenstorage/sdk-test). Since sdk-test has been compiled with a previous version of the `api.proto`, it ensures that any changes to the current API does not break previous clients.


lpabon:
Works:

PR #564, opened 2018-08-08
Add LocateResponse object

pault84:

Added a Locate object that can be used to populate the VolumeConsumer response


api/api.proto
```diff
@@ -2483,3 +2483,10 @@ message CatalogResponse {
 	// Report of total directories and files count
 	Report report = 2;
 }
+
+message LocateResponse {
```
lpabon: Could you add documentation to the `LocateResponse` object? What it is used for, why it is needed, etc.

api/api.proto
```diff
@@ -2483,3 +2483,10 @@ message CatalogResponse {
...
+    //  Map of mounts
+    map<string, string> mounts = 1;
```
lpabon: What are the keys and values? Could you add an example to the comment?

PR #566, opened 2018-08-10
Fix vmware vm uuid parsing and add DeleteFrom and DetachFrom

harsh-px:

1. User doesn't need to give vsphere vmuuid from environment. It's optional.
2. Add DeleteFrom and DetachFrom API's to detach and delete if invoked from a different node

### DIFF
--- a/pkg/storageops/vsphere/vsphere_util.go
+++ b/pkg/storageops/vsphere/vsphere_util.go
@@ -80,10 +80,7 @@ func ReadVSphereConfigFromEnv() (*VSphereConfig, error) {
 		cfg.InsecureFlag = true
 	}
 
-	cfg.VMUUID, err = storageops.GetEnvValueStrict("VSPHERE_VM_UUID")
-	if err != nil {
-		return nil, err
-	}
+	cfg.VMUUID, _ = storageops.GetEnvValueStrict("VSPHERE_VM_UUID")
 
 	return &cfg, nil
 }
--- a/pkg/storageops/aws/aws.go
+++ b/pkg/storageops/aws/aws.go
@@ -516,6 +516,10 @@ func (s *ec2Ops) Create(
 	return s.refreshVol(resp.VolumeId)
 }
 
+func (s *ec2Ops) DeleteFrom(id, _ string) error {
+	return s.Delete(id)
+}
+
 func (s *ec2Ops) Delete(id string) error {
 	req := &ec2.DeleteVolumeInput{VolumeId: &id}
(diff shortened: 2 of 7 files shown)


lpabon:
FYI: title has `add DeleteFrom and DeleteFrom`

pkg/storageops/vsphere/vsphere_util.go
```diff
@@ -80,10 +80,7 @@ func ReadVSphereConfigFromEnv() (*VSphereConfig, error) {
 		cfg.InsecureFlag = true
 	}
 
-	cfg.VMUUID, err = storageops.GetEnvValueStrict("VSPHERE_VM_UUID")
-	if err != nil {
-		return nil, err
-	}
+	cfg.VMUUID, _ = storageops.GetEnvValueStrict("VSPHERE_VM_UUID")
```
lpabon: curious why this change is needed?

harsh-px: VMUUID is required to be fetched from the env variable`VSPHERE_VM_UUID` only if one is running unit tests for dev.

When running in a real vSphere environment, Portworx (porx) will fetch this from a cloudprovider field that k8s populates in the node object. So we don't need this env. For e.g

```
- apiVersion: v1
  kind: Node
  metadata:
    creationTimestamp: 2018-08-13T03:46:12Z
  spec:
    externalID: 421076c2-e509-7897-e64f-84b0bff79cbe
    providerID: vsphere://421076c2-e509-7897-e64f-84b0bff79cbe
```

sangleganesh:
@harsh-px just for my info, is there a reason this code is a part of libopenstorage ?

sangleganesh:
@harsh-px iirc the reason storage ops of AWS was put in libopenstorage was to have AWS volume driver code reuse the storage ops. gke/vmware don't have that dependency (does not have volume driver implementation for osd). is there any interface/implementation required for osd itself from either gke/vmware ?

harsh-px:
> Please add a UT esp. for DetachFrom for completeness.

@adityadani added

> @harsh-px iirc the reason storage ops of AWS was put in libopenstorage was to have AWS volume driver code reuse the storage ops. gke/vmware don't have that dependency (does not have volume driver implementation for osd). is there any interface/implementation required for osd itself from either gke/vmware ?

Nothing in osd is using gke/vmware. Currently it's only porx. Project rico in libopenstorage was going to use all of these storage ops (https://github.com/libopenstorage/rico/tree/master/pkg/cloudprovider).

PR #570, opened 2018-08-16
sdk: Add NAME to credentials service

### LINKED ISSUE #569: add new name field in api for creds
bug report

**What happened**:
we need an ability to manage creds based on name. currently creds are being managed using a uuid, which can be inconvenient. so we need a new constant to access name.

Add new const for name around here:
https://github.com/libopenstorage/openstorage/blob/master/api/api.go#L72

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lpabon:
Add `name` to credentials as it is done for Volumes and Snapshots

Closes #569

* Had to remove the tests for the sdk from Travis-ci since they would fail as sdk-test would need to be updated
* SDK version bumped to 0.2.0


api/api.proto
```diff
@@ -1379,13 +1381,14 @@ message SdkSchedulePolicyInterval {
   // Number of instances to retain
   int64 retain = 1;
 
+  // Start oneof at field number 200 to allow for expansion
   oneof period_type {
     // Daily policy
-    SdkSchedulePolicyIntervalDaily daily = 2;
+    SdkSchedulePolicyIntervalDaily daily = 200;
```
sdeoras: how are these related to cred name change?

lpabon: Due to this change, I wanted to prepare the message for future ones that may come. I'm making sure we have enough field numbers to allow for expansion while trying to keep the oneof's at the end of the message. That way we can add more in the oneof's.

Long story short, this change brought up that we needed to make sure that our messages were easily enhanced. All others messages are ok, just the ones with oneof needed to be adjusted.

PR #574, opened 2018-08-18
Add Name field to node to store the scheduler node name

### LINKED ISSUE #573: Add node name to node/storage node object
Feature request

**What happened**:
In case of cloud environments, the scheduler nodes keep changing, but the storage id does not change as there is storage associated to it.
It is convenient to know the scheduler names which is why we plan to store in the node object and the cluster database. In most cases it will be the same, but could be different in case of cloud environments with auto scaling groups.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

piyush-nimbalkar:

Closes #573 


api/api.proto
```diff
@@ -728,6 +728,8 @@ message StorageNode {
   string hostname = 15;
   // User defined labels for the node
   map<string, string> node_labels = 16;
+  // Name of the node
+  string name = 17;
```
lpabon: See `api.go`. So this should be `scheduler_host_name`. The comments on this file are directly used for the documentation website.

api/server/sdk/node_test.go
```diff
@@ -220,6 +222,7 @@ func TestSdkNodeInspectCurrent(t *testing.T) {
 	// Create response
 	node := api.Node{
 		Id:       nodeid,
+		Name:     "nodename",
```
lpabon: Verify this value down below.

lpabon:
Bump the SDK version to `0.3.0`. We should have a SDK_CHANGELOG.md in the root of the repo. Could you copy https://github.com/libopenstorage/libopenstorage.github.io/blob/master/docs/changelog.md as SDK_CHANGELOG.md in the root and update it accordingly? That will make it easier to just take this file and use it for the website: https://libopenstorage.github.io/w/changelog.html .

lpabon:
Lastly, the unit tests are failing because the new value is not being copied from `api.Node` to `api.pb.go.StorageNode` in 

https://github.com/libopenstorage/openstorage/blob/master/api/api.go#L709-L721

lpabon:
Do think it should be also added here:

https://github.com/libopenstorage/openstorage/blob/e730b1983ba4c8d125e675a6aacf93edc5574482/cluster/manager/manager.go#L249-L253

lpabon:
One more place. Take a look at adding it here also:

https://github.com/libopenstorage/openstorage/blob/e730b1983ba4c8d125e675a6aacf93edc5574482/cluster/manager/manager.go#L1162-L1170

If you do, it could be beneficial to also add it to [config.go](https://github.com/libopenstorage/openstorage/blob/master/config/config.go#L30) and then add a fake name in [config-fake.yaml](https://github.com/libopenstorage/openstorage/blob/master/etc/config/config-fake.yaml).

Then you can run `make launch-sdk`, which creates `openstorage/mock-sdk-server` container with the `config-fake.yaml` and runs it. If you then run the command:

```
curl -X GET "http://localhost:9110/v1/nodes/current" -H "accept: application/json" |  jq
```

Then you should see the new variable and value you have set.

sangleganesh:
prefer to store information which does not change,

api/api.go
```diff
@@ -140,6 +140,8 @@ const (
 type Node struct {
 	// Id of the node.
 	Id string
+	// Name of the node.
```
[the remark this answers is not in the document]
piyush-nimbalkar: Calling it scheduler node name to be clear that it is a node name and the name of the scheduler.

piyush-nimbalkar:
@lpabon The additions you mentioned were already present.
`launch-sdk` is cool!

PR #576, opened 2018-08-20
AWS StorageOps: Add support for provisioning nvme drives.

adityadani:
- For some instance types like (C5, C5d, i3.metal, M5, M5d, R5, R5d, and z1d)
AWS provisions EBS volumes and exposes them as NVMe drives.
- To an Attach call, AWS takes in the usual /dev/xvd path.
- The device names that is specified in the block device mapping are renamed using NVMe device names (/dev/nvme[0-26]n1).

### DIFF
--- a/pkg/storageops/aws/aws.go
+++ b/pkg/storageops/aws/aws.go
@@ -37,19 +37,19 @@ var ErrAWSEnvNotAvailable = fmt.Errorf("AWS credentials are not set in environme
 
 // NewEnvClient creates a new AWS storage ops instance using environment vars
 func NewEnvClient() (storageops.Ops, error) {
-	region := os.Getenv("AWS_REGION")
-	if len(region) == 0 {
-		return nil, ErrAWSEnvNotAvailable
+	region, err := storageops.GetEnvValueStrict("AWS_REGION")
+	if err != nil {
+		return nil, err
 	}
 
-	instance := os.Getenv("AWS_INSTANCE_NAME")
-	if len(instance) == 0 {
-		return nil, ErrAWSEnvNotAvailable
+	instance, err := storageops.GetEnvValueStrict("AWS_INSTANCE_NAME")
+	if err != nil {
+		return nil, err
 	}
 
-	instanceType := os.Getenv("AWS_INSTANCE_TYPE")
-	if len(instanceType) == 0 {
-		return nil, ErrAWSEnvNotAvailable
+	instanceType, err := storageops.GetEnvValueStrict("AWS_INSTANCE_TYPE")
+	if err != nil {
+		return nil, err
 	}
 
 	if _, err := credentials.NewEnvCredentials().Get(); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/storageops/aws/aws.go
```diff
@@ -43,6 +47,11 @@ func NewEnvClient() (storageops.Ops, error) {
 		return nil, ErrAWSEnvNotAvailable
 	}
 
+	instanceType := os.Getenv("AWS_INSTANCE_TYPE")
```
harsh-px: Can you update the README.md for this new env variable?

harsh-px: Also trim whitespaces here just to be safe since we are using this to detect nvme's later on.

pkg/storageops/aws/aws.go
```diff
@@ -56,17 +65,21 @@ func NewEnvClient() (storageops.Ops, error) {
...
-		instance: instance,
-		ec2:      ec2,
+		instance:     instance,
+		instanceType: instanceType,
+		ec2:          ec2,
 	}
 }
 
+// nvmeInstanceTypes are list of instance types whose EBS volumes are exposed as NVMe block devices
+var nvmeInstanceTypes = []string{"c5", "c5d", "i3.metal", "m5", "m5d", "r5", "r5d", "z1d"}
```
harsh-px: this would mean we have to stay updated with the aws instance types

pkg/storageops/aws/aws.go
adityadani: I think it should be done for all the variables here.  If one fails nothing from storageops/aws is going to really work. I will add it for all variables.

pkg/storageops/aws/aws.go
adityadani: Yes that is right and this is the only we can detect a nvme drive.

pkg/storageops/aws/aws.go
harsh-px: Sounds good. Add it to `GetEnvValueStrict` in utils.go and use that instead.

pkg/storageops/aws/aws.go
```diff
@@ -281,15 +297,43 @@ func (s *ec2Ops) getPrefixFromRootDeviceName(rootDeviceName string) (string, err
...
+	if !found {
+		return "", fmt.Errorf("unable to map %v block device to an"+
+			" actual device path on the host", ipDevicePath)
+	}
+	// We name our EBS volumes /dev/xvd[f-p]. The nvme devices are serially
+	// named based on the input EBS volume device name is given.
+	// /dev/xvda -> /dev/nvme0 is reserved for root device.
+	// /dev/xvdf maps to /dev/nvme1 and so on
+	nvmeCount := letter[0] - 'f' + 1
+	devicePath = awsDevicePrefixNvme + strconv.FormatInt(int64(nvmeCount), 10) + "n1"
```
harsh-px: so an example `devicePath` will be `/dev/nvme1n1` ? Can you update the comment with the example.

volume/drivers/aws/aws.go
```diff
@@ -68,11 +68,15 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
...
+	instanceID, err := metadata("instance-id")
 	if err != nil {
 		return nil, err
 	}
-	logrus.Infof("AWS instance %v zone %v", instance, zone)
+	instanceType, err := metadata("instance-type")
+	if err != nil {
+		return nil, err
+	}
+	logrus.Infof("AWS instance %v zone %v", instanceID, zone)
```
piyush-nimbalkar: nit: print the instance type as well?

PR #578, opened 2018-08-22
AWS: Add a reliable way to determine nvme drive names

adityadani:

Use `nvme list` command to fetch the nvme drives.

- The current way of determining the nvme drive name is not reliable as
  the drive names might change due to different AMI types. Also the current logic
  breaks if an existing drive on the instance has a non-standard names such as /dev/xvdcn
- Instead use `nvme list` command to reliably get the mapping between EBS volumeID and a nvme drive.

Here is a sample output of `nvme list` command.

```
root@ip-172-20-52-215:/# nvme list
Node             SN                   Model                                    Namespace Usage                      Format           FW Rev
---------------- -------------------- ---------------------------------------- --------- -------------------------- ---------------- --------
/dev/nvme0n1     vol05ce9829d6b2d3a21 Amazon Elastic Block Store               1           0.00   B / 137.44  GB    512   B +  0 B   1.0
/dev/nvme3n1     vol055b3ae393bfc407b Amazon Elastic Block Store               1           0.00   B /  17.18  GB    512   B +  0 B   1.0
/dev/nvme2n1     vol0401d9e335cae567d Amazon Elastic Block Store               1           0.00   B /   8.59  GB    512   B +  0 B   1.0
/dev/nvme1n1     vol067baf02ad3ecd9fd Amazon Elastic Block Store               1           0.00   B / 107.37  GB    512   B +  0 B   1.0
```

### DIFF
--- a/pkg/storageops/aws/aws.go
+++ b/pkg/storageops/aws/aws.go
@@ -3,7 +3,6 @@ package aws
 import (
 	"fmt"
 	"os"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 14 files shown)


pkg/storageops/aws/aws.go
```diff
@@ -32,8 +33,11 @@ type ec2Ops struct {
 	mutex        sync.Mutex
 }
 
-// ErrAWSEnvNotAvailable is the error type when aws credentails are not set
-var ErrAWSEnvNotAvailable = fmt.Errorf("AWS credentials are not set in environment")
+var (
+	// ErrAWSEnvNotAvailable is the error type when aws credentails are not set
```
jvinod: credentials

PR #579, opened 2018-08-22
Add retry loop to Request if Service Unavailable. Default wait is 1 second

pault84:

Currently, the client always exits upon receiving service unavailable.
This code change will retry until success and wait for 1 second or amount of seconds suggested in the Retry-After header

This way we can limit the number of connections on the server and ask the client to retry if the queue is full.

### DIFF
--- a/api/client/request.go
+++ b/api/client/request.go
@@ -15,6 +15,10 @@ import (
 	"time"
 )
 
+const (
+	maxRetryDuration = 5 * time.Minute
+)
+
 // Request is contructed iteratively by the client and finally dispatched.
 // A REST endpoint is accessed with the following convention:
 // base_url/<version>/<resource>/[<instance>]
@@ -226,6 +230,7 @@ func parseHTTPStatus(resp *http.Response, body []byte) error {
 }
 
 // Do executes the request and returns a Response.
+// Do executes the request and returns a Response.
 func (r *Request) Do() *Response {
 	var (
 		err  error
@@ -234,6 +239,7 @@ func (r *Request) Do() *Response {
 		url  string
 		body []byte
 	)
+
 	if r.err != nil {
 		return &Response{err: r.err}
 	}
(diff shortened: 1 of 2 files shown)


api/client/request.go
```diff
@@ -258,17 +261,30 @@ func (r *Request) Do() *Response {
...
+	for i := 0; ; i++ {
+		resp, err = r.client.Do(req)
+		if err != nil {
+			return &Response{err: err}
+		}
+		if resp.Body != nil {
+			defer resp.Body.Close()
+			body, err = ioutil.ReadAll(resp.Body)
+			if resp.StatusCode == http.StatusServiceUnavailable {
+				if retryafter, err := strconv.Atoi(resp.Header["Retry-After"][0]); err == nil {
```
sangleganesh: retry detecting code must check if len(resp.Header["Retry-After"]) > 0 before accessing [0].

api/client/request.go
```diff
@@ -258,17 +261,30 @@ func (r *Request) Do() *Response {
...
+					wait = time.Duration(retryafter) * time.Second
+				}
+			} else {
```
sangleganesh: return response instead of breaking ?

api/client/request.go
pault84: good catch

api/client/request.go
pault84: Makes sense

api/client/request.go
```diff
@@ -258,17 +261,30 @@ func (r *Request) Do() *Response {
...
-	if err != nil {
-		return &Response{err: err}
-	}
-	if resp.Body != nil {
-		defer resp.Body.Close()
-		body, err = ioutil.ReadAll(resp.Body)
-	}
-	if err != nil {
-		return &Response{err: err}
+	for i := 0; ; i++ {
```
lpabon: This looks like it could block forever. Set a maximum timeout value or max number of retries.

api/client/request.go
```diff
@@ -258,17 +261,30 @@ func (r *Request) Do() *Response {
...
+		resp, err = r.client.Do(req)
```
lpabon: To test this, you can write a temporary http handler in httptest, and have it delay or return the status code needed to start the timer.

PR #581, opened 2018-08-24
Add queue_depth setting to volume spec

jvinod:

Add queue_depth to device spec


api/api.proto
```diff
@@ -308,6 +308,8 @@ message VolumeSpec {
   bool journal = 25;
   // Sharedv4 is true if this volume can be accessed via sharedv4.
   bool sharedv4 = 26;
+  // QueueDepth defines the desired block device queue depth
+  int32 queue_depth = 27;
```
lpabon: Need to add this to VolumeSpecUpdate message. Then need to add this to [mergeSpecs()](https://github.com/libopenstorage/openstorage/blob/master/api/server/sdk/volume_ops.go#L332)

lpabon: Also, bump the version of the SDK, and add a comment to SDK_CHANGELOG.md in the root of the tree.

https://github.com/libopenstorage/openstorage/blob/2205e14a19248a554ea635ca62d80187b70063e1/api/api.proto#L2247-L2252

PR #587, opened 2018-08-25
Return VolNotFound if disk in not present in gcloud

piyush-nimbalkar:
Need to handle NotFound error explicitly,
so that client know that the drive is not present.

Closes #

### DIFF
--- a/pkg/storageops/gce/gce.go
+++ b/pkg/storageops/gce/gce.go
@@ -3,6 +3,7 @@ package gce
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"path"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/storageops/gce/gce.go
```diff
@@ -265,7 +267,13 @@ func (s *gceOps) DeviceMappings() (map[string]string, error) {
 
 func (s *gceOps) DevicePath(diskName string) (string, error) {
 	d, err := s.service.Disks.Get(s.inst.project, s.inst.zone, diskName).Do()
-	if err != nil {
+	if gerr, ok := err.(*googleapi.Error); ok &&
+		gerr.Code == http.StatusNotFound {
+		return "", storageops.NewStorageError(
+			storageops.ErrVolNotFound,
+			fmt.Sprintf("Disk: %s not found in zone %s", diskName, s.inst.Zone),
```
harsh-px: this won't build

```
pkg/storageops/gce/gce.go:274:65: s.inst.Zone undefined (type *instance has no field or method Zone, but does have zone)
pkg/storageops/gce/gce.go:275:10: s.inst.Name undefined (type *instance has no field or method Name, but does have name)
```

PR #591, opened 2018-08-26
Putting device path check in storageops for GCE

piyush-nimbalkar:
- throwing error for DeviceMappings if unable to get actual device path
- throwing new error ErrInvalidDevicePath for device path check errors

Closes #

### DIFF
--- a/pkg/storageops/gce/gce.go
+++ b/pkg/storageops/gce/gce.go
@@ -306,8 +306,8 @@ func (s *gceOps) DevicePath(diskName string) (string, error) {
 
 	for _, instDisk := range inst.Disks {
 		if instDisk.Source == d.SelfLink {
-			pathbyID := fmt.Sprintf("%s%s", googleDiskPrefix, instDisk.DeviceName)
-			devPath, err := s.diskIDToBlockDevPathWithRetry(pathbyID)
+			pathByID := fmt.Sprintf("%s%s", googleDiskPrefix, instDisk.DeviceName)
+			devPath, err := s.diskIDToBlockDevPathWithRetry(pathByID)
 			if err == nil {
 				return devPath, nil
 			}
@@ -712,6 +712,9 @@ func (s *gceOps) diskIDToBlockDevPath(devPath string) (string, error) {
 		}
 
 		devPath = strings.TrimSpace(string(output))
+	} else {
+		return "", fmt.Errorf("%s was expected to be a symlink to actual "+
+			"device path", devPath)
 	}
 
 	return devPath, nil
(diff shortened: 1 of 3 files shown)


pkg/storageops/gce/gce.go
```diff
@@ -651,6 +677,46 @@ func (s *gceOps) getDisksFromAllZones(labels map[string]string) (map[string]*com
...
+		retryCount++
+		if retryCount >= devicePathMaxRetryCount {
+			break
+		}
+		time.Sleep(devicePathRetryInterval)
+	}
+	return "", err
+}
+
+func (s *gceOps) diskIDToBlockDevPath(devPath string) (string, error) {
```
adityadani: @harsh-px If this function is called only for storageops perspective, should it return an error if it is unable to find a symbolic link to an actual device path?

pkg/storageops/gce/gce.go
```diff
@@ -597,7 +620,10 @@ func (s *gceOps) waitForAttach(
 	devicePath, err := task.DoRetryWithTimeout(
 		func() (interface{}, bool, error) {
 			devicePath, err := s.DevicePath(disk.Name)
-			if err != nil {
+			if se, ok := err.(*storageops.StorageError); ok &&
```
adityadani: In the DevicePath API(), we are not determining whether a volume is attached on remote node, but inferring it if it the disk shows as attached and the instance does not show any reference for this disk.

From GCE perspective is it safe to assume that, Disk's status changing and the Disk appearing in DesribeInstance output an atomic operation?

pkg/storageops/gce/gce.go
```diff
@@ -259,7 +266,15 @@ func (s *gceOps) DeviceMappings() (map[string]string, error) {
 			continue
 		}
 
-		m[fmt.Sprintf("%s%s", googleDiskPrefix, d.DeviceName)] = path.Base(d.Source)
+		pathByID := fmt.Sprintf("%s%s", googleDiskPrefix, d.DeviceName)
+		devPath, err := s.diskIDToBlockDevPath(pathByID)
```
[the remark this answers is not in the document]
piyush-nimbalkar: DeviceMappings is called to get properties. 
- This is called after the drive is attached and verified.
- If this fails, we return the DefaultProvider properties anyway
Hence did not want to delay the initialization.

pkg/storageops/gce/gce.go
harsh-px: Yes, this function is expected to pass for all paths (symbolic and regular). So any error would be unexpected and not ignored.

If storageops fails to find the dev path and does not return error, the path will not be useful to callers who expected a block device path.

pkg/storageops/gce/gce.go
harsh-px: Discussed offline. This can be improved to just use disk users.

PR #594, opened 2018-08-29
Add a new NULL implementation for Cluster interface.

adityadani:
The cluster pkg has two primary interfaces
- Cluster
- ClusterListener

The ClusterManger in openstorage implements the Cluster interface. It also invokes the APIs
from the ClusterListener interface. Whoever implements the ClusterListener interface will get
events from the ClusterManager. By adding a NULL implementation we would be able to write tests
for the implementors for ClusterListener interface.


adityadani:
@sangleganesh I have updated the PR.

PR #608, opened 2018-09-11
alerts new filter types and enum bug fix

### LINKED ISSUE #572: redesign alerts interface
feature request

**What happened**:
we would like to work with alerts in such a way that alerts are raised and deleted by both the client functions and also via events. For instance, raising an alert A could delete an alert B and so on. We want to be able to express such dependencies and keep the interface simple to reason about.

So this issue is to discuss the design proposal before proceeding to implementation.

1) A new package called alerts would be written
2) It would define an interface for basic alert operations
```
// Manager manages alerts.
type Manager interface {
	// Raise raises an alert.
	Raise(alert *api.Alert) error
	// Enumerate lists all alerts filtered by a chain of filters.
	Enumerate(filters ...Filter) ([]*api.Alert, error)
	// Delete deletes alerts filtered by a chain of filters.
	Delete(filters ...Filter) error
	// SetRules sets a set of rules to be performed on alert events.
	SetRules(rule ...Rule) error
}
```
So the basic idea is to define filters and rules. For instance, various filters could be defined as follows:
```
// MatchAlertType creates a filter that matches on alert type.
func MatchAlertType(alertType int64) Filter {
	return &filter{filter: matchAlertType, value: alertType}
}

// MatchResourceType creates a filter that matches on resource type.
func MatchResourceType(resourceType api.ResourceType) Filter {
	return &filter{filter: matchResourceType, value: resourceType}
}

// MatchResourceID creates a filter that matches on resource id.
func MatchResourceID(resourceID string) Filter {
	return &filter{filter: matchResourceID, value: resourceID}
}

// MatchAny creates a filter that matches on any alert.
func MatchAny() Filter {
	return &filter{filter: matchAny}
}

// MatchTime creates a filter that matches on alert raised in a given time window.
func MatchTime(start, stop time.Time) Filter {
	return &filter{filter: matchTime, value: timeZone{start: start, stop: stop}}
}
```
and various rules could be defined as follows:
```
// OnRaise creates a rule that activates when a raised alerts matches filter.
func OnRaise(filter Filter, f func()) Rule {
	return &rule{event: onRaise, filter: filter, f: f}
}

// OnDelete creates a rule that activates when deleted alert matches filter.
func OnDelete(filter Filter, f func()) Rule {
	return &rule{event: onDelete, filter: filter, f: f}
}
```
Furthermore, an implementation of these interfaces would be done against kvdb.

Basic question we want to answer here is the type of kvdb queries we want to make. All paths to reach alerts should try to minimize the kvdb traffic. In other words, the idea is to also design kvdb tree structs such that Filters reflect easy querying on such kvdb tree.

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

sdeoras:

This PR is a follow up on previous pr #575 and fixes a bug and adds new filter types.
Also added deprecated tag to older alert pkg.

Closes #572 

### DIFF
--- a/alerts/api.go
+++ b/alerts/api.go
@@ -127,11 +127,11 @@ func NewCustomAction(f func(manager Manager, filters ...Filter) error, filters .
 // NewRaiseRule creates a rule that runs action when a raised alerts matche filter.
 // Action happens before incoming alert is raised.
 func NewRaiseRule(name string, filter Filter, action Action) Rule {
-	return &rule{name: name, event: RaiseEvent, filter: filter, action: action}
+	return &rule{name: name, event: raiseEvent, filter: filter, action: action}
 }
 
 // NewDeleteRule creates a rule that runs action when deleted alerts matche filter.
 // Action happens after matching alerts are deleted.
 func NewDeleteRule(name string, filter Filter, action Action) Rule {
-	return &rule{name: name, event: DeleteEvent, filter: filter, action: action}
+	return &rule{name: name, event: deleteEvent, filter: filter, action: action}
 }
--- a/alerts/api.go
+++ b/alerts/api.go
@@ -16,81 +16,88 @@ func NewManager(kv kvdb.Kvdb, options ...Option) (Manager, error) {
 
 // NewTTLOption provides an option to be used in manager creation.
 func NewTTLOption(ttl uint64) Option {
-	return &option{optionType: TTLOption, value: ttl}
+	return &option{optionType: ttlOption, value: ttl}
 }
 
 // NewTimeSpanOption provides an option to be used in filter definition.
-// Filters that take options, apply options only during matching
+// Filters that take options, apply options only during matching alerts.
 func NewTimeSpanOption(start, stop time.Time) Option {
-	return &option{optionType: TimeSpanOption, value: NewTimeSpanFilter(start, stop)}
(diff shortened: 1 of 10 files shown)


alerts/api.go
```diff
@@ -42,14 +42,25 @@ func NewTimeSpanFilter(start, stop time.Time) Filter {
...
 }
 
 // NewCountSpanFilter provides a filter that matches on alert count.
 func NewCountSpanFilter(minCount, maxCount int64) Filter {
 	return &filter{filterType: CountSpanFilter, value: []int64{minCount, maxCount}}
 }
 
+// NewMinSeverityFilter provides a filter that matches on alert when severity is greater than
+// or equal to the minSev value.
+func NewMinSeverityFilter(minSev api.SeverityType) Filter {
```
adityadani: lets reduce the scope of minSeverityFilter to only specific resource types.
In this way we can only query a subtree.
For ex. All WARN node resource alerts

sdeoras: i added a way to tell efficient filters how to filter after fetching. This is done using options and each of the efficient filters can take a number of options. So for instance, you could fetch using resource type and then have the fetched filters go through timeSpan check or severity check or both. Options provide a generic way to express such fetching/filtering for any of the efficient filters.

PR #613, opened 2018-09-13
Document CloudbackupEnumerateRequest better

pault84:
There was some confusion regarding the use of SdkCloudBackupEnumerateRequest
this documents the valid combinations and their expectations.

### DIFF
--- a/api/api.proto
+++ b/api/api.proto
@@ -1890,8 +1890,8 @@ message SdkCloudBackupDeleteAllResponse {
 // and do not provide `cluster_id` and `all`.
 // * For a specific volume in a specific cluster: Set `src_volume_id` to your desired volume id
 // and specify `cluster_id`.
-// * For a specific volume in all clusteres: Set `src_volume_id` to your desired volume id
-// and specify set `all` to true, do not provide `cluster_id`.
+// * For a specific volume in all clusters: Set `src_volume_id` to your desired volume id
+// and set `all` to true, do not provide `cluster_id`.
 // * For all volumes in current cluster: do not provide `cluster_id`, `volume_id` and `all`.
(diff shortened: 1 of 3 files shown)


api/api.proto
```diff
@@ -1883,15 +1883,27 @@ message SdkCloudBackupDeleteAllRequest {
...
 
-// Defines a request to list the backups stored by a cloud provider
+// Defines a request to list the backups stored by a cloud provider.
+// The following combinations can be used to get cloud backup information:
+//
+// * For a specific volume in current cluster: Set `src_volume_id` to your desired volume id
+// and do not provide `cluster_id` and `all`.
+// * For a specific volume in a specific cluster: Set `src_volume_id` to your desired volume id
+// and specify `cluster_id`.
+// * For a specific volume in all clusteres: Set `src_volume_id` to your desired volume id
```
lpabon: clusteres

api/api.proto
```diff
@@ -1883,15 +1883,27 @@ message SdkCloudBackupDeleteAllRequest {
...
+// and specify set `all` to true, do not provide `cluster_id`.
```
lpabon: 'specify set' should probably be 'set'

api/api.proto
```diff
@@ -1883,15 +1883,27 @@ message SdkCloudBackupDeleteAllRequest {
...
+// * For all volumes in current cluster: do not provide `cluster_id`, `volume_id` and `all`.
```
lpabon: 'and all' or 'or all'?

pault84: i did not want them to think they shouldn't provide volume_id OR 'all' they shouldn't provide any

PR #617, opened 2018-09-18
sdk: Update Mount/Attach/Detach/Unmount APIs

### LINKED ISSUE #601: Umount and/or Detach remote volumes from the SDK
FEATURE

**What happened**:
If possible, determine a way to unmount and/or detach a volume that is attached to another node.

lpabon:
1. Options needed to be added to the APIs (except for Mount)
2. Breaking change!!: moved the API calls from the OpenStorageVolume service to the new OpenStorageMountAttach service.

Closes #601

Removed `test-sdk` from travis for this release since sdk-test will need to be updated due to braking changes

Here are the options to use from discussions with @adityadani and @piyush-nimbalkar : 

* Attach: Options Secret, SecretKey, and SecretContext only for attach
* Detach: Options ForceDetach, UnmountBeforeDetach as bools and always set RedirectDetach (do not give the choice)
* Mount: No options
* Unmount: Options DeleteAfterUnmount as a bool. WaitBeforeDelete will be always set but the caller can have it turned off.

More info on options here:

https://github.com/libopenstorage/openstorage/blob/5919918c912441715c152be19e10959c69f0b77c/pkg/options/options.go#L9-L51


api/server/sdk/volume_node_ops.go
```diff
@@ -55,9 +84,35 @@ func (s *VolumeServer) Detach(
 		return nil, status.Error(codes.InvalidArgument, "Must supply volume id")
 	}
 
-	err := s.driver.Detach(req.GetVolumeId(), nil)
+	// Check if already attached
```
adityadani: nit: Check if already detached.

api/api.proto
```diff
@@ -1531,7 +1531,17 @@ message SdkVolumeUnmountRequest {
   // MountPath for device
   string mount_path = 2;
   // Options to unmount device
-  map<string, string> options = 3;
+  message Options {
+    // Delete the mount path on the node after unmounting
+    bool delete_mount_path = 1;
```
adityadani: lets update this to `do_not_delete_mount_path` and always delete the mountpath by default.

api/server/sdk/volume_node_ops.go
```diff
@@ -93,11 +153,21 @@ func (s *VolumeServer) Unmount(
 		return nil, status.Error(codes.InvalidArgument, "Invalid Mount Path")
 	}
 
-	err := s.driver.Unmount(req.GetVolumeId(), req.GetMountPath(), req.GetOptions())
+	options := make(map[string]string)
+	options[mountattachoptions.OptionsWaitBeforeDelete] = "true"
```
adityadani: I think by default we should set OptionsDeleteAfterUnmount to true as well and update the docs accordingly.
Also it does not make sense to set WaitBeforeDelete to true if DeleteAfterUnmount was set to false.

adityadani: Only set WaitBeforeDelete to true if DeleteAfterUnmount is set.

PR #620, opened 2018-09-19
Modify DeviceMount mounter type to handle device symbolic links.

adityadani:
A lot of devices passed to mounter could be symbolic links pointing to a different device.
For ex. in case of LVM the device passed to mounter could be /dev/mapper/lvm1 which has a symbolic link to /dev/dm-0
The underlying mount table can inter changeably use either of the device which might cause mounter to return
a device as not mounted even if it is mounted but with a different device name.

This change handles such devices by adding an entry to the mounter's mount table for the target device in the Load API.

Added UTs for the DeviceMount type of Mounter.

Closes #


pkg/mount/mount.go
```diff
@@ -334,49 +337,72 @@ func (m *Mounter) reload(device string, newM *Info) error {
...
 		}
-		mount, ok := m.mounts[sourcePath]
-		if !ok {
-			mount = &Info{
-				Device:     devicePath,
-				Fs:         v.Fstype,
-				Minor:      v.Minor,
-				Mountpoint: make([]*PathInfo, 0),
+
+		addMountTableEntry := func(mountSourcePath, deviceSourcePath string, updatePaths bool) {
```
jvinod: Move this out as a func?

PR #621, opened 2018-09-20
sdk development for alerts

sdeoras:
**What this PR does / why we need it**:
This PR is a follow up on previous work done to overhaul alerts. It specifically addresses sdk.

**Which issue(s) this PR fixes** (optional)
Closes #

**Special notes for your reviewer**:

PR #625, opened 2018-09-25
delete rpc in alerts sdk

sdeoras:

This is a follow up on #621. Adding delete rpc

Closes #621 


api/api.proto
```diff
@@ -533,9 +533,13 @@ message SdkAlertsResourceIdQuery {
 // Each query object is one of the three query types and a list of
 // options.
 message SdkAlertsQuery {
+    // query using one of the three query types.
```
lpabon: A lot of the comments start with lower case. Please change them to Upper case.

api/api.proto
```diff
@@ -558,6 +562,19 @@ message SdkAlertsEnumerateResponse {
     repeated Alert alerts = 1;
 }
 
+// SdkAlertsDeleteRequest is a request message to delete alerts.
+message SdkAlertsDeleteRequest {
+  // queries is a list of queries to find matching alerts.
+  // Output of each of these queries is added to a global pool
```
lpabon: This `Output of each of these queries is added to a global pool and returned as output of an RPC call.` is confusing. Does it mean, `Returns and deletes any alerts matching the query` ?

api/api.proto
```diff
@@ -592,7 +609,25 @@ service OpenStorageAlerts {
     rpc Enumerate(SdkAlertsEnumerateRequest)
     returns (SdkAlertsEnumerateResponse) {
         option(google.api.http) = {
-                get: "/v1/alerts"
+                get: "/v1/alerts/enumerate"
```
lpabon: This should be `/v1/alerts`. When you do a GET on this endpoint it should return alerts.

api/api.proto
```diff
@@ -592,7 +609,25 @@ service OpenStorageAlerts {
...
+    // * Query that takes resource type and alert type as input and
+    // * Query that takes resource id, alert type and resource type as input.
+    //
+    // #### Input
+    // SdkAlertsEnumerateRequest takes a list of such queries and all alerts
+    // that match at least one of the queries are deleted.
+    rpc Delete(SdkAlertsDeleteRequest)
+    returns (SdkAlertsDeleteResponse) {
+      option(google.api.http) = {
+                get: "/v1/alerts/delete"
```
lpabon: This should be `delete: "/v1/alerts"`

api/api.proto
```diff
@@ -533,9 +533,13 @@ message SdkAlertsResourceIdQuery {
 // Each query object is one of the three query types and a list of
```
lpabon: You are missing changing the version to 0.11.0 and updating the SDK_CHANGELOG.md

api/server/sdk/alerts.go
```diff
@@ -33,10 +33,10 @@ import (
 // AlertsServer implements api.OpenStorageAlertsServer.
 // In order to use this server implementation just have
 // AlertsServer pointer properly instantiated with a valid
-// alerts.Reader.
+// alerts.FilterDeleter.
 type AlertsServer struct {
-	// Reader holds pointer to alerts Reader
-	Reader alerts.Reader
+	// FilterDeleter holds pointer to alerts FilterDeleter
+	FilterDeleter alerts.FilterDeleter
```
lpabon: Please change this to non-exported.

api/server/sdk/alerts.go
```diff
@@ -158,3 +158,54 @@ func (g *AlertsServer) Enumerate(ctx context.Context,
...
+		defer cancel()
+	}
+
+	group, _ := errgroup.WithContext(ctx)
+	errChan := make(chan error)
+
+	resp := new(api.SdkAlertsDeleteResponse)
+
+	queries := request.GetQueries()
+	if queries == nil {
```
lpabon: You probably want to check this first before allocating any objects/ctx's.

api/server/sdk/alerts.go
```diff
@@ -158,3 +158,54 @@ func (g *AlertsServer) Enumerate(ctx context.Context,
...
+		return nil, status.Error(codes.InvalidArgument, "must provide a query")
```
lpabon: Use upper case: "Must provide..."

api/server/sdk/alerts.go
```diff
@@ -158,3 +158,54 @@ func (g *AlertsServer) Enumerate(ctx context.Context,
...
+	select {
+	case err := <-errChan:
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		} else {
+			return resp, nil
+		}
+	case <-ctx.Done():
+		return nil, status.Error(codes.DeadlineExceeded,
+			"deadline is reached, server side func exiting")
```
lpabon: Use uppercase

api/server/sdk/alerts.go
```diff
@@ -158,3 +158,54 @@ func (g *AlertsServer) Enumerate(ctx context.Context,
...
+	// wait for err-group processes to be done
+	go func() {
+		errChan <- group.Wait()
+	}()
+
+	// wait only as long as context deadline allows
+	select {
+	case err := <-errChan:
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
```
lpabon: Use `status.Errorf(codes.Internal, "Some explanation here: %v", err)`

lpabon: Remember that the reader of these errors most likely has no idea how OpenStorage or any of its drivers works, so the explanation should be clear.

api/server/sdk/alerts_test.go
```diff
@@ -362,9 +362,68 @@ func TestAlertsServer_Enumerate_Error(t *testing.T) {
...
-		s.MockReader().EXPECT().Enumerate(filters...).Return(myAlerts, err).Times(1)
+		s.MockFilterDeleter().EXPECT().Enumerate(filters...).Return(myAlerts, err).Times(1)
 		// Get info
 		_, outErr := c.Enumerate(context.Background(), configs[0].req)
 		assert.Error(t, outErr, err.Error())
 	}
 }
+
+// TestAlertsServer_Delete tests delete functionality over gRPC using mock.
+func TestAlertsServer_Delete(t *testing.T) {
```
lpabon: No need for `_` in the name

api/server/sdk/alerts_test.go
```diff
@@ -362,9 +362,68 @@ func TestAlertsServer_Enumerate_Error(t *testing.T) {
...
+		// Get info
+		_, err := c.Delete(context.Background(), &api.SdkAlertsDeleteRequest{
+			Queries: config.req.Queries,
+		})
+		assert.NoError(t, err)
+	}
+}
+
+// TestAlertsServer_Delete_Error tests errors returned from server code.
+func TestAlertsServer_Delete_Error(t *testing.T) {
```
lpabon: Remove `_`

api/server/sdk/alerts.go
sdeoras: ideally i would like this to be unexported, but i need to use it elsewhere (px test). One way to achieve would be to create an exported NewAlertsServer(filterDeleter alerts.FilterDeleter) func that would create an instance of interface defined in proto file.

PR #628, opened 2018-09-26
CSI 0.3.0 Snapshot Support

### LINKED ISSUE #522: csi: Support for Snapshot Create

### LINKED ISSUE #523: csi: Support for Snapshot Delete

### LINKED ISSUE #525: csi: Support for creating a volume from a snapshot using information from CreateVolume request

lpabon:
Add Snapshot support for CSI 0.3.0

Closes #525 
Closes #523 
Closes #522


csi/controller.go
```diff
@@ -455,9 +471,122 @@ func csiRequestsSharedVolume(req *csi.CreateVolumeRequest) bool {
...
+			return nil, status.Errorf(codes.Internal, "Failed to get time snapshot was created: %v", err)
+		}
+		return &csi.CreateSnapshotResponse{
+			Snapshot: &csi.Snapshot{
+				Id:             v.GetId(),
+				SourceVolumeId: v.GetSource().GetParent(),
+				CreatedAt:      createdAt.Unix(),
+				Status: &csi.SnapshotStatus{
+					// This means that we are not uploading our snapshot
+					// We may add support flow cloud snaps in future patches
```
adityadani: nit: support for cloud snaps

csi/controller.go
```diff
@@ -455,9 +471,122 @@ func csiRequestsSharedVolume(req *csi.CreateVolumeRequest) bool {
...
+					Type: csi.SnapshotStatus_READY,
+				},
+			},
+		}, nil
+	}
```
adityadani: Shouldn't we check for err != nil and see if the error is NotFound?

csi/controller.go
```diff
@@ -455,9 +471,122 @@ func csiRequestsSharedVolume(req *csi.CreateVolumeRequest) bool {
...
+
+// ListSnapshots is not supported (we can add this later)
+func (o *OsdCsiServer) ListSnapshots(
+	ctx context.Context,
+	req *csi.ListSnapshotsRequest,
+) (*csi.ListSnapshotsResponse, error) {
+
+	// The function ListSnapshots is also not published as
+	// supported by this implementation
+	return nil, status.Error(codes.Unimplemented, "ListSnashots is not implemented")
```
adityadani: nit: ListSnapshots

csi/controller.go
```diff
@@ -455,9 +471,122 @@ func csiRequestsSharedVolume(req *csi.CreateVolumeRequest) bool {
...
-	ControllerGetCapabilities(context.Context, *ControllerGetCapabilitiesRequest) (*ControllerGetCapabilitiesResponse, error)
-*/
+	if len(req.GetSourceVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume id must be provided")
+	} else if len(req.GetName()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Name must be provided")
+	}
+
+	// Check if the snapshot with this name already exists
+	v, err := util.VolumeFromName(o.driver, req.GetName())
```
harsh-px: A create call should fail if the resource already exists.

This is keeping inline with behavior of other systems like REST POSTs, kubectl creates, docker volume create...

csi/controller.go
```diff
@@ -455,9 +471,122 @@ func csiRequestsSharedVolume(req *csi.CreateVolumeRequest) bool {
...
+	}
+
+	// Get any labels passed in by the CO
+	_, locator, _, err := o.specHandler.SpecFromOpts(req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
+	}
+
+	// Create snapshot
+	readonly := true
```
harsh-px: this should be taken from `CreateSnapshotRequest`. Reading the doc for this request object, I see it already states that it is possible to specify whether snapshot is read only or read write.

If not specified, we can default to read only.

csi/controller.go
```diff
@@ -455,9 +471,122 @@ func csiRequestsSharedVolume(req *csi.CreateVolumeRequest) bool {
...
+	snapshotID, err := o.driver.Snapshot(req.GetSourceVolumeId(), readonly, &api.VolumeLocator{
+		Name:         req.GetName(),
+		VolumeLabels: locator.GetVolumeLabels(),
+	}, false)
```
harsh-px: We don't want the backend implementation to retry. 
https://github.com/libopenstorage/stork/blob/master/drivers/volume/portworx/portworx.go#L611 is how stork uses it.

The PX implementation will retry and eventually create a snapshot with same name. When the caller reinvokes the request with same name, PX will fail this with already exists error.

cc: @disrani-px

csi/controller.go
lpabon: @harsh-px CSI Spec defines that all calls are idempotent. If the create call has a name that has already been taken, then the spec specifies that the function must return the ID of that snapshot

csi/controller.go
harsh-px: It retries only if noRetry is set to false.

Since we are allowing existing snapshots, I think we are fine here.

csi/controller.go
lpabon: I think we should fix https://github.com/libopenstorage/openstorage/blob/master/pkg/util/volume.go#L38 so that it returns an exported named error value. Then we should add it here and in other areas. 

I created #631 to track this.

PR #636, opened 2018-10-03
CloudBackup API changes

talakad:
Modified Cloud Backup API to such that cloud backup and restores statuses can tracked based on the task ID rather than volume ID. This will enable multiple simultaneous backups and restores on the same volume.   
Closes #


volume/drivers/fake/fake_test.go
```diff
@@ -130,8 +130,9 @@ func TestFakeCloudBackupCreate(t *testing.T) {
 		VolumeID:       "abc",
 		CredentialUUID: "def",
 	}
-	err = d.CloudBackupCreate(req)
+	r, err := d.CloudBackupCreate(req)
```
disrani-px: On error, r should be nil

volume/drivers/fake/fake_test.go
```diff
@@ -145,9 +146,9 @@ func TestFakeCloudBackupCreate(t *testing.T) {
 	req.VolumeID = volid
 
 	// Fail because no cred id
-	err = d.CloudBackupCreate(req)
+	r, err = d.CloudBackupCreate(req)
 	assert.Error(t, err)
-
+	assert.Empty(t, r.TaskID)
```
disrani-px: Same here, r should be nil on error

volume/drivers/fake/fake_test.go
```diff
@@ -184,23 +186,24 @@ func testInitForCloudBackups(t *testing.T, d *driver) (string, *api.CloudBackupC
 		CredentialUUID: credid,
 	}
 
-	id, err := d.cloudBackupCreate(req)
+	taskId, id, err := d.cloudBackupCreate(req)
```
disrani-px: This is calling the internal method which seems wrong. Should use the interface?

volume/drivers/fake/fake_test.go
```diff
@@ -254,6 +260,11 @@ func TestFakeCloudBackupEnumerateWithoutMatches(t *testing.T) {
 	})
 	assert.NoError(t, err)
 	assert.Len(t, resp.Backups, numbackups)
+	for _, id := range taskIds {
+		_, err := d.kv.Delete(backupsKeyPrefix + "/" + id)
```
disrani-px: Call CloudBackupDelete() instead of deleting from kvdb directly?

volume/drivers/fake/fake_test.go
```diff
@@ -276,17 +290,32 @@ func TestFakeCloudBackupEnumerateMatchingVolumes(t *testing.T) {
 	})
 	assert.NoError(t, err)
 	assert.Len(t, resp.Backups, 1)
+	for _, id := range taskIds {
+		_, err := d.kv.Delete(backupsKeyPrefix + "/" + id)
```
disrani-px: Same about calling CloudBackupDelete()

volume/drivers/fake/fake_test.go
```diff
@@ -408,6 +444,11 @@ func TestFakeCloudBackupStatusWithoutMatches(t *testing.T) {
 	}
 	assert.Equal(t, 50, nbackups)
 	assert.Equal(t, 50, nrestores)
+
+	for _, id := range taskIds {
+		_, err := d.kv.Delete(backupsKeyPrefix + "/" + id)
```
disrani-px: Same comment as above

volume/drivers/fake/fake_test.go
```diff
@@ -523,14 +564,14 @@ func TestFakeCloudBackupStateChange(t *testing.T) {
 	d, err := newFakeDriver(map[string]string{})
 	assert.NoError(t, err)
 
-	backupId, _, vol := testInitForCloudBackups(t, d)
+	taskId, _, _, vol := testInitForCloudBackups(t, d)
 
 	// Update element on db
 	var elem *fakeBackups
-	_, err = d.kv.GetVal(backupsKeyPrefix+"/"+backupId, &elem)
+	_, err = d.kv.GetVal(backupsKeyPrefix+"/"+taskId, &elem)
```
disrani-px: Not related to your change but these tests seem weird/wrong. They shouldn't be modifying kvdb directly. There should be hooks in the fake driver for doing it.

api/api.proto
```diff
@@ -2163,24 +2171,30 @@ message SdkCloudBackupStatus {
...
 }
 
 // Defines a request to retreive the status of a backup or restore for a
 // specified volume
 message SdkCloudBackupStatusRequest {
   // This is an optional value which is used to get information on the
-  // status of a backup for the specified volume. If no volume id is provided,
-  // then status for all volumes is returned.
+  // status of a backup for the specified volume. If no volume id and TaskId
+  // is provided, then status for all volumes is returned.
```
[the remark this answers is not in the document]
talakad: volumeId is ignored if both are provided. Thats mentioned in comments on taskID field.

api/server/backup.go
```diff
@@ -242,7 +242,7 @@ func (vd *volAPI) cloudBackupStateChange(w http.ResponseWriter, r *http.Request)
 
 	err = d.CloudBackupStateChange(stateChangeReq)
 	if err != nil {
-		vd.sendError(method, stateChangeReq.SrcVolumeID, w, err.Error(), http.StatusInternalServerError)
+		vd.sendError(method, "", w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
talakad: It could. However everywhere else, volume ID is used for logging purposes with ID prefix and this could clobber with those debugs. Hence did not add.

volume/drivers/fake/fake.go
```diff
@@ -410,40 +410,44 @@ func (d *driver) CredsValidate(uuid string) error {
...
+	taskId, _, err := d.cloudBackupCreate(input)
+	resp := &api.CloudBackupCreateResponse{TaskID: taskId}
+	return resp, err
 }
 
 // cloudBackupCreate uploads snapshot of a volume to the cloud and returns the
-// backup id
-func (d *driver) cloudBackupCreate(input *api.CloudBackupCreateRequest) (string, error) {
+// backup task id
+func (d *driver) cloudBackupCreate(input *api.CloudBackupCreateRequest) (string, string, error) {
```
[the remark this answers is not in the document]
talakad: fake driver uses this ID for other purposes.. can't remove

volume/drivers/fake/fake_test.go
talakad: Yep. will fix.

volume/drivers/fake/fake_test.go
talakad: This is fake driver setting things up.

volume/drivers/fake/fake_test.go
talakad: This is not deleting cloudsnaps, rather cleaning up kvdb that is  used by fake driver.

api/api.proto
```diff
@@ -2014,10 +2014,14 @@ message SdkCloudBackupCreateRequest {
   string credential_id = 2;
   // Full indicates if full backup is desired even though incremental is possible
   bool full = 3;
+  // TaskId is the optional unique id of the task performing this backup
```
lpabon: What is the purpose of providing this id? Is it returned in the responses? (it should be documented here) Should the caller provide the id or is the server going to generate one? 

Seems weird to request the id from the caller.

api/api.proto
```diff
@@ -2014,10 +2014,14 @@ message SdkCloudBackupCreateRequest {
...
+  string task_id = 4;
 }
 
 // Empty response
 message SdkCloudBackupCreateResponse {
+	// TaskId is the id of the task performing the backup
```
lpabon: tabs are wrong. Check out the top of the file for VIM settings.

api/api.proto
```diff
@@ -2163,24 +2171,32 @@ message SdkCloudBackupStatus {
   google.protobuf.Timestamp completed_time = 6;
   // NodeID is the ID of the node where this Op is active
   string node_id = 7;
+  // SourceVolumeID is the the volume that is either being backed up to cloud
+  // or target volume to which cloudsnap is being restored
```
lpabon: What is cloudsnap? I think that is a Px word, but what is the similar technology in OpenStorage?

api/api.proto
```diff
@@ -2163,24 +2171,32 @@ message SdkCloudBackupStatus {
...
+  string src_volume_id = 8;
+  // ErrString indicates the failure cause for failed backup/restore
```
lpabon: Should this be `info`? We could use this for good status information also.

api/api.proto
```diff
@@ -2163,24 +2171,32 @@ message SdkCloudBackupStatus {
...
-  // status of a backup for the specified volume. If no volume id is provided,
-  // then status for all volumes is returned.
+  // status of a backup for the specified volume. If no volume id and TaskId
+  // is provided, then status for all volumes is returned.
   string volume_id = 1;
   // Local indicates if only those backups/restores that are
   // active on current node must be returned
   bool local = 2;
+  // TaskId is the backup/restore task id, if this is specified,
+  // VolumeId is ignored.
```
lpabon: this should be `volume_id`

api/api.proto
```diff
@@ -2163,24 +2171,32 @@ message SdkCloudBackupStatus {
...
+  string task_id = 3;
 }
 
 // Defines a response containing the status of the backups for a specified volume
 message SdkCloudBackupStatusResponse {
   // Statuses is list of currently active/failed/done backup/restores where
-  // the key is the volume ids of current volumes being backed up or restored
+  // the key is the id of the task performing backup/restore.
```
lpabon: Isn't task_id optional? What happens when it not provided, what is the key?

api/server/sdk/cloud_backup.go
```diff
@@ -44,15 +44,19 @@ func (s *CloudBackupServer) Create(
...
-	}); err != nil {
+		TaskID:         req.GetTaskId(),
+	})
+	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to create backup: %v", err)
 	}
 
-	return &api.SdkCloudBackupCreateResponse{}, nil
+	return &api.SdkCloudBackupCreateResponse{
+		TaskId: r.TaskID,
```
lpabon: This is really confusing. there is a task id from the caller and a task id from CloudBackupCreate. I do not think the caller should provide an id. Is it necessary for a use case?

api/server/sdk/cloud_backup.go
```diff
@@ -222,7 +228,7 @@ func (s *CloudBackupServer) StateChange(
 	req *api.SdkCloudBackupStateChangeRequest,
 ) (*api.SdkCloudBackupStateChangeResponse, error) {
 
-	if len(req.GetSrcVolumeId()) == 0 {
+	if len(req.GetTaskId()) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Must provide volume id")
```
lpabon: Wrong error message.

lpabon: Is taskid required here?

api/server/sdk/cloud_backup.go
talakad: Hope above comments are helpful and no more confusion. Again, callers do not have to provide it.  They can, if they want this call to be idempotent.

api/api.proto
talakad: Key is the task ID as the status response structure describes.  both task id and volume-id are optional, in which case all existing tasks will be returned. I will add it to the comments.

api/api.proto
talakad: Okay.

api/api.proto
talakad: you mean ErrInfo? this is expected to be non-empty only if the status says failed as the comment indicates.

api/api.proto
talakad: other fields are not that way..

api/api.proto
talakad: Okay.

api/api.proto
talakad: This is to allow callers to specify taskID so that the call can idempotent.  If task ID provided already exists , success is returned, if not it will be created and success is returned.

api/api.proto
lpabon: I mean just `info`. I would have suggested `status` but that is taken already. In a way I feel this is an object like in Kubernetes, where they have status, and users can read what is happening to the object when the check the status (which in kubernetes is a `[]string`).

I'm suggesting that this could be called `repeated string info`, where the server can continue adding information to the object and the caller can get a sequence of what happened (like a log).

I'm not 100% on this, just suggesting. 

If you do not think this is a good idea, then at least I would suggest changing the name to `error_details` instead of string.

api/api.proto
lpabon: In Golang it is called VolumeId but in other languages and in the proto file it is `volume_id`. In the documentation generated with this proto file, we should use proto names as links.

api/server/sdk/cloud_backup.go
lpabon: I think we should force them to provide it. We do this in Volume Create and in Snapshot Create. We should force them here also. Also in both of those it is called `name`. Suggesting that we call it `name` of the backup here also to be consistent with other creations.

api/server/sdk/cloud_backup.go
lpabon: I'm good with this now that I understand what it is for. 👍

SDK_CHANGELOG.md
```diff
@@ -5,6 +5,16 @@
 
 ## Releases
 
+### v0.18.0 - Tech Preview (10/23/2018)
+
+* Following CloudBackup APIs have been refactored to include task id rather than
+* source volume id.
```
lpabon: There is a bullet in this one, but it is a continuation of the previous sentence.

api/api.proto
```diff
@@ -2058,10 +2058,14 @@ message SdkCloudBackupCreateRequest {
   string credential_id = 2;
   // Full indicates if full backup is desired even though incremental is possible
   bool full = 3;
+  // Name of the task performing this backup
```
lpabon: Probably need to add `...this backup. This value is required and used for idempotency.`

api/api.proto
```diff
@@ -2207,24 +2215,32 @@ message SdkCloudBackupStatus {
...
+  // VolumeId is an optional value which is used to get information on the
+  // status of a backup for the specified volume. If no volume id and Name
+  // is provided, then status for all volumes is returned.
   string volume_id = 1;
   // Local indicates if only those backups/restores that are
   // active on current node must be returned
   bool local = 2;
+  // Name of the backup/restore task, if this is specified,
+  // volume_id is ignored.
+  string Name = 3;
```
lpabon: needs to be `name` not `Name`

api/api.proto
```diff
@@ -2207,24 +2215,32 @@ message SdkCloudBackupStatus {
...
 }
 
 // Defines a request to retreive the status of a backup or restore for a
 // specified volume
 message SdkCloudBackupStatusRequest {
-  // This is an optional value which is used to get information on the
-  // status of a backup for the specified volume. If no volume id is provided,
-  // then status for all volumes is returned.
+  // VolumeId is an optional value which is used to get information on the
+  // status of a backup for the specified volume. If no volume id and Name
```
lpabon: Is this supposed to be `If no volume id or name ...` ?

api/api.proto
```diff
@@ -2283,9 +2299,9 @@ enum SdkCloudBackupRequestedState {
 // Defines a request to change the state of a backup or restore to or
 // from a cloud provider
 message SdkCloudBackupStateChangeRequest {
-  // Describes the volume id on which backup/restore
+  // Describes the backup/restore Name
   // state change is being requested
-  string src_volume_id = 1;
+  string Name = 1;
```
lpabon: `name`

api/server/sdk/cloud_backup.go
```diff
@@ -37,22 +37,28 @@ func (s *CloudBackupServer) Create(
...
+		return nil, status.Error(codes.InvalidArgument, "Must supply a name")
+	} else if len(req.GetVolumeId()) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Must supply a volume id")
 	} else if len(req.GetCredentialId()) == 0 {
 		return nil, status.Error(codes.InvalidArgument, "Must supply credential uuid")
 	}
 
 	// Create the backup
-	if err := s.driver.CloudBackupCreate(&api.CloudBackupCreateRequest{
+	r, err := s.driver.CloudBackupCreate(&api.CloudBackupCreateRequest{
```
lpabon: Is idempotency checked by this function?

talakad: yes

SDK_CHANGELOG.md
talakad: yep. will fix that.

api/api.proto
talakad: ok.

api/api.proto
talakad: ok.

api/api.proto
talakad: yeah, name.

api/api.proto
talakad: it should be name.

PR #645, opened 2018-10-11
PWX-6567: Add discard option tp volume spec

prabirpaul:

Adds discard option to volume spec. It is used at volume mount to enabled/disable discard support.

Closes # PWX-6567


api/api.proto
```diff
@@ -313,6 +313,8 @@ message VolumeSpec {
   // Use to force a file system type which is not recommended.
   // The driver may still refuse to use the file system type.
   bool force_unsupported_fs_type = 28;
+  // Discard is true if the volume will be mounted with discard support.
```
lpabon: Question on this comment. Does that mean that `discard` is set by the system? If it is true if the system decided to support discard? Or is the idea to have the caller set discard to true?

prabirpaul: Currently all volumes get mounted with discard option turned on. We are going to make this a volume create option, so the user can turn it off if needed. Here's the relevant pr:
https://github.com/portworx/porx/pull/2342

PR #649, opened 2018-10-12
PXW-6571 Added basic stubs for the API's we need to implement in the …

tapas-sharma:
Added basic stubs for the API's we need to implement in the SDK, made appropriate changes in the api.proto file, tested by using make launch-sdk. No business logic or tests have been added till now

PXW-6571
Closes # PXW-6571

This is the first commit having the basic stubs, will be adding the test and business logic shortly.

### DIFF
--- a/api/api.proto
+++ b/api/api.proto
@@ -932,6 +932,50 @@ service OpenStorageCluster {
         get: "/v1/clusters/current"
       };
     }
+    // Creates Pair with a remote cluster and returns details about the remote cluster
+    rpc CreatePair(ClusterPairCreateRequest)
+    returns (ClusterPairCreateResponse){
+      option(google.api.http) = {
+        post: "/v1/clusters/pair"
+        body: "*"
+      };
+    }
+    // Process incoming request from a remote cluster
+    rpc ProcessPairRequest(ClusterPairProcessRequest)
+    returns (ClusterPairProcessResponse){
+      option(google.api.http) = {
+        post: "/v1/clusters/pair/process"
+        body: "*"
+      };
+    }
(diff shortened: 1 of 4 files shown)


api/api.proto
```diff
@@ -932,6 +932,50 @@ service OpenStorageCluster {
         get: "/v1/clusters/current"
       };
     }
+    // Creates Pair with a remote cluster and returns details about the remote cluster
```
lpabon: Please create a new service called `OpenStorageClusterPair` to group these functions, that way it will eliminate the name collision problem.

lpabon: Please add adequate documentation for non-openstorage users to understand.

api/api.proto
```diff
@@ -932,6 +932,50 @@ service OpenStorageCluster {
...
+    rpc CreatePair(ClusterPairCreateRequest)
```
lpabon: I suggest creating a single issue with the API proposal and then create a PR per API function.

api/api.proto
```diff
@@ -1427,6 +1471,34 @@ service OpenStorageCloudBackup {
     }
 }
 
+// Provides Cloud migration features
+service OpenStorageCloudMigrate {
```
lpabon: Probably call this 'VolumeMigration' since OpenStorage does not migrate clouds.

api/api.proto
```diff
@@ -2575,6 +2660,10 @@ message ClusterPairDeleteRequest {
     string cluster_id = 1;
 }
 
+// Response to delete a cluster pair
+// swagger:model
```
lpabon: Please remove.

PR #665, opened 2018-10-19
Add volume placement specs

harsh-px:

**What this PR does / why we need it**: This adds support for providing a volume placement spec in the volume create request.  Examples on how this will be used in Kubernetes are at: https://github.com/harsh-px/k8s-utils/tree/master/scripts/volume-placement

The spec has been influenced from Kubernetes pod affinities and anti-affinities: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity


api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+
+message VolumePlacementSpec {
+  // Defines a list of rules as part of the placement spec
+  repeated VolumePlacementRule rules = 1;
+}
+
+message VolumePlacementRule {
+  // Defines the number of volume replicas affected by this rule. If not provided,
+  // rule would affect all replicas
+  // +optional
```
lpabon: When the documentation is create for the SDK website, it takes the comments as markdown, so it would show up as `replicas +optional`. To show something is optional, use `(optional)` at the beginning of the comment.

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+  uint64 affectedReplicas = 1;
```
lpabon: use `int64`.  CSI had to change all of its values to int64 from uin64 at v0.2, More info here: https://github.com/container-storage-interface/spec/issues/172

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+  // Defines the affinity rule for the volume placement
+  VolumePlacementAffinity affinity = 2;
+}
+
+// VolumePlacementAffinity allows to specify affinity and anti-affinity rules based on
```
lpabon: Change `allows to specify` to `specifies`

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+// storage pools, nodes and volumes.
+message VolumePlacementAffinity {
+  // Defines the storage pool affinity for volume placement
+  AffinityRule poolAffinity = 1;
```
lpabon: This is an example, but all names should not use camel case. Instead they should use `_` and the Golang generator changes them to camel case.

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+  // rule would affect all replicas
+  // +optional
+  uint64 affectedReplicas = 1;
+  // Defines the affinity rule for the volume placement
+  VolumePlacementAffinity affinity = 2;
+}
+
+// VolumePlacementAffinity allows to specify affinity and anti-affinity rules based on
+// storage pools, nodes and volumes.
+message VolumePlacementAffinity {
```
lpabon: Q: Can all these settings be set at the same time? If any of these settings cannot/should not be set together they should be in a `oneof`

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+  AffinityRule nodeAntiAffinity = 4;
+
+  // Defines the volume affinity for volume placement
+  AffinityRule volumeAffinity = 5;
+  // Defines the volume anti-affinity for volume placement
+  AffinityRule volumeAntiAffinity = 6;
+}
+
+message AffinityRule {
+  // If false, the affinity rule is best-effort/preferred rather than enforced
```
lpabon: I think this is supposed to be if `true`, right?

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+
+
+// A label selector is a label query over a set of resources. The result of matchLabels and
+// matchExpressions are ANDed. An empty label selector matches all objects. A null
+// label selector matches no objects.
+message LabelSelector {
+  // matchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
+  // map is equivalent to an element of matchExpressions, whose key field is "key", the
+  // operator is "In", and the values array contains only "value". The requirements are ANDed.
+  map<string, string> matchLabels = 1;
```
lpabon: this can be just `labels`

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+  // If false, the affinity rule is best-effort/preferred rather than enforced
+  bool bestEffort = 1;
+  // A label query over storage pools to select
+  LabelSelector labelSelector = 2;
+}
+
+
+// A label selector is a label query over a set of resources. The result of matchLabels and
+// matchExpressions are ANDed. An empty label selector matches all objects. A null
+// label selector matches no objects.
```
lpabon: This is confusing because a nil and empty selector could be the same from the developer point of view. I think `nil` should act as an empty selector which matches all objects.

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+  // +optional
+  uint64 affectedReplicas = 1;
+  // Defines the affinity rule for the volume placement
+  VolumePlacementAffinity affinity = 2;
+}
+
+// VolumePlacementAffinity allows to specify affinity and anti-affinity rules based on
+// storage pools, nodes and volumes.
+message VolumePlacementAffinity {
+  // Defines the storage pool affinity for volume placement
```
lpabon: Do you mind adding some more text to each of these comments. something like `Use this value when...`

api/api.proto
```diff
@@ -2684,3 +2686,74 @@ message LocateResponse {
...
+
+// A label selector requirement is a selector that contains values, a key, and an operator that
+// relates the key and values.
+message LabelSelectorRequirement {
+  // key is the label key that the selector applies to.
+  string key = 1;
+
+  // operator represents a key's relationship to a set of values.
+  // Valid operators are In, NotIn, Exists and DoesNotExist.
+  string operator = 2;
```
lpabon: This should be an enum. No string command in gRPC/protobuf :) . Add as an enum embedded in this message

api/api.proto
```diff
@@ -179,6 +179,18 @@ enum OperationFlags {
   OP_FLAGS_DETACH_FORCE = 2;
 }
 
+enum VolumePlacementEnforcementType {
```
lpabon: Need comments for the enum type and its values.

lpabon: Also move this to `VolumePlacementRule` so that the scope namespace is set automatically by the generator.

api/api.proto
```diff
@@ -179,6 +179,18 @@ enum OperationFlags {
...
+  Required = 0;
+  Preferred = 1;
+}
+
+enum LabelSelectorOperator {
```
lpabon: Move this to `LabelSelectorRequirement` to have the scope and namespace automatically setup.

api/api.proto
```diff
@@ -2684,3 +2698,53 @@ message LocateResponse {
     // <containerid>: /var/www
     map<string, string> dockerids = 2;
 }
+
+message VolumePlacementSpec {
+  // Defines a list of rules as part of the placement spec
+  repeated VolumePlacementRule rules = 1;
+}
+
+message VolumePlacementRule {
```
lpabon: The messages themselves need at least a small brief comment. This one may want to explain what happens if you combine different optional values. If you cannot combine them, then they should go into a `oneof`.

api/api.proto
```diff
@@ -2684,3 +2698,53 @@ message LocateResponse {
     // <containerid>: /var/www
     map<string, string> dockerids = 2;
 }
+
+message VolumePlacementSpec {
+  // Defines a list of rules as part of the placement spec
```
lpabon: Does that mean that every rule must successfully be applied? Should add comments here on what happens if many rules are applied and if one conflicts with others.

api/api.proto
```diff
@@ -2684,3 +2698,53 @@ message LocateResponse {
...
+
+message AffinityRule {
+  // This is a list of label selector requirements. The requirements are ANDed.
+  repeated LabelSelectorRequirement match_expressions = 1;
+}
+
+
+// A label selector is a label query over a set of resources. The result of match_labels and
+// match_expressions are ANDed. A nil or empty label selector matches all objects.
+message LabelSelector {
```
lpabon: Should this a `LabelSelectorAny` ? Or is it missing value? or is this supposed to be an enum?

api/api.proto
```diff
@@ -2684,3 +2698,53 @@ message LocateResponse {
...
+message VolumePlacementSpec {
+  // Defines a list of rules as part of the placement spec
+  repeated VolumePlacementRule rules = 1;
+}
+
+message VolumePlacementRule {
+  // Defines the number of volume replicas affected by this rule. If not provided,
+  // rule would affect all replicas
+  // (optional)
+  int64 affected_replicas = 1;
```
lpabon: do we really need int64 for all the replicas? It should probably be int32, right?

api/api.proto
```diff
@@ -2684,3 +2698,53 @@ message LocateResponse {
...
+  // Defines the weight of the rule which allows to break to tie with other matching rules
+  // (optional)
```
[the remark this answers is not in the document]
harsh-px: @adityadani `optional` isn't supported in proto3. https://developers.google.com/protocol-buffers/docs/proto3#default

It was only in proto2.

api/api.proto
harsh-px: Done

api/api.proto
harsh-px: Done

api/api.proto
harsh-px: Added comment above here

```
message VolumePlacementStrategy {
  // Defines a list of rules as part of the placement spec. All the rules specified will
  // be applied for volume placement.
  // Rules that have enforcement as "required" are strictly enforced while "preferred" are best effort.
  // In situations, where 2 or more rules conflict, the weight of the rules will dictate which wins.
  repeated VolumePlacementRule rules = 1;
}
```

harsh-px: It's not a oneof since we have use cases to have both. Also I want the spec to allow multiple rules. An implementation can fail the request when parsing the strategy if it doesn't support it.

api/api.proto
harsh-px: This was a remnant of the previous change. `LabelSelector` is no longer needed. Removed it.

api/api.proto
harsh-px: Added comment under `message VolumePlacementStrategy`.

api/api.proto
harsh-px: Changed to int32.

api/api.proto
```diff
@@ -2694,3 +2698,72 @@ message LocateResponse {
...
+    Required = 0;
+    // This specifies that the rule is preferred and can be best effort
+    Preferred = 1;
+  }
+
+  // Specifies the rule enforcement policy. Can take values: required or preferred.
+  // (optional)
+  EnforcementType enforcement = 3;
+  // Defines the affinity rule for the volume placement
+  AffinityRule affinity = 4;
```
sangleganesh: the anti/affinity should be a part of the rule itself instead of having two of them in a rule ?

harsh-px: I discussed this with vinod offline.

We have uses cases where we both both as part of the same rule. For e.g

So below, one of the replicas need to honor both the `affinity` and `antiAffinity` rule.

We can't add that as part of the 2nd rule since that will affect other replicas.

```
apiVersion: libopenstorage.org/v1alpha1
kind: VolumePlacementStrategy
metadata:
  name: data-collocate-with-log-placement
spec:
  rules:
    - affectedReplicas: 1 
      enforcement: required
      affinity:
        matchExpressions:
        - key: type
          operator: In
          values:
          - log
      antiAffinity:
        matchExpressions:
        - key: type
          operator: In
          values:
          - data
    - enforcement: preferred
      antiAffinity:
        matchExpressions:
        - key: media_type
          operator: NotIn
          values:
          - "SSD"
```

sangleganesh: it won't work for a third rule, because the rule has "1" affinity part and "1" anti-affinity part, because we now have rules within rules.

PR #668, opened 2018-10-19
Set Default Retain Values to Schedule Policy Object depending on the type

ipochi:
Set Default Retain Values to Schedule Policy Object depending on the type
So if the golang default values or Zero is passed by the user
they are converted internally to the default values specifed in pxctl

Closes #

### DIFF
--- a/api/server/sdk/utils.go
+++ b/api/server/sdk/utils.go
@@ -84,10 +84,41 @@ func sdkSchedToRetainInternalSpec(
 	req *api.SdkSchedulePolicyInterval,
 ) (*sched.RetainIntervalSpec, error) {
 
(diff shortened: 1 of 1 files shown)


lpabon:
@ipochi Travis is failing because you passed the limit of the git brief message. I think max is either supposed to be 50 or 72. I think our Travis checks are lenient and check if you pass 90 chars. You will need to amend the commit message and make the commit top message less than 72 chars.

lpabon:
@ipochi I think this is supposed to check only for 0. Let the implementation then check the code you have added here. The SDK should not check such details, imho.

ggriffiths:

PR #673, opened 2018-10-22
Add io_strategy - specification for I/O characteristics

jvinod:

Add io_strategy - specification for I/O characteristics


api/api.proto
```diff
@@ -257,6 +257,14 @@ message Group {
   string id = 1;
 }
 
+// IoStrategy defines how I/O should be performed to backing storage media.
+message IoStrategy {
```
adityadani: nit: Should we rename this to AsyncIoStrategy?
Also we can be explicit indicating what the field does by renaming it to `enable_kaio` because ultimately this will end up in a storage class.

PR #678, opened 2018-10-23
Call the ClusterListener APIs asynchronously

adityadani:
- In the updateClusterStatus routine we continuously post the Node updates to all the cluster
  listeners. However they are being done synchronously .
- As in this routine the error from the Listener APIs is not processed, they can be called asynchronously.

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -22,6 +22,8 @@ import (
 	"github.com/libopenstorage/openstorage/config"
 	"github.com/libopenstorage/openstorage/objectstore"
 	"github.com/libopenstorage/openstorage/osdconfig"
+	"github.com/libopenstorage/openstorage/pkg/pcqueue"
+	jobsched "github.com/libopenstorage/openstorage/pkg/sched"
 	sched "github.com/libopenstorage/openstorage/schedpolicy"
 	"github.com/libopenstorage/openstorage/secrets"
 	"github.com/libopenstorage/systemutils"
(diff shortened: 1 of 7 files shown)


sangleganesh:
with this change, more than one update can run in parallel, right ? you probably want to queue them up and process them in order, asynchronously (pcqueue with size = 1?)

adityadani:
@sangleganesh I have updated the PR.

cluster/manager/manager.go
```diff
@@ -84,14 +87,18 @@ func Init(cfg config.ClusterConfig) error {
...
 	}
-
 	inst = &ClusterManager{
 		listeners:    list.New(),
 		config:       cfg,
 		kv:           kv,
 		nodeCache:    make(map[string]api.Node),
 		nodeStatuses: make(map[string]api.Status),
 	}
+	if jobsched.Instance() == nil {
```
sangleganesh: nit: not needed anymore

pkg/pcqueue/pcqueue.go
```diff
@@ -0,0 +1,68 @@
...
+		cv:    sync.NewCond(mtx),
+		taskQ: list.New()}
+	workerFn := func() { q.dequeue() }
+	for i := uint8(0); i < numWorkers; i++ {
+		go workerFn()
+	}
+	return q
+}
+
+func (w *taskQueue) dequeue() {
```
lpabon: Why do this and not use channels? Is there a benefit?

PR #679, opened 2018-10-23
Update migrate APIs to use taskId

disrani-px:
Starting a migrate task will return a taskId. Cancel can be called on the task.
If a task was started on a group, cancelling with the taskId will cancel all the
migrations in that group.
Also return task Id in status.

Closes #


lpabon:
LGTM but I'll defer to @talakad and Tapas

PR #682, opened 2018-10-25
SDK can be started without objects

### LINKED ISSUE #677: SDK should be started without objects, then added later
BUG

**What happened**:
The SDK currently requires cluster and volume objects to be available before starting the SDK. 

The SDK should be able to be started without objects, then cluster and volume objects can be added once. When the SDK is started without objects, it should return the appropriate helpful error to the caller when a function is called which requires that object.

lpabon:
This provides a way for the SDK to be started without cluster,
volume driver, or alerts. The respective handlers should check
if their necessary objects are available before continuing.

An object can be added to the SDK while running. For this to
work, the SDK has a rwlock interceptor, which is used to wait
until all requests are complete before adjust the object handlers.

Part of #677 

I tried to make sure during tests that we can catch any non-handled checks, but I could not figure out how to do this automatically. We will need to make sure as we add new handlers to check for objects, specially during reviews.


lpabon:
This change needs to go in before #692

api/server/sdk/server.go
```diff
@@ -58,7 +63,17 @@ type ServerConfig struct {
 type Server struct {
 	*grpcserver.GrpcServer
 
-	restPort             string
+	restPort            string
+	onceSetCluster      sync.Once
```
adityadani: sync.Once not used anywhere

api/server/sdk/server.go
```diff
@@ -304,3 +377,16 @@ func (s *Server) restServerSetupHandlers() (*http.ServeMux, error) {
 
 	return mux, nil
 }
+
+// This interceptor provides a way to lock out any calls while we adjust the server
+func (s *Server) rwlockIntercepter(
```
adityadani: This is great!

api/server/sdk/server.go
lpabon: oh, nice catch, i was trying different algorithms. I will remove it

Issue #670, opened 2018-10-22
Kubernetes-csi example driver


lpabon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE

**What happened**:
As part of the [Kubernetes-CSI](https://github.com/kubernetes-csi) team, we are looking for a CSI driver that can support provisioning and snapshot of volumes and that it is able to disconnect the volume from one server and attach to another. The driver is should also support _shared_ and _block_ volumes to be able to thoroughly test the Kubernetes CSI implementation.

I would like to suggest enhancing the NFS driver in OpenStorage to support:

* Block (using loopback devices)
* Shared (normal NFS volumes)
* Provisioning
* Snapshots

Most of this work is being done one the branch [lpabon:nfs-csi](https://github.com/libopenstorage/openstorage/compare/master...lpabon:nfs-csi) branch.

Once the driver is completed, I would also need the appropriate _yaml_ spec to be able to deploy. To fully deploy, it needs:

* an etcd instance (1 is fine)
* an nfs server. Something like https://github.com/lpabon/nfs-ganesha-kubernetes
* StorageClasses, 1 for shared, 1 for block
* The yaml for this driver

PR #690, opened 2018-10-29
Update OpenStorage Readme

lpabon:
Need to update the openstorage readme now that we have the OpenStorage SDK and the NFS CSI Sample driver in #670 is coming.


gourao:
This looks good to me

Issue #653, opened 2018-10-15
Add pairing and cloud migration APIs to SDK


sharma-tapas:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST

**What happened**:
Currently, the SDK does not have support for 
1. Cluster Pairing
2. Volume Cloud Migration
This PR should be adding the following API's

**- Cluster Pairing**
- [ ] Create : Create pair with a remote cluster
- [ ] Process : handles an incoming pair request from a remote cluster
- [ ] Get : returns pair information for a cluster
- [ ] Enumerate : returns list of cluster pairs
- [ ] Delete : Delete a cluster pairing
- [ ] GetPairToken : gets the authentication token for this cluster

**- Volume Cloud Migration**
- [ ] Start : starts a migrate operation
- [ ] Cancel : cancels/stops a migrate operation
- [ ] Status : status of a migrate operation

We will be creating a new service for the cluster pairing called OpenStorageClusterPair and for volume cloud migration as OpenStorageVolumeCloudMigrate which will house the above.

PR #692, opened 2018-10-30
Cluster pairing and volume migration

sharma-tapas:
This PR adds the ClusterPairing and VolumeMigration API's into the SDK.

Part of #653 

Ignore other PR's this is the squashed one, will be closing the old PR's.


api/api.proto
```diff
@@ -2518,7 +2613,7 @@ message CloudMigrateStartRequest {
 }
 
 // Response to start a cloud migration
-message CloudMigrateStartResponse {
+message SdkCloudMigrateStartResponse {
```
lpabon: Structure like this should instead be:

```
message SdkCloudMigrateStartRespone {
    CloudMigrateStartResponse response = 1;
}
```
So that we do not break the old REST interface. And consequently break drivers (like Portworx) that use these APIs

api/api.proto
```diff
@@ -2556,6 +2655,10 @@ message CloudMigrateInfoList {
     repeated CloudMigrateInfo list = 1;
 }
 
+// Empty Request
+message SdkCloudMigrateStatusRequest {
+
```
lpabon: empty line

api/api.proto
```diff
@@ -2670,6 +2777,10 @@ message ClusterPairGetResponse {
     ClusterPairInfo pair_info = 1;
 }
 
+// Empty Request
+message SdkClusterPairsEnumerateRequest{
+}
+
 // Response to enumerate all the cluster pairs
 // swagger:response
```
lpabon: remove this swagger stuff since this structure will now be used by the SDK.

api/api.proto
```diff
@@ -2684,6 +2795,7 @@ message ClusterPairsEnumerateResponse {
     map<string, ClusterPairInfo> pairs = 2;
```
lpabon: What is the key? Ask @disrani-px

api/api.proto
```diff
@@ -2684,6 +2795,7 @@ message ClusterPairsEnumerateResponse {
...
 }
 
+
```
lpabon: remove

api/server/sdk/cluster_pair.go
```diff
@@ -0,0 +1,134 @@
...
+	} else if len(req.GetRemoteClusterToken()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply Authentication Token")
+	} else if req.GetRemoteClusterPort() == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply Remote cluster Port")
+	}
+
+	resp, err := s.cluster.CreatePair(req)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Cannot create cluster with remote pair %s : %v",
+			req.GetRemoteClusterIp(), err.Error())
```
lpabon: no need to use `err.Error()` for `%v`, just use `err`. Fix in all occurrences.

api/server/sdk/cluster_pair.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return resp, nil
+}
+
+// Get information about a cluster pair
+func (s *ClusterPairServer) Get(
+	ctx context.Context,
+	req *api.ClusterPairGetRequest,
+) (*api.ClusterPairGetResponse, error) {
+	name := req.GetId()
```
lpabon: why do this? Just use req.GetId() instead.

api/server/sdk/volume_migrate.go
```diff
@@ -0,0 +1,78 @@
...
+	} else if len(req.GetTargetId()) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "Must supply valid Target cluster ID")
+	}
+	resp, err := s.driver.CloudMigrateStart(req)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Cannot start migration for %s : %v", req.GetClusterId(), err.Error())
+	}
+	retVal := &api.SdkCloudMigrateStartResponse{}
+	if resp != nil {
+		retVal.TaskId = resp.GetTaskId()
```
lpabon: We need to trust that if it is not an err, resp is not nil. That would be bad.

Do not use the style in the code, instead use:

```go
return &api.SdkCloudMigrateStartResponse{
    TaskId: resp.GetTaskId(),
}, nil
```

api/client/volume/client.go
```diff
@@ -735,8 +735,8 @@ func (v *volumeClient) SnapshotGroup(groupID string, labels map[string]string) (
 	return response, nil
 }
 
-func (v *volumeClient) CloudMigrateStart(request *api.CloudMigrateStartRequest) (*api.CloudMigrateStartResponse, error) {
-	startResponse := &api.CloudMigrateStartResponse{}
+func (v *volumeClient) CloudMigrateStart(request *api.CloudMigrateStartRequest) (*api.SdkCloudMigrateStartResponse, error) {
+	startResponse := &api.SdkCloudMigrateStartResponse{}
```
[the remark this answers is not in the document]
disrani-px: Yeah, don't change the name of the structures being used by the REST API

api/server/sdk/cluster_pair.go
```diff
@@ -0,0 +1,134 @@
...
+	cluster cluster.Cluster
+}
+
+// Create a new cluster with remote pair
+func (s *ClusterPairServer) Create(
+	ctx context.Context,
+	req *api.ClusterPairCreateRequest,
+) (*api.ClusterPairCreateResponse, error) {
+
+	if len(req.GetRemoteClusterIp()) == 0 {
```
disrani-px: These checks aren't required here. This should just be a pass through. The implementation will do the checks. Same for APIs below

lpabon:
Also the spacing in `api.proto` is incorrect. Take a look at the top of the file for spacing instructions.

api/api.proto
```diff
@@ -951,6 +951,74 @@ service OpenStorageCluster {
...
+    // ProcessRequest handles a remote cluster's pair request
+    //
+    // ##### Example
+    // {% codetabs name="Golang", type="go" -%}
+    // id, err := client.Create(context.Background(), &api.ClusterPairProcessRequest{
+    //   SourceClusterId: "ID",
+    //   RemoteClusterToken: "<Auth-Token>",
+    // })
+    // {%- endcodetabs %}
+    rpc ProcessRequest(ClusterPairProcessRequest)
```
disrani-px: This will only be used between 2 clusters, doesn't need to be exposed via SDK

api/server/sdk/cluster_pair.go
sharma-tapas: This reduces the function call from 2 to 1, we can use the "name" var for both length check and pass it to the client API call. We can call the function twice for sure, but though to save a few function call here.

api/server/sdk/volume_migrate.go
sharma-tapas: Changed it, forgot that Go will anyhow make the var's in the struct null.

api/api.proto
```diff
@@ -951,6 +951,74 @@ service OpenStorageCluster {
...
+      returns (ClusterPairProcessResponse){
+        option(google.api.http) = {
+          post: "/v1/clusterspairs/process"
+          body:"*"
+        };
+      }
+      // Get information about a cluster pair
+      rpc Get(ClusterPairGetRequest)
```
lpabon: Let's change this to Inspect() as other SDK calls

api/api.proto
```diff
@@ -951,6 +951,74 @@ service OpenStorageCluster {
...
+        }
+      // Enumerate returns list of cluster pairs
+      rpc Enumerate(SdkClusterPairsEnumerateRequest)
+        returns (ClusterPairsEnumerateResponse){
+          option(google.api.http) = {
+            get: "/v1/clusterspairs"
+          };
+        }
+        // GetToken returns a auth token
+        rpc GetToken(ClusterPairTokenGetRequest)
```
lpabon: ClusterPairTokenGetRequest takes a value as a command, so this cannot be a `Get` call.

Let's chage this to just `Token()` and be a `post` on `/v1/clusterpairs/token` REST.

api/server/sdk/cluster_pair.go
lpabon: Do you mean that the function `GetId()` gets called once? This get's optimized by the compiler, so no issues.

api/api.proto
disrani-px: Post shouldn't be used for read operarions

api/server/sdk/cluster_pair.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Cannot list cluster pairs : %v", err)
+	}
+	return &api.SdkClusterPairEnumerateResponse{
+		Result: resp,
+	}, nil
+}
+
+// Token gets the authentication token for this cluster
+func (s *ClusterPairServer) Token(
```
disrani-px: Just Token() doesn't really make sense. This is like saying Volume(). It should be GetToken()

api/server/sdk/cluster_pair.go
```diff
@@ -0,0 +1,135 @@
...
+	}, nil
+}
+
+// Inspect information about a cluster pair
+func (s *ClusterPairServer) Inspect(
+	ctx context.Context,
+	req *api.SdkClusterPairInspectRequest,
+) (*api.SdkClusterPairInspectResponse, error) {
+
+	if len(req.GetId()) == 0 {
```
disrani-px: Redundant check, should be done by the implementation

api/server/sdk/cluster_pair.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Cannot generate token : %v", err)
+	}
+	return &api.SdkClusterPairTokenResponse{
+		Result: resp,
+	}, nil
+}
+
+// ClearToken gets the authentication token for this cluster
+func (s *ClusterPairServer) ClearToken(
```
disrani-px: This API doesn't make sense. If it's to Clear the token it shouldn't return the token.

api/server/sdk/volume_migrate.go
```diff
@@ -0,0 +1,143 @@
...
+	"google.golang.org/grpc/status"
+)
+
+// Start a volume migration
+func (s *VolumeServer) Start(
+	ctx context.Context,
+	req *api.SdkCloudMigrateStartRequest,
+) (*api.SdkCloudMigrateStartResponse, error) {
+
+	if len(req.GetName()) == 0 {
```
disrani-px: All redundant checks. Should be done by the implementation not here.

api/api.proto
```diff
@@ -2517,18 +2611,69 @@ message CloudMigrateStartRequest {
...
+  // Defines a migration request for all volumes in a cluster
+  message MigrateAllVolumes { 
+  }
+
+  // ID of the cluster to which volumes are to be migrated
+  string cluster_id = 1;
+  
+  // Unique name assocaiated with this migration. 
+  // This is a required field for idempotency
+  string name = 2;
```
disrani-px: This is not a name. It is a task ID and that is what is returned when enumerating migrations.

api/server/sdk/cluster_pair.go
lpabon: I think this is clear and makes sense. It is like saying time.Now().Unix(), returns time in Unix() format. This would say `migrate.Token()` which returns a token. Seems perfectly aligned.

api/server/sdk/cluster_pair.go
lpabon: That is ok, the message here is that this check returns a 4xx or codes.InvalidArgument error code. While returning errors from the golang interface would have to be a 5xx or codes.Internal since the SDK would not know.

To make it clear to the caller, this check provides a very specific message that the Id is required and the error is caused by an invalid request.

api/server/sdk/cluster_pair.go
disrani-px: We don't do Volume() or Cluster() right? So why have different conventions for different APIs? It should be consistent I think. There is no other API here that uses just a noun for the get call.

api/server/sdk/cluster_pair.go
lpabon: I agree that it should not return a token. Is that what happens when GetToken in golang is set with the boolean to clear it? if it does, then yeah, we should not return a token.

api/server/sdk/cluster_pair.go
disrani-px: And a clear message is already being returned by the implementation. It is not the SDKs job to verify this. Different implementations could have different requirements.

api/server/sdk/volume_migrate.go
lpabon: See comment on 4xx and 5xx messages. We will try to keep these to a minimum

api/server/sdk/cluster_pair.go
disrani-px: There is no ClearToken() in the golang. While getting a token you can specify if you want to reset it before returning. So I'm not sure why this was split up in to 2 APIs here.

api/api.proto
lpabon: Can two migrations be created with the same task id? My guess is that they cannot, please let me know if I am wrong. The SDK has all idempotency values called as `name`. So this function would have to adjust to that. We will internally change it to task_id.

api/server/sdk/volume_migrate.go
disrani-px: Again, overreach from the SDK. In case the name isn't required to be passed in for the Portworx implementation, so this check is actually incorrect.

api/api.proto
disrani-px: The MigrateInfo returns taskIds, while this would take in a name. So this will cause confusion.

api/server/sdk/cluster_pair.go
lpabon: It does not matter that there is no ClearToken() in golang. The SDK's requirement is to provide a clear message and API to the callers.

Just like we created volume.Clone() (does not exist in golang) and EnumerateWithFilters. Etc.

api/server/sdk/volume_migrate.go
lpabon: Name is required for idempotency, as All SDK calls are idempotent. The SDK and CSI are both idempotent. The Golang implementations may not be.

api/api.proto
lpabon: It will not cause confusion at all because Volume.Create takes `name` and it provides a `volume_id`. This is the same.

api/server/sdk/cluster_pair.go
disrani-px: There is an Volume.Inspect(), is there not? Having just a noun as the API doesn't make sense. Using old Unix APIs is not a valid argument.

api/server/sdk/cluster_pair.go
disrani-px: Having a GetToken() with a parameter whether to reset is clear enough.

api/server/sdk/volume_migrate.go
disrani-px: A taskId is not required. It will be generated if not supplied. The golang implementation is idempotent. Just requiring a name here doesn't make anything idempotent. The implementation needs to guarantee that.

api/api.proto
disrani-px: It is not the same. For volumes there are 2 separate fields, name and Id. For migrations there is just one field i.e taskId, there is no name returned in the migrateInfo.

api/server/sdk/cluster_pair.go
lpabon: i thought more on the Token vs GetToken, and i think GetToken() is fine. That is what the golang proto buf does for every value anyways.. GetXXX().

api/server/sdk/cluster_pair.go
lpabon: We discussed offline, and we think that the underlying Golang interfaces can do one of two things:

* Return grpc code status errors
* Return wellknown error codes which we can then map to gRPC codes.

api/server/sdk/volume_migrate.go
lpabon: talked offline, @disrani-px will check in #696 checks. He will point out any that do not need checks.

api/api.proto
lpabon: Talked offline, and we will rename this back to task_id and make it optional. This will allow the caller to decide if they want it idempotent or not. The documentation will show this and that it is optional.

Also, we need an issue for the cloudbackup change `name` to `task_id`.

sharma-tapas:
@lpabon  and @disrani-px  Moved most of this changes to a new PR #696

PR #696, opened 2018-11-01
Implementation of ClusterPairing and Migration API's for the SDK.

sharma-tapas:
This commit contains the API's in the SDK for the ClusterPairing and VolumeMigration use cases.

Part of #653

@lpabon This is a cleaner PR with all commits squashed into one and lesser merge conflicts.


lpabon:
This looks like a replacement for #692, right? We need to make sure @disrani-px is ok with the review here also.

api/server/sdk/server.go
```diff
@@ -369,6 +372,32 @@ func (s *Server) restServerSetupHandlers() (*http.ServeMux, error) {
...
+	err = api.RegisterOpenStorageMigrateHandlerFromEndpoint(
+		context.Background(),
+		gmux,
+		s.Address(),
+		[]grpc.DialOption{grpc.WithInsecure()})
+	if err != nil {
+		return nil, err
+	}
+
+	err = api.RegisterOpenStorageMigrateHandlerFromEndpoint(
```
lpabon: I think this one is not needed since 384 is adding it already.

lpabon:
Also, from #692 @disrani-px and I discussed it and here are the results:

* [Token](https://github.com/libopenstorage/openstorage/pull/692#discussion_r230204079): Use GetToken, since golang protobuf does this anyways for all accesses.
* [Checking parameters](https://github.com/libopenstorage/openstorage/pull/692#discussion_r230206386): @disrani-px will review this PR and let you know which of the values not to check for. We will also be adding an issue to have the inner/Golang interfaces the ability to return known errors so that the SDK does not just return codes.Internal.
* [ResetToken](https://github.com/libopenstorage/openstorage/pull/692#discussion_r229874517): Using ResetToken is ok
* [Idempotency](https://github.com/libopenstorage/openstorage/pull/692#discussion_r230209700): `name` will be changed back to `task_id` and be made optional. This will allow the client to determine if they want the function to be idempotent or not. The documentation for `task_id` will show to set the value for idempotency.

api/server/sdk/server.go
sharma-tapas: Done.

sharma-tapas:
@lpabon  Following comments were addressed in this commit,
* Token: Used GetToken.
* Idempotency: name is changed back to task_id and is made optional.

Waiting for @disrani-px  for the following,
* Checking parameters: @disrani-px will review this PR and let you know which of the values not to check for. We will also be adding an issue to have the inner/Golang interfaces the ability to return known errors so that the SDK does not just return codes.Internal.

PR #723, opened 2018-11-08
New API for volume/snapshot capacity usage.

talakad:
Returns exclusive and shared capacity in use by a volume/snapshot.

Where possible this would show the actual storage in use by volume/snapshot.
Closes #


api/api.proto
```diff
@@ -460,6 +460,19 @@ message Stats {
   uint64 interval_ms = 10;
 }
 
+message CapacityUsageInfo {
+  // ExclusiveBytes used by this snap
+  uint64 exclusive_bytes = 1;
+  // SharedBytes used by this snap with other snaps and/or parent
+  uint64 shared_bytes = 2;
+  // TotalByes used by this volume
```
kschoche: typo TotalByes

api/api.proto
```diff
@@ -460,6 +460,19 @@ message Stats {
   uint64 interval_ms = 10;
 }
 
+message CapacityUsageInfo {
+  // ExclusiveBytes used by this snap
+  uint64 exclusive_bytes = 1;
```
lpabon: Use int64's

api/api.proto
```diff
@@ -460,6 +460,19 @@ message Stats {
   uint64 interval_ms = 10;
 }
 
+message CapacityUsageInfo {
+  // ExclusiveBytes used by this snap
```
lpabon: Provide enough information on each documentation to explain what it is used for and how to interpret it for those who do not know much about storage.

api/api.proto
```diff
@@ -460,6 +460,19 @@ message Stats {
...
+  uint64 exclusive_bytes = 1;
+  // SharedBytes used by this snap with other snaps and/or parent
+  uint64 shared_bytes = 2;
+  // TotalBytes used by this volume
+  uint64 total_bytes = 3;
+  // ErrorString indicates any error encountered for excl/shared
+  // fields
+  string error_details = 4;
```
lpabon: This is confusing because errors should be returned in the RPC call instead.

api/api.proto
```diff
@@ -460,6 +460,19 @@ message Stats {
...
+  // IsSnap is set true for shapshots
+  bool is_snap = 5;
```
lpabon: What is the purpose of this value? What would be the use of it?

api/api.proto
```diff
@@ -460,6 +460,19 @@ message Stats {
   uint64 interval_ms = 10;
 }
 
+message CapacityUsageInfo {
```
lpabon: Add a small comment to the message itself

api/server/sdk/volume_ops.go
```diff
@@ -352,6 +352,50 @@ func (s *VolumeServer) Stats(
...
+			err.Error())
+	}
+
+	resp := &api.SdkVolumeCapacityUsageResponse{}
+	resp.CapacityUsageInfo = &api.CapacityUsageInfo{}
+	if dResp != nil {
+		resp.CapacityUsageInfo.ExclusiveBytes = dResp.CapacityUsageInfo.ExclusiveBytes
+		resp.CapacityUsageInfo.SharedBytes = dResp.CapacityUsageInfo.SharedBytes
+		resp.CapacityUsageInfo.TotalBytes = dResp.CapacityUsageInfo.TotalBytes
+		if dResp.ErrorDetails != "" {
```
lpabon: Instead of comparing strings, let's make this an `error` val.

talakad: done.

PR #725, opened 2018-11-10
Extend storage resource to have cache as an attribute

sulakshm:
Extend storage resource to have cache as an attribute. This allows assigned drives on a Node to be marked for caching purpose.

Closes #

This is an initial extension, there will be follow-up bells and whistles to extend this further.


api/api.proto
```diff
@@ -2542,7 +2544,7 @@ message SdkVersion {
     // SDK version major value of this specification
     Major = 0;
     // SDK version minor value of this specification
-    Minor = 23;
+    Minor = 24;
```
lpabon: Thanks, this looks good. I can see that StorageResource is used by StorageNode. Could you please add a comment to the SDK_CHANGELOG.md at the root of the file.

lpabon:
Also, you may want to check your commit message. Looks like the 'sign-off' is wrong. Did you use `git commit -s` to setup the signature?  It looks like the signature does not have a space between the email and the name.

```
    Signed-off-by: Lakshmi Narasimhan Sundararajan<lns@portworx.com>

```

Example:
```
commit fad6f85249201512e36ea2051607a0411d8ed198 (github/fake-backup)
Author: Luis Pabón <luis@portworx.com>
Date:   Tue Nov 6 14:37:38 2018 -0500

    Fix SrcVolumeId not in SdkCloudStatus
    
```

api/api.proto
sulakshm: yes done and revised.

PR #735, opened 2018-11-15
sdk: Rework all REST routes

lpabon:
The routes in the SDK Gateway needed to be fixed due to collisions.


api/api.proto
```diff
@@ -1501,34 +1501,37 @@ service OpenStorageCloudBackup {
     }
 
   // Return a list of backups for the specified volume
-  rpc Enumerate(SdkCloudBackupEnumerateRequest)
-    returns (SdkCloudBackupEnumerateResponse) {
+  rpc EnumerateWithFilters(SdkCloudBackupEnumerateWithFiltersRequest)
+    returns (SdkCloudBackupEnumerateWithFiltersResponse) {
       option(google.api.http) = {
-        get: "/v1/cloudbackups/volume"
+        post: "/v1/cloudbackups/enumerate/filters"
```
disrani-px: This should not be a post, it should be a get like it was before. post is used for updates, not for getting information

PR #745, opened 2018-11-18
Add Eta fields  to cloudbackup status structure.

talakad:

Cloud backup status returned by volume driver was missing these fields, which are already in sdk.
Closes #

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -340,12 +340,6 @@ type CloudBackupInfo struct {
 	Metadata map[string]string
 	// Status indicates the status of the backup
 	Status string
-	// BytesTotal is the number of bytes being transferred
-	BytesTotal uint64
-	// BytesDone is the number of bytes already transferred
-	BytesDone uint64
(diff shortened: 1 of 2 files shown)


api/api.go
```diff
@@ -414,8 +414,12 @@ type CloudBackupStatus struct {
 	OpType CloudBackupOpType
 	// State indicates if the op is currently active/done/failed
 	Status CloudBackupStatusType
-	// BytesDone indicates total Bytes uploaded/downloaded
+	// BytesDone indicates Bytes uploaded/downloaded so far
 	BytesDone uint64
+	// BytesTotal is the total number of bytes being transferred
```
disrani-px: These were added to CloudBackupInfo, needs to be removed from there
https://github.com/libopenstorage/openstorage/blob/9afeef4311981abcba1c723997d09ad515a0e82f/api/api.go#L329

talakad: thanks, will remove from there.

PR #748, opened 2018-11-20
Rel 4 cloudsnap eta

talakad:
Cherrypicking changes related to cloud backup and cloud migrate ETA from master.
Closes #


api/api.proto
```diff
@@ -2743,8 +2746,16 @@ message CloudMigrateInfo {
...
     // Contains the reason for the migration error
     string error_reason = 10;
-    // time_remaining is the estimated time remaining to completion
-    int64 time_remaining = 11;
+    // StartTime indicates Op's start time
+    google.protobuf.Timestamp start_time = 11;
+    // CompletedTime indicates Op's completed time
+    google.protobuf.Timestamp completed_time = 12;
+    // BytesTotal is the number of bytes being transferred
+    uint64 bytes_total = 13;
```
lpabon: We should not be using uint64 due to issues in Java clients. We need to use int64.

talakad: Since this is how its in master.. can we push this in.. and then create a ticket for change in both master and this branch?

lpabon: More info: https://github.com/container-storage-interface/spec/issues/172

PR #754, opened 2018-11-21
Bind mount path so we can chattr on it.

piyush-nimbalkar:
If the original mount path is read-only and we cannot chattr it,
then we create a bind mount in `/var/lib/osd/mounts`. We can then
chattr this bind mounted path and proceed with the mount.

Closes #

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"strings"
 	"sync"
 	"syscall"
@@ -20,6 +21,7 @@ import (
 	"github.com/libopenstorage/openstorage/pkg/keylock"
 	"github.com/libopenstorage/openstorage/pkg/options"
 	"github.com/libopenstorage/openstorage/pkg/sched"
+	"github.com/libopenstorage/openstorage/volume"
 	"github.com/sirupsen/logrus"
 )
 
@@ -479,8 +481,20 @@ func (m *Mounter) Mount(
 
 	// Record previous state of the path
 	pathWasReadOnly := m.isPathSetImmutable(path)
+	isBindMounted := false
+
 	if err := m.makeMountpathReadOnly(path); err != nil {
-		return fmt.Errorf("failed to make %s readonly Err: %v", path, err)
+		if strings.Contains(err.Error(), "Inappropriate ioctl for device") {
+			logrus.Warnf("failed to make %s readonly. Err: %v", path, err)
+			// If we cannot chattr the original mount path, we bind mount it to
+			// a path in osd mount path and then chattr it
+			if err := m.bindMountOriginalPath(path); err != nil {
+				return err
+			}
+			isBindMounted = true
+		} else {
+			return fmt.Errorf("failed to make %s readonly. Err: %v", path, err)
+		}
 	}
 
 	// The device is not mounted at path, mount it and add to its mountpoints.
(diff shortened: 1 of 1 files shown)


pkg/mount/mount.go
```diff
@@ -479,8 +481,20 @@ func (m *Mounter) Mount(
 
 	// Record previous state of the path
 	pathWasReadOnly := m.isPathSetImmutable(path)
+	isBindMounted := false
+
 	if err := m.makeMountpathReadOnly(path); err != nil {
-		return fmt.Errorf("failed to make %s readonly Err: %v", path, err)
+		if strings.Contains(err.Error(), "Inappropriate ioctl for device") {
```
talakad: when do we see this error? this is not expected in normal path. right?

pkg/mount/mount.go
```diff
@@ -702,6 +784,16 @@ func New(
 	return nil, ErrUnsupported
 }
 
+// bindMountPath return the path used to bind mount to a user given mount path.
+// This allows us to chattr the path in case we cannot chattr the original mount path.
+func bindMountPath(path string) (string, error) {
```
adityadani: This will not work for shared volumes where the last item in the path can be the volumeID and we get two mount request on the same node.

pkg/mount/mount.go
adityadani: If the mountpath such as `/var/lib/kubelet` is on a filesystem like overlay which does not support chattr

pkg/mount/mount.go
```diff
@@ -501,6 +530,36 @@ func (m *Mounter) Mount(
 	return nil
 }
 
+func (m *Mounter) bindMountOriginalPath(path string) error {
+	bindMountPath, err := bindMountPath(path)
```
adityadani: nit: rename the function to getBindMountPath()

pkg/mount/mount.go
```diff
@@ -702,6 +784,16 @@ func New(
...
 }
 
+// bindMountPath return the path used to bind mount to a user given mount path.
+// This allows us to chattr the path in case we cannot chattr the original mount path.
+func bindMountPath(path string) (string, error) {
+	pathSlice := strings.Split(path, "/")
+	if len(pathSlice) == 0 {
+		return "", fmt.Errorf("mount path cannot be empty")
+	}
+	return filepath.Join(volume.MountBase, pathSlice[len(pathSlice)-1]), nil
```
adityadani: The source for bind mount should have a prefix like `/var/lib/osd/mounts/readonly/<id>`. Docker volume driver mounts the volumes on volume.MountBase/<id>

pkg/mount/mount.go
```diff
@@ -501,6 +530,36 @@ func (m *Mounter) Mount(
 	return nil
 }
 
+func (m *Mounter) bindMountOriginalPath(path string) error {
+	bindMountPath, err := bindMountPath(path)
+	if err != nil {
+		return err
+	}
+	if bindMountPath == path {
```
adityadani: this should not happen right?

pkg/mount/mount.go
```diff
@@ -479,8 +502,23 @@ func (m *Mounter) Mount(
...
 	// Record previous state of the path
 	pathWasReadOnly := m.isPathSetImmutable(path)
+	var (
+		isBindMounted bool = false
+		bindMountPath string
+	)
+
 	if err := m.makeMountpathReadOnly(path); err != nil {
-		return fmt.Errorf("failed to make %s readonly Err: %v", path, err)
+		if strings.Contains(err.Error(), "Inappropriate ioctl for device") {
```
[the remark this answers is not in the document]
piyush-nimbalkar: As discussed offline, not going to check whether it is mounted or not, as there could be an existing mount from the container orchestrator. This could happen only if an existing fuse mount is not removed. In that case we would get an ioctl error and we will mount on top the existing mount. So it will not be a problem even if there exists a mount.

PR #757, opened 2018-11-26
added config caller

sdeoras:

The osdconfig.NewManager API in openstorage is currently starting a kvdb watch on two sub trees.

However in porx, the NewManager API is called by multiple components which causes unnecessary watches to be started on etcd. Currently there are 12 osdconfig watches instead of 2 started when PX comes up.

A watch on etcd is an expensive op and should be started only once for osdconfig.
Ideally the NewManager API and the watch logic should be separated.

Closes #

### DIFF
--- a/osdconfig/new.go
+++ b/osdconfig/new.go
@@ -14,6 +14,13 @@ func NewManager(kv kvdb.Kvdb) (ConfigManager, error) {
 	return newManager(kv)
 }
 
+// NewCaller can be used to instantiate ConfigCaller
+// It does not start kvdb watches and is therefore,
+// less expensive than NewManager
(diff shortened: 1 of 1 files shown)


osdconfig/new.go
```diff
@@ -40,3 +47,20 @@ func newManager(kv kvdb.Kvdb) (*configManager, error) {
 
 	return manager, nil
 }
+
+// newCaller can be used to instantiate configManager,
+// however, it is exported as ConfigCaller and avoids
+// starting kvdb watches.
+// Those not needing kvdb wtches should use this instead of
+// config manager.
+func newCaller(kv kvdb.Kvdb) (*configManager, error) {
```
adityadani: nit: newManager can call newCaller. Reduces code duplication.

PR #759, opened 2018-11-26
changing osdconfig usage to caller

sdeoras:

Follow up on the previous pr, changing osdconfig usage to a caller based api (no watches required)

Closes #


cluster/manager/manager.go
```diff
@@ -1223,7 +1223,7 @@ func (c *ClusterManager) StartWithConfiguration(
 		return err
 	}
 
-	c.startClusterDBWatch(lastIndex, kv)
+	_ = c.startClusterDBWatch(lastIndex, kv)
```
adityadani: this change is not needed right?

sdeoras: correct, my editor was complaining on unhandled error :)

PR #765, opened 2018-11-28
Returning 422 for invalid credentials

piyush-nimbalkar:

Sending http error code 422 when we get an error validating a credential.
Based on server error, return a fixed ErrInvalidCredential error to differentiate between invalid creds and other errors.
 
Closes #

### DIFF
--- a/api/server/creds.go
+++ b/api/server/creds.go
@@ -84,7 +84,7 @@ func (vd *volAPI) credsValidate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := d.CredsValidate(uuid); err != nil {
-		vd.sendError(vd.name, method, w, err.Error(), http.StatusInternalServerError)
+		vd.sendError(vd.name, method, w, err.Error(), http.StatusUnprocessableEntity)
 		return
 	}
 	w.WriteHeader(http.StatusOK)
--- a/volume/errors.go
+++ b/volume/errors.go
@@ -1,22 +1,13 @@
 package volume
 
-// Custom credential operation error codes
-const (
-	_ = iota + 100
-	// ErrInvalidCredential is returned when a credential is invalid
-	ErrInvalidCredential
-)
-
 // CredentialError error returned for credential operations
 type CredentialError struct {
-	// Code is one of credential operation error codes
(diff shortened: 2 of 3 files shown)


volume/errors.go
```diff
@@ -0,0 +1,24 @@
+package volume
```
lpabon: This is just like grpc status. Can we use that instead?

api/server/creds.go
```diff
@@ -84,7 +84,7 @@ func (vd *volAPI) credsValidate(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := d.CredsValidate(uuid); err != nil {
-		vd.sendError(vd.name, method, w, err.Error(), http.StatusInternalServerError)
+		vd.sendError(vd.name, method, w, err.Error(), http.StatusUnprocessableEntity)
```
lpabon: Isn't this assuming what the error is? d.CredsValidate(uuid) should return an known error, then api/server/creds can check for this error

volume/errors.go
```diff
@@ -0,0 +1,24 @@
...
+
+// Custom credential operation error codes
+const (
+	_ = iota + 100
+	// ErrInvalidCredential is returned when a credential is invalid
+	ErrInvalidCredential
```
disrani-px: I don't think we need Code, since we don't know what failed when validating. The message has enough info on what went wrong

api/server/creds.go
piyush-nimbalkar: Currently we return an error with message which is the reason for cred validation. So for any cred validation error, I want to return 422.
500 is too generic and later we would want to send 500 for something else, hence using 422 here to indicate cred validation failure.

volume/errors.go
piyush-nimbalkar: Was just trying to make it future proof, where we should ideally be sending specific error codes. For now I will switch to grpc status instead of introducing a new error.

disrani-px: I don't think we should use grpc errors for the REST APIs

piyush-nimbalkar: Okay, will keep it as is and remove the code.

PR #769, opened 2018-11-30
Add method to update scheduler node name of current node

piyush-nimbalkar:

We need a way to update the scheduler node name in cluster database at any point in time instead of just startup. This PR has a proposed method of having separate method to update specifically the scheduler node name, similar to `UpdateLabels()` and `UpdateData()`.

Another solution was to add a new method - `UpdateNode(node api.StorageNode)`, and update all the parameters that have changed in the StorageNode object. We could probably replace the `UpdateLabels` method with this because `StorageNode` object already has the labels.

Let me know what you guys think.

Closes #

### DIFF
--- a/api/client/cluster/client.go
+++ b/api/client/cluster/client.go
@@ -192,6 +192,10 @@ func (c *clusterClient) UpdateLabels(nodeLabels map[string]string) error {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


lpabon:
Looks good. There probably should be a unit tests for the manager function right?

piyush-nimbalkar:
Updated the branch and added a test for UpdateSchedulerNodeName.

PR #774, opened 2018-12-04
Change rand/rw iops and bandwidth to uint64 instead of float

kschoche:
Signed-off-by: Kyle Schochenmaier <kyle@portworx.com>

float64 is not compatible with protobuf version we use and we're only using it to store tenth of KB/s, use uint64 instead
Closes #


kschoche:
Closing as not needed anymore

PR #778, opened 2018-12-05
Add task id and cluster id to cloud migrate status request

disrani-px:

Closes #

### DIFF
--- a/api/api.proto
+++ b/api/api.proto
@@ -2766,8 +2766,19 @@ message CloudMigrateInfoList {
     repeated CloudMigrateInfo list = 1;
 }
 
-// Empty Request
+// Request for cloud migration operation status
 message SdkCloudMigrateStatusRequest {
+    // Request contains the task id and cluster id for which status should be
+    // returned
+    CloudMigrateStatusRequest request = 1;
+}
+
+// Request for cloud migration operation status
+message CloudMigrateStatusRequest {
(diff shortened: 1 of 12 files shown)


api/api.proto
```diff
@@ -2766,8 +2766,19 @@ message CloudMigrateInfoList {
...
+    // returned
+    CloudMigrateStatusRequest request = 1;
+}
+
+// Request for cloud migration operation status
+message CloudMigrateStatusRequest {
+    // Task id for which to return status
+    string task_id = 1;
+    // ID of the cluster for which to return migration statuses
+    string cluster_id = 2;
```
talakad: How does this work? i guess if nothing is mentioned, all task ids are returned. if cluster id is not passed, will it return taskID from both the clusters?

api/api.proto
```diff
@@ -2766,8 +2766,19 @@ message CloudMigrateInfoList {
     repeated CloudMigrateInfo list = 1;
 }
 
-// Empty Request
+// Request for cloud migration operation status
 message SdkCloudMigrateStatusRequest {
+    // Request contains the task id and cluster if for which status should be
```
talakad: typo: cluster id

api/api.proto
disrani-px: Correct, it'll try to match both if both are specified, otherwise it'll try to match either.

PR #781, opened 2018-12-06
Use empty request if nothing was sent for migrate status

disrani-px:

Closes #

### DIFF
--- a/api/server/migrate.go
+++ b/api/server/migrate.go
@@ -68,9 +68,12 @@ func (vd *volAPI) cloudMigrateStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


api/server/migrate.go
```diff
@@ -68,9 +68,12 @@ func (vd *volAPI) cloudMigrateStatus(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := json.NewDecoder(r.Body).Decode(statusReq); err != nil {
-		vd.sendError(vd.name, method, w, err.Error(), http.StatusBadRequest)
-		return
+	// Use empty request if nothing was sent
```
talakad: when would this happen? backward compatibility?

disrani-px: Yes, old clients don't send anything for the request.

PR #787, opened 2018-12-07
Use sdk for docker commands and add docker volume integration test

### LINKED ISSUE #785: docker: Integration test
Create a small shell script under the hack directory that:

1. starts the osd process and sets up the docker unix domain socket.
2. using docker create and delete a volume.

This should test: Docker -> osd[Docker Server] -> osd[SDK Server] -> osd[Golang interfaces]

Then we can add to .travis

### LINKED ISSUE #786: docker: Change basic docker commands to use sdk
Includes the following:
* docker volume create
* docker volume rm
* docker volume inspect
* docker volume list
* path

ggriffiths:
* Switched the following commands to use the sdk server instead of the go interfaces
  * create
  * remove
  * get
  * list
  * path
* Also made the above commands parse out a token from the docker request and added to sdk request metadata.
* Added hack/docker-integration-test.sh
  * starts osd as a background process
  * tests docker volume create, inspect, rm

*Which issue(s) this PR fixes** (optional)
Closes #786 
Closes #785 

* includes some code copied from @lpabon's sdk-auth-4-ownership branch which hasn't been merged yet

### DIFF
--- a/api/server/docker.go
+++ b/api/server/docker.go
@@ -1,20 +1,25 @@
 package server
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
 	"os"
 	"path"
+	"sync"
 
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/spec"
 	"github.com/libopenstorage/openstorage/config"
+	"github.com/libopenstorage/openstorage/pkg/grpcserver"
 	"github.com/libopenstorage/openstorage/pkg/options"
 	"github.com/libopenstorage/openstorage/pkg/util"
 	"github.com/libopenstorage/openstorage/volume"
 	"github.com/libopenstorage/openstorage/volume/drivers"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 const (
@@ -26,6 +31,10 @@ const (
 type driver struct {
 	restBase
 	spec.SpecHandler
+
+	sdkUds string
+	conn   *grpc.ClientConn
+	mu     sync.Mutex
 }
 
 type handshakeResp struct {
@@ -64,8 +73,13 @@ type capabilitiesResponse struct {
 	Capabilities capabilities
 }
 
-func newVolumePlugin(name string) restServer {
-	return &driver{restBase{name: name, version: "0.3"}, spec.NewSpecHandler()}
+func newVolumePlugin(name, sdkUds string) restServer {
+	d := &driver{
+		restBase:    restBase{name: name, version: "0.3"},
+		SpecHandler: spec.NewSpecHandler(),
+		sdkUds:      sdkUds,
+	}
+	return d
 }
 
 func (d *driver) String() string {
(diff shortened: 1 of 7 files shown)


api/server/docker.go
```diff
@@ -169,79 +215,132 @@ func (d *driver) mountpath(name string) string {
 	return path.Join(volume.MountBase, name)
 }
 
+func (d *driver) getConn() (*grpc.ClientConn, error) {
+	if d.conn == nil {
```
lpabon: i hadn't placed any locks here, but we should have locks on this function so that we do not connect many times before d.conn becomes non-null

api/testing/testing_test.go
```diff
@@ -63,6 +63,7 @@ func TestAll(t *testing.T) {
 
 	server.StartPluginAPI(
 		nfs.Name,
+		"localhost:9100",
```
lpabon: is an sdk server running at this port? where was it created?

api/spec/spec_handler.go
```diff
@@ -44,6 +44,8 @@ type SpecHandler interface {
 		string,
 	)
 
+	GetTokenFromString(str string) (string, bool)
```
lpabon: Need comment for this one.

api/server/docker.go
```diff
@@ -169,79 +215,132 @@ func (d *driver) mountpath(name string) string {
...
+	ctx := r.Context()
+
+	// decode request
 	request, err := d.decode(method, w, r)
 	if err != nil {
 		return
 	}
 
+	// attach token in context metadata
+	d.attachToken(ctx, request)
```
lpabon: this function mutates `ctx`. Instead it should return ctx:

```go
ctx, err = d.attachToken(ctx, request)
```

Which is aligned with other context calls.

api/server/docker.go
```diff
@@ -169,79 +215,132 @@ func (d *driver) mountpath(name string) string {
...
 	if err != nil {
-		d.logRequest(method, "").Warnf("Cannot locate volume driver")
 		d.errorResponse(method, w, err)
 		return
 	}
+	volumes := api.NewOpenStorageVolumeClient(conn)
 
-	_, _, _, _, name := d.SpecFromString(request.Name)
+	// find id to delete
+	enumerateResp, err := volumes.EnumerateWithFilters(ctx, &api.SdkVolumeEnumerateWithFiltersRequest{
```
lpabon: This and line 144 can be made into a function that returns a volume id.

api/server/docker.go
ggriffiths: Good idea, will add

api/testing/testing_test.go
ggriffiths: No there isn't, will just set to the empty string to avoid confusion

api/server/docker.go
```diff
@@ -169,79 +226,123 @@ func (d *driver) mountpath(name string) string {
...
 
-	v, err := volumedrivers.Get(d.name)
+	// attach token in context metadata
+	ctx = d.attachToken(ctx, request)
+
+	// get spec for deletion
+	specParsed, _, _, _, name := d.SpecFromString(request.Name)
+	d.logRequest(method, name).Infoln("")
+	if !specParsed {
+		_, _, _, err = d.SpecFromOpts(request.Opts)
```
adityadani: nit: if you are not using the opts anywhere for remove, there is no reason to error out here?

api/server/docker.go
```diff
@@ -127,6 +141,38 @@ func (d *driver) volFromName(name string) (*api.Volume, error) {
...
 
+func (d *driver) volFromNameSdk(ctx context.Context, volumes api.OpenStorageVolumeClient, name string) (*api.Volume, error) {
+	// get volume id
+	volId, err := d.volIdFromName(ctx, volumes, name)
+	if err != nil {
+		return nil, err
+	}
+
+	// inspect for actual volume
+	inspectResp, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{
```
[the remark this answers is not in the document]
ggriffiths: @lpabon says he'll talk to you offline about this

ggriffiths:
Yes, will have another PR for attach/mount/detach/unmount @adityadani

PR #790, opened 2018-12-11
Only send e-mail notifications from Travis when the build fails, or transitions from failure to success.

rodrigc:

Reduce e-mails sent out by travis


rodrigc:
Similar to what was done here:
https://github.com/portworx/kvdb/pull/69

Issue #799, opened 2018-12-20
REST 9001 w/ Security


ggriffiths:
Changes include:
* Redirect Volume SDK calls
* Testing

lpabon:
Part of #448

Issue #448, opened 2018-05-31
SDK Auth Proposal


lpabon:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE

# SDK User Authentication

## Overview

The SDK API is currently available to any who can access it and provides no security model. This proposal provides a model for securing the SDK access with authentication and authorization of users. The model used by this proposal is based on the [OpenID Connect standard](http://openid.net/connect/), or OIDC, used by Kubernetes, Google, Twitter, SalesForce, just to name a few.

## Goals

* Provide a method where users can be authenticated
* Provide a method where users can be authorized
* Have no user management, only verification of tokens

## Non-goals

* User management is something left to the user to manage. This project will only verify the tokens provided by the user.

## Design

* Admin would configure OpenStorage server for a specified OIDC issuer.
* User would get a token from a OIDC issuer
* User would pass token with every SDK API call to OpenStorage.
* OpenStorage SDK Server would get the public keys from the issuer.
* OpenStorage would verify the token and check if the user is authorized.

#### Server configuration
To verify tokens, The server would need to be configured to allow a single, or multiple (TBD), OIDC issuers. For each issuer, the system would need the following configuration:

* oidc-client-id: Client ID provided by authentication issuer
* oidc-issuer-url: Issuer url, (e.g. https://accounts.google.com)
* oidc-ca-file: Path to certifificate. Needed if accessing internal issuers.
* oidc-admin-users: A list of admin users. Format is TBD, but could be something like: `sub:username`, `email:some@email.com`, following the format `{JWT claim key}:{JWT claim value}`

#### Generating a Token
A user would use a tool like [oidc-gen-token](https://github.com/lpabon/oidc-gen-token), [kube-oidc](https://github.com/ericchiang/kube-oidc), or any other tool which would authenticate the user with an OpenID Connect issuer to provide them with a OIDC token.

User would then use this token for all the API calls to the SDK. Both the REST gRPC Gateway and the gRPC endpoint would both support this model.

#### REST Gateway token management
The REST Gateway will not verify the token. Instead it will only repackage it for the gRPC calls. It will 
expect the token in the header payload of the HTTP request, then it will place it in the metadata of the gRPC `context`.

#### gRPC Server
The gRPC Server will have an authentication interceptor which will verify the token stored in the metadata of the `context`. To verify this token it will use [go-oidc](https://github.com/coreos/go-oidc), which takes of pulling the public keys from the issuer and then verifying the token.

## Authorization
[...]

PR #800, opened 2018-12-20
REST9001 Create to SDK plus Infrastructure

lpabon:
This is the first of many changes to come to `api/server/volume.go`
to redirect volume calls to the SDK. 

Part of #448 #799 

* REST9001 unit tests to use fully running SDK with the Fake driver
and not the golang mocks to get better test coverage.
* Authentication support. Client API for Golang has been setup
to support addition of a auth token.
* All volume unit tests commented out while we move them to
the new model.
* Removed unnecessary old tests


api/server/volume_test.go
```diff
@@ -1,110 +1,88 @@
...
 	// create a volume client
 	driverclient := volumeclient.VolumeDriver(client)
-
-	res, err := driverclient.Create(req.GetLocator(), req.GetSource(), req.GetSpec())
-
-	assert.NotNil(t, err)
-	assert.EqualValues(t, "", res)
-	assert.Contains(t, err.Error(), "error in create")
+	_, err = driverclient.Create(req.GetLocator(), req.GetSource(), req.GetSpec())
+	assert.Error(t, err)
```
lpabon: I need to add here a call to the SDK and get back information about the volume to assert that it is equal to what was requested.

PR #801, opened 2019-01-02
Role support

lpabon:
Adds a role manager and the appropriate SDK calls.

Part of #448 

* This new style of creating the implementation of the SDK interface outside the SDK is probably the way will do it in the future. This is the first (afaik) object created where it does not have the older Golang interface. Instead it uses the golang interface created by the api.proto protobuf output.
* After this change, I think it would be a good idea to bring the openstorage-sdk-auth code to openstorage and deprecate that repo. I think it has served its purpose.


pkg/role/sdkserviceapi.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	// Determine if there is collision with default roles
+	if _, ok := defaultRoles[req.GetRole().GetName()]; ok {
+		return nil, status.Errorf(
+			codes.InvalidArgument,
+			"Name %s already used by system role", req.GetRole().GetName())
+	}
+
+	// Reviewer:
```
pault84: I'm actually of the opinion that all Create calls should return the created object (as created) that might be my web dev background but it's nice for verification reasons.

pkg/role/role.go
```diff
@@ -0,0 +1,32 @@
+/*
+Package role manages roles in Kvdb and provides validation
+Copyright 2018 Portworx
```
pault84: Is this ok since we technically don't own openstorage.

pkg/role/sdkserviceapi.go
```diff
@@ -0,0 +1,357 @@
...
+
+		// 'xxx*'
+		return strings.HasPrefix(s, match)
+	}
+
+	// no wildcard stars given in rule
+	return rule == s
+}
+
+// NewSdkSErviceAPIRoles returns a new SDK role manager
```
ggriffiths: comment should be `NewSdkRoleManager` returns ...

pkg/role/sdkserviceapi.go
```diff
@@ -0,0 +1,357 @@
...
+			codes.InvalidArgument,
+			"Name %s already used by system role", req.GetRole().GetName())
+	}
+
+	// Reviewer:
+	// Should we check here if the value already exists and return the object
+	// currently in the db? That is is more accurate and idempotent.
+	// This will require a change to the return value of this function (and Update)
+
+	_, err := r.kv.Put(prefixWithName(req.GetRole().GetName()), req.GetRole(), 0)
```
adityadani: You should use the API kv.Create(). If the key already exists it will return an error.
The intent of the user (and the API) was to create, if there exist a role we should return an appropriate error. If we do not return an error and return the existing role, then it is confusing.

pkg/role/sdkserviceapi.go
```diff
@@ -0,0 +1,357 @@
...
+func (r *SdkRoleManager) Delete(
+	ctx context.Context,
+	req *api.SdkRoleDeleteRequest,
+) (*api.SdkRoleDeleteResponse, error) {
+	if len(req.GetName()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply a name for role")
+	}
+
+	// Determine if there is collision with default roles
+	if _, ok := defaultRoles[req.GetName()]; ok {
```
adityadani: May be an admin, would want to remove the system roles to make it more secure ? In that case we might need a new permission to be able to delete system roles.  We can have a separate discussion for this and not handle it in this PR.

pkg/role/role.go
lpabon: Yeah, that is fine. At Red Hat, we used `Copyright <year> Red Hat` in the same manner since we worked for Red Hat. The company has the copyright

pkg/role/sdkserviceapi.go
```diff
@@ -0,0 +1,357 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	return fmt.Errorf("no accessable rule to authorize access found")
+}
+
+func (r *SdkRoleManager) checkRoleForErrors(role *api.SdkRole) error {
```
ggriffiths: nit: maybe just validateRole for this function? A little more to the point

pkg/role/sdkserviceapi.go
lpabon: yeah good point. i'll return it.

PR #805, opened 2019-01-03
Refactor label management.

### LINKED ISSUE #804: Refactor volume labels
BUG

**What happened**:
Current labeling model is confusing and needs to be cleaned up.

We will be removing the labels from Volume.VolumeSpec and instead make sure all labels go into the Volume.Locator.

lpabon:
Only Volume.Locator.VolumeLabels are now used. Volume.Spec.VolumeLabels
have been removed. SDK has also been adjusted to have simple, clear
volume label usage.

Closes #804


api/api.proto
```diff
@@ -290,8 +283,6 @@ message VolumeSpec {
   bool dedupe = 8;
   // SnapshotInterval in minutes, set to 0 to disable snapshots
   uint32 snapshot_interval = 9;
-  // VolumeLabels configuration labels
-  map<string, string> volume_labels = 10;
```
adityadani: We cannot remove the VolumeLabels right away from the Spec. The kubernetes native driver uses that to pass PVC  annotations. We will need to first migrate them to locator.VolumeLabels. I think for now we can mark this field as deprecated and remove its usage.

api/api.go
```diff
@@ -730,12 +730,6 @@ func (m *Volume) Contains(mid string) bool {
 // Copy makes a deep copy of VolumeSpec
 func (s *VolumeSpec) Copy() *VolumeSpec {
 	spec := *s
-	if s.VolumeLabels != nil {
```
adityadani: Migrate the spec.Labels to locator.Labels ?

api/api.proto
lpabon: I added it back and labeled as deprecated

PR #810, opened 2019-01-07
Fix docker volume inspect to not return token

### LINKED ISSUE #809: docker: volume inspect returns token

ggriffiths:

* Currently, for `docker volume inspect token=<token>,name=<name>` calls, we're returning the name the entire token=<token>,name=<name>. The expected behavior is to return only the name. This bug happened when we re-added logic to only return the name instead of the entire `request.Name`. request.Name is only returned when we cannot parse what the user has passed in.

Closes #809


lpabon:
included in #823

PR #813, opened 2019-01-08
OpenID Connect support

lpabon:
OIDC Support for token management and security

Part of #448

* Role in claims is now Roles
* New switches added to osd program to make it easier to use
* Reorganized token management
* Abstracted token management in the context to be consumed by Ownership
* Added `username` support to be consumed by Ownership


cmd/osd-token-generator/main.go
```diff
@@ -0,0 +1,128 @@
...
+		return
+	}
+	if len(*config) == 0 {
+		fmt.Println("Must provide a file name")
+		os.Exit(1)
+	}
+
+	// Get claims
+	claims := &auth.Claims{
+		Issuer: *issuer,
```
adityadani: Error out here if the issuer has https://

adityadani: Check if issuer has https:// and error out

pkg/auth/oidc.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	for _, requiredClaim := range requiredClaims {
+		if _, ok := claims[requiredClaim]; !ok {
+			// Claim missing
+			return nil, fmt.Errorf("Required claim %v missing from token", requiredClaim)
+		}
+	}
+
+	// Return claims
+	var sdkClaims Claims
```
adityadani: why not just return claims from line 76

pkg/auth/token.go
```diff
@@ -16,73 +16,60 @@ limitations under the License.
...
-	Issuer string
 }
 
-// Token returns a signed JWT containing the claims provided
-func Token(
-	claims *Claims,
-	signature *Signature,
-	options *Options,
-) (string, error) {
+// TokenType returns the type of token. Values are: os, oidc
```
adityadani: nit: comment should be for TokenIssuer

pkg/auth/oidc.go
lpabon: Because line 76 has no structure, it only gets the first level of the keys.

lpabon:
I'll fix these in the ownership PR which is coming tomorrow.

PR #815, opened 2019-01-08
Re-add support for REST Set

### LINKED ISSUE #814: Volume Set in REST server is not working
BUG

**What happened**:
The code does nothing now because the call was mistakenly commented out.

lpabon:
Code had been commented out incorrectly during REST -> SDK refactoring.

Closes #814

### DIFF
--- a/api/server/volume.go
+++ b/api/server/volume.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gorilla/mux"
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/errors"
+	clustermanager "github.com/libopenstorage/openstorage/cluster/manager"
 	"github.com/libopenstorage/openstorage/pkg/grpcserver"
 	"github.com/libopenstorage/openstorage/volume"
 	volumedrivers "github.com/libopenstorage/openstorage/volume/drivers"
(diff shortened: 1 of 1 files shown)


api/server/volume.go
```diff
@@ -110,6 +111,49 @@ func (vd *volAPI) parseParam(r *http.Request, param string) (string, error) {
 	return "", fmt.Errorf("could not parse %s", param)
 }
 
+func (vd *volAPI) nodeIPtoIds(nodes []string) ([]string, error) {
```
harsh-px: nit: can be `nodeIPtoIDs`

api/server/volume.go
```diff
@@ -110,6 +111,49 @@ func (vd *volAPI) parseParam(r *http.Request, param string) (string, error) {
...
+
+	if c == nil {
+		return nodeIds, fmt.Errorf("failed to get cluster instance.")
+	}
+
+	for _, idIp := range nodes {
+		if idIp != "" {
+			id, err := c.GetNodeIdFromIp(idIp)
+			if err != nil {
+				return nodeIds, err
```
harsh-px: nit: can be changed to 
```suggestion
				return nil, err
```

so that we either return all or none.

api/server/volume.go
lpabon: i just put the code back, that's what it was before

PR #817, opened 2019-01-09
Refactor label management.

lpabon:
(Cherry pick from Master #805)
Only Volume.Locator.VolumeLabels are now used. Volume.Spec.VolumeLabels
have been removed. SDK has also been adjusted to have simple, clear
volume label usage.


lpabon:
per discussion with @harsh-px , we do not need this for this branch

PR #822, opened 2019-01-15
Add method to check if a volume spec is equal to another

harsh-px:

**What this PR does / why we need it**: In Portworx, we have logic to check whether a particular volume's spec is the same as other's. Currently we directly use reflect.DeepEqual to do this. But this doesn't work since for certain fields like aggregration level, the default is 0 for a new volume's spec but we internally set this to 1 for the existing volume. So we need a deep check.

Closes #

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -3,6 +3,7 @@ package api
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


sangleganesh:
this works but would be difficult to maintain, since adding any new field need this code to be updated as well or else the filed will be skipped.
one option is to generate the comparison as we do for px (may need some work), or you could simply make a copy and reset the aggr field on both and check rest using deep.Equal and check aggr separately. my 2 cents.

PR #823, opened 2019-01-15
Docker auth support fixes

lpabon:
Docker calls some API functions like get() and path() before
calling the actual function to do the work. This makes the token
available to the first call, but not to the work function.

This fix provides a cookie for the volume, where it saves
the token for a time, so that it can be used for the actual function
that does the work.

Part of #448 
Includes #810 

### DIFF
--- a/api/server/docker.go
+++ b/api/server/docker.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"sync"
+	"time"
 
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/spec"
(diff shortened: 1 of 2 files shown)


api/server/docker.go
```diff
@@ -543,7 +574,66 @@ func (d *driver) get(w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(map[string]volumeInfo{"Volume": volInfo})
 }
 
+func (d *driver) unmountInsecure(w http.ResponseWriter, r *http.Request) {
```
lpabon: This code is the copied code before it was redirected to the SDK

lpabon:
Fixed by #831

PR #831, opened 2019-01-18
Docker CE Volume API support for security

lpabon:
Docker calls were found that not all can be secured. The
list and unmount must be allowed to be insecure to make
docker ce work correctly.

Closes #


api/server/docker.go
```diff
@@ -412,10 +422,35 @@ func (d *driver) mount(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	d.logRequest(method, request.Name).Infof("response %v", response.Mountpoint)
 	json.NewEncoder(w).Encode(&response)
 }
 
+func (d *driver) list(w http.ResponseWriter, r *http.Request) {
```
lpabon: Same code from before

PR #836, opened 2019-01-22
Auth support for container orchestrators

adityadani:

Adds a middleware on the REST server.
- The middleware consults the configured secrets provider and fetches the token from it.
- The token then is inserted in the header and passed onto the next handler on the REST server.

Currently handles the following APIs
- Create
- Delete
- Attach/Detach
- Mount/Unmount
- Set (only for resize)

For all the other APIs the middleware will not get invoked.

#799 

Middleware code changes
- api/server/middleware_auth.go & api/server/server.go

Auth and Secrets changes
- pkg/auth/secrets


lpabon:
New change looks good 👍

PR #844, opened 2019-01-25
No owner is public and only owner can delete vol

### LINKED ISSUE #838: Allow permission when ownership.owner is empty
BUG

**What happened**:
During upgrades of systems with existing volumes, they will have empty ownership. A user can then adjust the acls of a volume, blocking access to others, even though the user cannot set owner, only admin can.

Instead, we should make the volume public unless it has an owner. In other words, when the user changes the acls, they will not taken into affect because the owner will still be unassigned.

### LINKED ISSUE #839: Do not allow non-owners to delete the volume
BUG

**What happened**:
When a volume owner gives access to a group or collaborator, they should not have the permission to delete the volume. Deletion of a volume should be an admin or owner only command.

lpabon:

* Only owners are allowed to delete their volume
* Resource with no owner is public

Closes #838
Closes #839

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -952,10 +952,41 @@ func (v *Volume) IsPermitted(ctx context.Context) bool {
 }
 
 func (v *VolumeSpec) IsPermitted(ctx context.Context) bool {
+	if v.IsPublic() {
+		return true
+	}
+
+	// Volume is not public, check permission
+	if userinfo, ok := auth.NewUserInfoFromContext(ctx); ok {
+		// Check Access
+		return v.IsPermittedFromUserInfo(userinfo)
+	} else {
+		// There is no user information in the context so
+		// authorization is not running
+		return true
+	}
+}
+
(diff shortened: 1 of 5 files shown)


lpabon:
This PR brought to my attention that even though collaborators/groups where given access, there also should be access types. 

Instead I think the following would be better:

```
message Ownership {
  enum AccessType {
    // Read access only
    Read = 0;
    // Write access
    Write = 1;
    // Administrator access (can delete)
    Admin = 2;
  }

  message AccessControl {
    map <string, AccessType> groups = 1;
    map <string, AccessType> collaborators = 2;
  }
```

This would change groups and collaborators in the AccessControl from slices to maps to access types. These types are based on repo access types for Github.

These access types for volumes would have the following effects:

1. Read: Can backup, snap, enumerate, and list the volume, but cannot mount or delete.
2. Write: Can also mount/unmount the volume.
3. Admin: Can delete the volume

lpabon:
Proposing #851 instead

PR #848, opened 2019-01-27
sdk: Fix scaled volumes

### LINKED ISSUE #847: sdk: Scaling not working in docker
BUG

**What happened**:
Scaling volumes in SDK were taken from docker, and the code is missing this:

https://github.com/libopenstorage/openstorage/blob/04262c5fe78fca67cb82510fcb26161182e2e445/api/server/docker.go#L423

lpabon:
Scaled volumes are not working due to a misconfiguration

Closes #847 

* Compare changes to https://github.com/libopenstorage/openstorage/blob/release-5.0/api/server/docker.go#L422-L450 . They should be similar

### DIFF
--- a/api/server/sdk/volume_node_ops.go
+++ b/api/server/sdk/volume_node_ops.go
@@ -151,33 +151,44 @@ func (s *VolumeServer) Mount(
 	}
 
 	// Get access rights
-	if err := s.checkAccessForVolumeId(ctx, req.GetVolumeId()); err != nil {
+	// Checks for ownership
+	resp, err := s.Inspect(ctx, &api.SdkVolumeInspectRequest{
+		VolumeId: req.GetVolumeId(),
+	})
+	if err != nil {
 		return nil, err
 	}
+	vol := resp.GetVolume()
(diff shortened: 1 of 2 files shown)


api/server/sdk/volume_node_ops.go
```diff
@@ -151,33 +151,44 @@ func (s *VolumeServer) Mount(
 	}
 
 	// Get access rights
-	if err := s.checkAccessForVolumeId(ctx, req.GetVolumeId()); err != nil {
-		return nil, err
-	}
-
-	vols, err := s.driver().Inspect([]string{req.VolumeId})
+	// Checks for ownership
```
ggriffiths: Are we only checking for ownership now? And this is done implicitly via an inspect? Instead of the additional `s.checkAccessForVolumeId`

lpabon: checkAccessForVolumeId calls inspect for access, but now this call needs to call inspect directly to get access to the response and get the Volume object.

lpabon:
Is included in the PR #851

PR #851, opened 2019-01-28
Ownership support with access types

### LINKED ISSUE #838: Allow permission when ownership.owner is empty

### LINKED ISSUE #839: Do not allow non-owners to delete the volume

### LINKED ISSUE #847: sdk: Scaling not working in docker

lpabon:
This model of ownership provides a richer set of acls for access. Without this change, trying to solve #839 would have made it only a change for that one call. With this change now, we can set access for any call and the owner also has more control.

Closes #838
Closes #839
Closes #847
Replaces #844 

* Also has other fixes:
    * Clones can set ownership of new caller
    * Scaled volumes fixed


api/api.go
```diff
@@ -947,29 +947,58 @@ func (l *VolumeLocator) MergeVolumeSpecLabels(s *VolumeSpec) *VolumeLocator {
...
+	} else {
+		// There is no user information in the context so
+		// authorization is not running
+		return true
+	}
 }
 
-func (v *VolumeSpec) IsPermittedFromUserInfo(user *auth.UserInfo) bool {
+func (v *VolumeSpec) IsPermittedFromUserInfo(user *auth.UserInfo, accessType Ownership_AccessType) bool {
+	if v.IsPublic() {
```
ggriffiths: Redundant with `v.IsPublic` check in `IsPermitted`, but maybe safe to leave in if `IsPermittedFromUserInfo` is called directly? Though `FromUserInfo` implies that it might not check for volume info.

api/api.go
```diff
@@ -947,29 +947,58 @@ func (l *VolumeLocator) MergeVolumeSpecLabels(s *VolumeSpec) *VolumeLocator {
...
 	return true
 }
+
+func (v *VolumeSpec) IsPublic() bool {
+	return v.GetOwnership() == nil || v.GetOwnership().IsPublic()
+}
+
+// GetSnapshotCreatorOwnership returns the appropriate ownership for the
+// new snapshot and if an update is required
+func (v *VolumeSpec) GetCloneCreatorOwnership(ctx context.Context) (*Ownership, bool) {
```
ggriffiths: nit: should read GetCloneCreatorOwnership returns...

api/api.proto
```diff
@@ -1534,6 +1580,8 @@ service OpenStorageCloudBackup {
   // Creates a backup request for a specified volume. Use
   // OpenStorageCloudBackup.Status() to get the current status of the
   // backup request.
+  //
+  // Requires access AccessType.Read of volume
```
ggriffiths: Should this say Create needs `AccessType.Write`?

api/api.proto
```diff
@@ -1233,6 +1265,8 @@ service OpenStorageVolume {
   // SnapshotCreate creates a snapshot of a volume. This creates an immutable (read-only),
   // point-in-time snapshot of a volume. To create a new writable volume from
   // a snapshot, please use OpenStorageVolume.Clone().
+  //
+  // Requires access AccessType.Read of volume
```
ggriffiths: Wouldn't snapshot create also need AccessType.Write?

api/ownership.go
```diff
@@ -189,8 +198,13 @@ func (o *Ownership) Update(newownerInfo *Ownership, user *auth.UserInfo) error {
 		o.Acls = newownerInfo.GetAcls()
 
 		// Only the admin can change the owner
-		if o.IsAdminByUser(user) && len(newownerInfo.Owner) != 0 {
-			o.Owner = newownerInfo.Owner
+		if len(newownerInfo.Owner) != 0 {
```
ggriffiths: Could just use `o.HasAnOwner()` here to be more clear

api/api.proto
ggriffiths: nvm, just saw below that Snapshots just need read access.

api/api.proto
ggriffiths: Nvm, same as above.

docs/ownership.md
```diff
@@ -0,0 +1,30 @@
...
+* User creates a clone from a volume which they do not own, but have access to
+* Ownership information is carried along snapshots
+
+### Volume API access types
+The following rules apply to access levels.
+
+* Volume owners have admin level automatically
+* Only owners or admin can edit acls
+
+The following access types are required to access the shown Volume service
```
ggriffiths: :+1: Double checked that these permissions documented match the implementation.

api/api.go
lpabon: > IsPermittedFromUserInfo is called directly?

yeah exactly. since the function is exported, someone can call that without checking first if it is public.

api/api.go
lpabon: doh! 👍

api/ownership.go
lpabon: I like that, good idea. I'll make the change 👍

PR #854, opened 2019-01-30
PWX-7048: Fix top-mountpoint deletion

zoxpx:
* adding convenient util-functions IsSameFilesystem() and IsMountpoint() that
  that use low-level stat and unix device-IDs to determine if files belong to
  the same file-system
* fixing mount.removeMountPath() to make sure we are not deleting STILL MOUNTED mountpoints/binds

The current code can remove MOUNTED directories, breaking the container configuration:
1. start container with "-v /mnt:/mnt:shared"
2. mount storage-volume inside container _directly_ on /mnt
3. unmount volume
4. currently, the container internally remove its /mnt directory (which is a bind-mountpoint from host !!) - so the steps 1..3 are NOT REPEATABLE, and they should be

This PR fixes this behavior.

PWX-7048

Will also require back-port (cherry-pick) for px-2.1

### DIFF
--- a/csi/node_test.go
+++ b/csi/node_test.go
@@ -45,7 +45,7 @@ func TestNodePublishVolumeBadArguments(t *testing.T) {
 	}{
 		{
 			expectedErrorContains: "Volume id",
-			req:                   &csi.NodePublishVolumeRequest{},
+			req: &csi.NodePublishVolumeRequest{},
 		},
 		{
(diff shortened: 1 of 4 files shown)


zoxpx:
This PR had ridiculous amount of build-problems -- I'm giving up on it.

Closing this PR, and reopening as squash-merged https://github.com/libopenstorage/openstorage/pull/855

PR #855, opened 2019-01-30
PWX-7048: Fix top-mountpoint deletion

zoxpx:
* adding convenient util-functions IsSameFilesystem() and IsMountpoint() that
  that use low-level stat and unix device-IDs to determine if files belong to
  the same file-system
* fixing mount.removeMountPath() to make sure we are not deleting STILL MOUNTED mountpoints/binds

The current code can remove MOUNTED directories, breaking the container configuration:
1. start container with "-v /mnt:/mnt:shared"
2. mount storage-volume inside container _directly_ on /mnt
3. unmount volume
4. currently, the container internally remove its /mnt directory (which is a bind-mountpoint from host !!) - so the steps 1..3 are NOT REPEATABLE, and they should be

This PR fixes this behavior.

PWX-7048

Will also require back-port (cherry-pick) for px-2.1

### DIFF
--- a/pkg/util/fileutil.go
+++ b/pkg/util/fileutil.go
@@ -1,0 +1,54 @@
+/*
+Package util provides utility functions for OSD servers and drivers.
+Copyright 2017 Portworx
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package util
+
+import (
+	"fmt"
+	"path"
+	"syscall"
+)
+
+// IsSameFilesystem takes a group of files/directories, and returns TRUE if all files/dirs belong to the same file-system.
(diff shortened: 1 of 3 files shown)


zoxpx:
Hey @lpabon -- how do I fix the `Failed to connect to gRPC handler: Connection timed out` travis-build failure?
Looks like I triggered something unrelated to my changes, and can't get past it.  Already wasted more time on trying to make travis-build work, than on fixing the original bug.

lpabon:
>Hey @lpabon -- how do I fix the Failed to connect to gRPC handler: Connection timed out travis-build failure?
Looks like I triggered something unrelated to my changes, and can't get past it. Already wasted more time on trying to make travis-build work, than on fixing the original bug.

Is a known issue. We have to fix it in the unit tests.

pkg/util/fileutil.go
```diff
@@ -0,0 +1,54 @@
+/*
+Package util provides utility functions for OSD servers and drivers.
+Copyright 2017 Portworx
```
lpabon: 2017? You had these files just sitting around for two years ;-)

zoxpx: fixed in 846b50e

rodrigc:
Tested here:

* https://jenkins.portworx.co/blue/organizations/jenkins/CLI_TESTS%2FPX-TEST-CLI/detail/PX-TEST-CLI/543/pipeline#step-48-log-756
* https://jenkins.portworx.co/blue/organizations/jenkins/CLI_TESTS%2FPX-TEST-CLI/detail/PX-TEST-CLI/543/tests

PR #856, opened 2019-01-30
Additional fields to cloud-backup data structure to track group cloud…

talakad:
… backups.

Makes it easier for client to track status of group cloud backups.

Closes #


api/api.go
```diff
@@ -384,7 +386,8 @@ type CloudBackupStatusRequest struct {
 	// active on current node must be returned
 	Local bool
 	// Name of the backup/restore task. If this is specified, SrcVolumeID is
-	// ignored
+	// ignored. This could be GroupCloudBackupId too, and in that case multiple
+	// statuses belonging to the groupCloudBackupID is returned.
 	Name string
```
disrani-px: Maybe have a GroupName instead of overloading this name?

talakad: Not sure if adding the new field GroupName makes it simpler? client to specify either task Name or groupID in same field, removing the confusion. Adding new field would also mean status for combination of groupID-taskName.

disrani-px: Can we at least rename this from `Name` to `ID` like it originally was then to make is clearer? (while maintaining backward compatibility) We pass in `Name` here but status is returning `ID`

talakad: Yes. Changed it to ID while retaining the backward compatibility for clients using older api.

api/api.go
```diff
@@ -317,6 +317,8 @@ type CloudBackupRestoreRequest struct {
 }
 
 type CloudBackupGroupCreateResponse struct {
+	// ID for this group of backups
+	GroupCloudBackupID string
 	// Names of the tasks performing this group backup
 	Names []string
```
disrani-px: Change this to IDs too?

talakad: AFAIK, NO. This will need capability on clients side to interpret both old and new response structures( just like we do on server side for requests)

PR #857, opened 2019-01-31
Add storage policy support - release-5.0

ram-infrac:
This PR adds storage policy support to sdk. Storage Policy will allow to set/reset enforcement of volume configuration to be followed while creating volumes.

Closes #


api/api.proto
```diff
@@ -1573,6 +1577,139 @@ service OpenStorageCloudBackup {
...
+  // Deletes specified storage policy
+  rpc Delete(SdkOpenStoragePolicyDeleteRequest)
+    returns (SdkOpenStoragePolicyDeleteResponse) {
+      option(google.api.http) = {
+        delete: "/v1/storagepolicy/{name}"
+      };
+    }
+
+  // Enforces specified storage policy
+  rpc Enforce(SdkOpenStoragePolicyEnforceRequest)
```
ram-infrac: please suggest better name for this if any. I would like to address proto comments in one go, so that can it will be easy to push auto-generated files

api/api.proto
```diff
@@ -370,6 +372,10 @@ message VolumeSpecUpdate {
   oneof sharedv4_opt { bool sharedv4 = 24; }
   // QueueDepth defines the desired block device queue depth
   oneof queue_depth_opt { uint32 queue_depth = 25; }
+  // Encrypted is true if this volume will be cryptographically secured.
+  oneof encrypted_opt { bool encrypted = 26; }
```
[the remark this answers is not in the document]
lpabon: We had a conversation offline, and we will be creating VolumeSpecPolicy as a new message that will house values that need to override the spec. We cannot use VolumeSpecUpdate because these values are the ones that support being updated only.

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
     }
 }
 
+// START: OpenStoragePolicy Service
```
lpabon: Remove this

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+
+// OpenStoragePolicy service provides methods to manager storage policies.
```
lpabon: This needs to be on top of the service for the doc generator to work. See other services

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+      option(google.api.http) = {
+        delete: "/v1/storagepolicy/{name}"
+      };
+    }
+
+  // Enforces specified storage policy
+  rpc Enforce(SdkOpenStoragePolicyEnforceRequest)
+    returns (SdkOpenStoragePolicyEnforceResponse) {
+      option(google.api.http) = {
+        put: "/v1/storagepolicy/enforce"
```
lpabon: Needs to be like this:

```
post: "/v1/storagepolicy/enforce/{name}
```

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+      };
+    }
+
+  // Release specified storage policy constraint for volume
+  // creation
+  rpc Release(SdkOpenStoragePolicyReleaseRequest)
+    returns (SdkOpenStoragePolicyReleaseResponse) {
+      option(google.api.http) = {
+        put: "/v1/storagepolicy/release"
```
lpabon: Change to:

```
post : "/v1/storagepolicies/release/{name}
```

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+        put: "/v1/storagepolicies"
+        body: "*"
+      };
+    }
+
+  // Deletes specified storage policy
+  rpc Delete(SdkOpenStoragePolicyDeleteRequest)
+    returns (SdkOpenStoragePolicyDeleteResponse) {
+      option(google.api.http) = {
+        delete: "/v1/storagepolicy/{name}"
```
lpabon: Change to storage policies

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+  // creation
+  rpc Release(SdkOpenStoragePolicyReleaseRequest)
+    returns (SdkOpenStoragePolicyReleaseResponse) {
+      option(google.api.http) = {
+        put: "/v1/storagepolicy/release"
+      };
+    }
+}
+
+//   START - OpenStoragePolicy structs
```
lpabon: Remove this line. It makes the doc generator confused. We use the doc generator to make these webpages:

https://libopenstorage.github.io/w/master.generated-api.html

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+// Define a request to create storage policy
+message SdkOpenStoragePolicyCreateRequest {
+  SdkStoragePolicy storage_policy = 1;
+}
+
+// Empty response
+message SdkOpenStoragePolicyCreateResponse {
+}
+
+message SdkOpenStoragePolicyEnumerateRequest {
```
lpabon: Must write `Empty response` for empty responses, and `Empty request` for empty requests

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+}
+
+message SdkOpenStoragePolicyEnumerateResponse {
+  repeated SdkStoragePolicy storage_policies =1;
```
lpabon: Needs documentation

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+}
+
+// Empty response
+message SdkOpenStoragePolicyCreateResponse {
+}
+
+message SdkOpenStoragePolicyEnumerateRequest {
+}
+
+message SdkOpenStoragePolicyEnumerateResponse {
```
lpabon: Needs documentation

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+  repeated SdkStoragePolicy storage_policies =1;
+}
+
+message SdkOpenStoragePolicyInspectRequest {
```
lpabon: All of these need documentation

api/api.proto
```diff
@@ -1573,6 +1583,139 @@ service OpenStorageCloudBackup {
...
+message SdkOpenStoragePolicyEnforceResponse {
+}
+
+message SdkOpenStoragePolicyReleaseRequest {
+}
+
+message SdkOpenStoragePolicyReleaseResponse {
+}
+
+//  END - OpenStoragePolicy structs
```
lpabon: Remove these `END` comments

api/server/sdk/sdk_test.go
```diff
@@ -92,7 +93,17 @@ func newTestServer(t *testing.T) *testServer {
...
-	var err error
+	kv, err := kvdb.New(mem.Name, "policy", []string{}, nil, logrus.Panicf)
+	assert.NoError(t, err)
+	// Init storage policy manager
+	err = policy.Init(kv)
+	sp, err := policy.Inst()
+	assert.NoError(t, err)
+	assert.NotNil(t, sp)
+	// disable policy enforcement by default
+	_, err = sp.Release(context.Background(), &api.SdkOpenStoragePolicyReleaseRequest{})
```
lpabon: Why release when it was just created? The kvdb is brand new, right?

api/server/sdk/volume_ops.go
```diff
@@ -115,11 +116,17 @@ func (s *VolumeServer) Create(
...
 
-	spec := req.GetSpec()
 	locator := &api.VolumeLocator{
 		Name: req.GetName(),
 	}
 	source := &api.Source{}
+	// Update given spec according to enforcement policy set
+	// In case policy is not set, should fall back to default way
+	// of creating volume
+	spec, err := GetEnforcedVolSpecs(locator, source, req.GetSpec())
```
lpabon: I thought this was going to be in the driver. If it is here, that is find. Just add to the Docker API and CSI API also.

api/server/sdk/volume_ops.go
```diff
@@ -491,3 +506,49 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
 	}
 	return current
 }
+
+func GetEnforcedVolSpecs(locator *api.VolumeLocator, source *api.Source, spec *api.VolumeSpec) (*api.VolumeSpec, error) {
```
lpabon: Why do we need source? Remove if we do not need.

lpabon: Does this function need to be exported?

api/server/sdk/volume_ops.go
```diff
@@ -491,3 +506,49 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+	if locator != nil {
+		if val, ok := locator.VolumeLabels["enforce_disable"]; ok {
```
lpabon: Nice. Add comment here so that others in the future who read this understand why this is needed.

api/server/sdk/volume_ops.go
```diff
@@ -491,3 +506,49 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+	}
+
+	// check if custom policy passed with volume
+	if spec.GetStoragePolicy() != "" {
+		inspReq := &api.SdkOpenStoragePolicyInspectRequest{
+			Name: spec.GetStoragePolicy(),
+		}
+		customPolicy, err := storPolicy.Inspect(context.Background(), inspReq)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Failed to get %s policy details :%v", inspReq.Name, err)
```
lpabon: Don't change the code that was sent by storPolicy.Inspect(). It has already figured out the problem.

Just return err.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	req *api.SdkOpenStoragePolicyInspectRequest,
+) (*api.SdkOpenStoragePolicyInspectResponse, error) {
+	if req.GetName() == "" {
+		return nil, status.Error(codes.InvalidArgument, "Must supply a Storage Policy Name")
+	}
+
+	var volSpecs *api.VolumeSpecUpdate
+	kvp, err := p.kv.GetVal(prefixWithName(req.GetName()), &volSpecs)
+	if err == kvdb.ErrNotFound {
+		return nil, status.Errorf(codes.NotFound, "Policy %s not found", req.GetName())
```
lpabon: Good.

api/server/sdk/volume_ops.go
```diff
@@ -491,3 +506,49 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+		}
+		// update spec with custom storage policy
+		updatedSpec := mergeVolumeSpecs(spec, customPolicy.GetStoragePolicy().GetPolicy())
+		return updatedSpec, nil
+	}
+
+	// apply default policy to volume
```
lpabon: This is interesting. Please send a note to @jvinod and ask, if a user enforces a policy, I'm thinking that it will apply to ALL requests, no matter what they choose, right?

api/server/sdk/volume_ops.go
```diff
@@ -491,3 +506,49 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+	policy, err := storPolicy.GetEnforcement()
+	if err == kvdb.ErrNotFound {
+		return spec, nil
+	}
+	// err means there is policy stored, but we are not able to retrive it
+	// hence we are not allowing volume create operation
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Unable to get enforced policy details %v", err)
+	}
+	if policy.GetName() == "" || policy.GetPolicy() == nil {
```
lpabon: Can we make the GetEnforcement() call check these things?

api/server/volume.go
```diff
@@ -165,7 +166,13 @@ func (vd *volAPI) create(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	id, err := d.Create(dcReq.Locator, dcReq.Source, dcReq.Spec)
+	spec, err := sdkVol.GetEnforcedVolSpecs(dcReq.Locator, dcReq.Source, dcReq.Spec)
```
lpabon: I thought this was going to be in the driver. If it is here, that is find. Just add to the Docker API and CSI API also.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
+package storagepolicy
```
lpabon: Needs Apache license header. Same with all new files

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	policyPrefix = "/storage/policy"
+	policyPath   = "/policies"
+	EnforcePath  = "/storage/policy/enforce"
+)
+
+var (
+	// Check interface
+	_ PolicyManager = &SdkPolicyManager{}
+
+	errPolicyManagerNotInitialized = errors.New("Policy Manager is not initialized")
```
lpabon: Why have these if they are not even exported? Just use fmt.Errorf() in the appropriate return.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	errPolicyManagerInitialized    = errors.New("Policy Manager is already initialized")
+
+	inst *SdkPolicyManager
+	Inst = func() (PolicyManager, error) {
+		return policyInst()
+	}
+)
+
+func Init(kv kvdb.Kvdb) error {
```
lpabon: might as well return here the value of `inst`. That way the caller does not have to make two calls to initialize then use.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+// SdkPolicyManager is an implementation of the
+// Storage Policy Manager for the SDK
+type SdkPolicyManager struct {
+	kv kvdb.Kvdb
+}
+
+const (
+	policyPrefix = "/storage/policy"
+	policyPath   = "/policies"
+	EnforcePath  = "/storage/policy/enforce"
```
lpabon: why is this exported? This should not be exported.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	return policyPrefix + policyPath + "/" + name
+}
+
+func (p *SdkPolicyManager) Create(
+	ctx context.Context,
+	req *api.SdkOpenStoragePolicyCreateRequest,
+) (*api.SdkOpenStoragePolicyCreateResponse, error) {
+	if req.StoragePolicy.GetName() == "" {
+		return nil, status.Error(codes.InvalidArgument, "Must supply a Storage Policy Name")
+	}
```
lpabon: for readability, use: if , else if, 

that puts these checks together.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	return inst, nil
+}
+
+// Simple function which creates key for Kvdb
+func prefixWithName(name string) string {
+	return policyPrefix + policyPath + "/" + name
+}
+
+func (p *SdkPolicyManager) Create(
```
lpabon: All of the exposed functions need simple comments

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+		&api.SdkOpenStoragePolicyInspectRequest{
+			Name: req.GetName(),
+		},
+	)
+	if err != nil {
+		return nil, status.Errorf(codes.NotFound, "Policy with name %s not found", req.GetName())
+	}
+
+	m := jsonpb.Marshaler{}
+	policyStr, err := m.MarshalToString(policy.StoragePolicy)
```
lpabon: We should not place the spec in Enforcement since this is a copy. If the original is then changed, this version still has the old information.

Instead this should only save the name of the policy as a pointer to the actual policy managed by this system. That way when it is updated, then GetEnforce..() is called, it gets the latest changes.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	// empty represents no policy enforcement is enabled
+	_, err := p.kv.Update(EnforcePath, api.SdkStoragePolicy{}, 0)
+	if err != kvdb.ErrNotFound && err != nil {
+		return nil, status.Errorf(codes.Internal, "Disable enforcement failed with: %v", err)
+	}
+
+	return &api.SdkOpenStoragePolicyReleaseResponse{}, nil
+}
+
+func (p *SdkPolicyManager) GetEnforcement() (*api.SdkStoragePolicy, error) {
```
lpabon: Here, this should get the -name- of the policy, then do an inspect on that name

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -0,0 +1,263 @@
...
+	if err == kvdb.ErrNotFound {
+		return nil, status.Errorf(codes.NotFound, "Storage Policy %s not found", req.StoragePolicy.GetPolicy())
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "Failed to update storage policy: %v", err)
+	}
+
+	return &api.SdkOpenStoragePolicyUpdateResponse{}, nil
+}
+
+func (p *SdkPolicyManager) Delete(
```
lpabon: If the name being deleted is also enforced, then this should call the Release() function.

lpabon:
Also, the SDK version patch number needs to be updated as well as the SDK_CHANGELOG.md

api/api.proto
```diff
@@ -476,6 +520,16 @@ message CapacityUsageInfo {
   int64 total_bytes = 3;
 }
 
+// A SdkStoragePolicy represent minimum set of volume specs to be follow while creating volumes.
```
lpabon: "represents"

api/api.proto
```diff
@@ -476,6 +520,16 @@ message CapacityUsageInfo {
...
+// If enabled, VolumeSpecUpdate will be used before creating volume to ensure minimum volume creation rules
```
lpabon: "If enforce in OpenStoragePolicy service, VolumeSpecPolicy..."

api/api.proto
```diff
@@ -1573,6 +1621,151 @@ service OpenStorageCloudBackup {
     }
 }
 
+// OpenStoragePolicy service provides methods to manager storage policies.
+//
+// Policies can be used to enforces a set of volume configurations to be followed
+// while creating volumes.
+service OpenStoragePolicy {
+  // TODO: storage policy owenership
```
lpabon: Remove

api/api.proto
```diff
@@ -1573,6 +1621,151 @@ service OpenStorageCloudBackup {
...
+        post: "/v1/storagepolicy/enforce/{name}"
+      };
+    }
+
+  // Release specified storage policy constraint for volume
+  // creation
+  rpc Release(SdkOpenStoragePolicyReleaseRequest)
+    returns (SdkOpenStoragePolicyReleaseResponse) {
+      option(google.api.http) = {
+        post: "/v1/storagepolicy/release"
```
lpabon: /v1/storagepolicies/release

lpabon: And you need to add the `body: "*"`

api/spec/spec_handler.go
```diff
@@ -447,6 +448,9 @@ func (d *specHandler) SpecOptsFromString(
 	if ok, forceUnsupportedFsType := d.getVal(forceUnsupportedFsTypeRegex, str); ok {
 		opts[api.SpecForceUnsupportedFsType] = forceUnsupportedFsType
 	}
+	if ok, storagepolicy := d.getVal(storagePolicyRegex, str); ok {
```
lpabon: Don't you also have to add it to the case statement? I may be wrong.

api/api.proto
```diff
@@ -476,6 +520,18 @@ message CapacityUsageInfo {
...
+// A SdkStoragePolicy represents minimum set of volume specs to be 
+// follow while creating volumes.
+// If enforce in OpenStoragePolicy service, VolumeSpecPolicy will be 
+// used before creating volume to ensure minimum volume creation rules
+// are enforced.
+message SdkStoragePolicy {
+  // Name of storage policy.
+  string name = 1;
+  // VolumeSpecs to apply while creating volume.
+  VolumeSpecPolicy policy = 2;
```
ram-infrac: @lpabon , should we also include **IsEnforced** here. So that user with enumerate or inspect call will know policy is enforced or not. That way they don't need to make another call for getting enforcement details ?

lpabon: I think you have a great idea, but if we do it with IsEnforced, they would have to enumerate and search each. Instead, let's make a new API called GetEnforced (which is the one your already have implemented). That way, they can call that to verify which one is enforced.

api/api.proto
```diff
@@ -2597,7 +2814,7 @@ message SdkVersion {
     // SDK version minor value of this specification
     Minor = 34;
```
lpabon: This seems to be incorrect. According to the SDK CHANGELOG, this is supposed to be `33`, but I see `34`. It is supposed to be 0.33.1, right?

api/server/sdk/volume_ops.go
```diff
@@ -393,7 +400,7 @@ func (s *VolumeServer) CapacityUsage(
 	return resp, nil
 }
 
-func (s *VolumeServer) mergeVolumeSpecs(vol *api.VolumeSpec, req *api.VolumeSpecUpdate) *api.VolumeSpec {
+func (s *VolumeServer) mergeVolumeSpecsUpdate(vol *api.VolumeSpec, req *api.VolumeSpecUpdate) *api.VolumeSpec {
```
lpabon: why change the name here?

cli/cli_test.go
```diff
@@ -39,7 +39,8 @@ func TestCmdMarshalProto(t *testing.T) {
  "sharedv4": false,
  "queue_depth": 0,
  "force_unsupported_fs_type": false,
- "nodiscard": false
+ "nodiscard": false,
+ "storage_policy": ""
```
lpabon: Is this supposed to be equal to https://github.com/libopenstorage/openstorage/pull/857/files#diff-604f4192deea5adf9c159e5f88f64630R59 ?

api/server/sdk/volume_ops.go
ram-infrac: Nope, this shouldn't. good catch.

cli/cli_test.go
ram-infrac: that's json name if you see `"force_unsupported_fs_type"` at https://github.com/libopenstorage/openstorage/pull/857/files#diff-604f4192deea5adf9c159e5f88f64630R58 
it's defined here api.pb.go  https://github.com/libopenstorage/openstorage/blob/5a0f8125e94f6b045d5a702cce1669c51e704585/api/api.pb.go#L1534

api/server/sdk/volume_ops.go
```diff
@@ -623,11 +662,21 @@ func mergeVolumeSpecsPolicy(vol *api.VolumeSpec, req *api.VolumeSpecPolicy) *api
...
-		spec.HaLevel = vol.GetHaLevel()
+	spec.HaLevel = vol.GetHaLevel()
+	if req.GetHaLevelOpt() != nil {
+		switch req.GetHaLevelOperator() {
+		case api.VolumeSpecPolicy_Maximum:
+			if vol.GetHaLevel() > req.GetHaLevel() {
+				spec.HaLevel = req.GetHaLevel()
+			}
+		case api.VolumeSpecPolicy_Minimum:
+			if vol.GetHaLevel() < req.GetHaLevel() {
```
ram-infrac: shouldn't it keep HA level as it is if above spec policy min HA level
line no 627

PR #858, opened 2019-01-31
Fix goroutine creation bug and add exponential back-off logic

pault84:

It fixes an issue where we would be spinning up goroutines that stayed open until the first request got closed. Also added Exponential back-off to limit the requests hitting the server.


pault84:
> this is the client code - shouldn't there be something at the server to close idle connections from erroneous clients?

Will make a change there too I just didn't want to touch timeouts that we have never had. Will look at idleConnTimeout

PR #864, opened 2019-02-01
Let users with admin access type change ownership

### LINKED ISSUE #862: Should only allow Admin access type to update ownership
BUG

**What happened**:
A collaborator/group can raise their access type level by changing the ownership. To prevent this, only users with Admin access type should be allowed to change ownership of a resource.

### LINKED ISSUE #863: Ownership should not be changed if the update failed
BUG
**What happened**:
Even though the update returned an error for trying to update the owner, it was still changing the ACLs

When the update fails, the original ownership should remain

lpabon:
Updating the ownership could only be done by the owner or cluster admin, even though a collaborator may have admin access type rights. This PR adds support for collaborators or groups with admin access type rights to also update the ownership. These users cannot change the owner, so the owner still has more rights, in that they can remove access to these collaborator/groups.

Also fixes #863 were the ownership was being changed even though it returned an error.

Closes #862
Closes #863

### DIFF
--- a/api/ownership.go
+++ b/api/ownership.go
@@ -190,12 +190,14 @@ func (o *Ownership) Update(newownerInfo *Ownership, user *auth.UserInfo) error {
 	} else {
 		// Auth is enabled
 
-		// Only the owner or admin can change the group
-		if user.Username != o.Owner && !o.IsAdminByUser(user) {
+		// Only the owner, user with access type admin,
+		// or admin can change the group
+		if user.Username != o.Owner &&
+			!o.IsAdminByUser(user) &&
+			!o.IsPermitted(user, Ownership_Admin) {
 			return status.Error(codes.PermissionDenied,
-				"Only owner can update volume acls")
+				"Only owner or those with admin access type can update volume acls")
 		}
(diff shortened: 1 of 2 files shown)


api/ownership.go
```diff
@@ -190,12 +190,14 @@ func (o *Ownership) Update(newownerInfo *Ownership, user *auth.UserInfo) error {
...
-		// Only the owner or admin can change the group
-		if user.Username != o.Owner && !o.IsAdminByUser(user) {
+		// Only the owner, user with access type admin,
+		// or admin can change the group
+		if user.Username != o.Owner &&
+			!o.IsAdminByUser(user) &&
+			!o.IsPermitted(user, Ownership_Admin) {
 			return status.Error(codes.PermissionDenied,
-				"Only owner can update volume acls")
+				"Only owner or those with admin access type update volume acls")
```
ggriffiths: nit: maybe add `can` before update to make this error a little more clear

Only owner or those with admin access type **can** update volume acls

lpabon: sounds good!

PR #870, opened 2019-02-05
Add method to check if a volume spec is equal to another

harsh-px:
Reopening previously closed PR https://github.com/libopenstorage/openstorage/pull/822 after discussion with @disrani-px . A dedicate equals method is better to compare individual items of the spec. Using something generic like reflect.DeepEquals creates issues when drivers change certain spec fields after the user specifies them.

Closes #

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math"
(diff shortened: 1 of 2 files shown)


harsh-px:
We will be doing the check in portworx itself

PR #874, opened 2019-02-06
don't fail health checks when out of quorum

harsh-px:

Closes #


harsh-px:
Instead of corrupting existing health checks, a different approach has been decided.

PR #879, opened 2019-02-11
Enable auth for cloud pairing and move system token generator

### LINKED ISSUE #865: Add SystemTokenGenerator to pkg/auth

### LINKED ISSUE #872: Support auth for cloud pairing

ggriffiths:

- GetPairToken now gets a system token if auth is enabled, otherwise it returns a randomly generated token as before.
- We get this token in `CreatePair` and store it as a secret for subsequent calls.

Closes #872 
Closes #865 


cluster/manager/pair.go
```diff
@@ -26,9 +30,29 @@ func (c *ClusterManager) CreatePair(
...
+		// Check if valid jwt token
+		if !auth.IsJwtToken(request.RemoteClusterToken) {
+			logrus.Warnf("RemoteClusterToken is not a valid JWT token")
+			return nil, nil
+		}
+
+		// Check for existing token. Store if new token is passed in
+		// or if there is no token stored yet at all.
+		tokenSecretKey := clusterTokenSecretKey(remoteIp)
+		token, err := c.SecretGet(tokenSecretKey)
```
ggriffiths: Simplify to `SecretSet`

cluster/manager/pair.go
```diff
@@ -26,9 +30,29 @@ func (c *ClusterManager) CreatePair(
 ) (*api.ClusterPairCreateResponse, error) {
 	remoteIp := request.RemoteClusterIp
 
+	if auth.Enabled() {
```
ggriffiths: Only check for auth on remote cluster, not this client.

cluster/manager/pair.go
```diff
@@ -26,9 +30,29 @@ func (c *ClusterManager) CreatePair(
...
+		// Check if valid jwt token
+		if !auth.IsJwtToken(request.RemoteClusterToken) {
```
ggriffiths: No need to check for this

cluster/manager/pair.go
```diff
@@ -317,6 +342,41 @@ func (c *ClusterManager) GetPairToken(
 	}, nil
 }
 
+func (c *ClusterManager) generatePairToken() ([]byte, error) {
+	var token []byte
+	var systemSecret string
+
+	if auth.Enabled() {
+		generator, err := auth.Init(&auth.Config{
```
ggriffiths: Switch this to have `SystemTokenManager`, `SharedSecretTokenGenerator` and then `TokenGenerator.Generate`

cluster/manager/pair.go
```diff
@@ -317,6 +342,41 @@ func (c *ClusterManager) GetPairToken(
 	}, nil
 }
 
+func (c *ClusterManager) generatePairToken() ([]byte, error) {
+	var token []byte
+	var systemSecret string
```
ggriffiths: Add this to clusterConfig

ggriffiths:
Reviewed with @lpabon, reworking the token generator logic.

cluster/manager/manager.go
```diff
@@ -52,26 +53,27 @@ var (
...
+	gossipPort         string
+	gEnabled           bool
+	selfNode           api.Node
+	selfNodeLock       sync.Mutex // Lock that guards data and label of selfNode
+	system             systemutils.System
+	configManager      osdconfig.ConfigCaller
+	schedManager       sched.SchedulePolicyProvider
+	objstoreManager    objectstore.ObjectStore
+	secretsManager     secrets.Secrets
+	systemTokenManager auth.TokenGenerator
```
ggriffiths: This is the only field I've added.

pkg/auth/selfsigned-generate.go
```diff
@@ -1,55 +0,0 @@
-/*
```
ggriffiths: This file is now in token.go

pkg/auth/auth.go
```diff
@@ -64,9 +67,13 @@ func getUsername(usernameClaim UsernameClaimType, claims *Claims) string {
 	return claims.Subject
 }
 
-// AuthorizationEnabled returns true if the storage system has authorization
-// and authentication enabled.
-func AuthorizationEnabled(ctx context.Context) bool {
-	_, ok := NewUserInfoFromContext(ctx)
-	return ok
+// Enabled returns whether or not auth is enabled.
+func Enabled() bool {
```
ggriffiths: I wanted to keep `Enabled` inside of the `auth` package,  so I needed to introduce this flag for setting when auth is enabled or disabled. 

Before, we checked the SystemToken instance to if there was an issuer register. If auth was disabled, the issuer was empty. This method will not work anymore because I've created a new `systemtoken` package  that depends on `auth`. The old way of doing this would cause a circular dependency.

cluster/manager/manager.go
```diff
@@ -1887,17 +1894,17 @@ func (c *ClusterManager) SecretGetDefaultSecretKey() (interface{}, error) {
 
 // SecretCheckLogin validates session with secret store
 func (c *ClusterManager) SecretCheckLogin() error {
-	return c.SecretCheckLogin()
+	return c.secretsManager.SecretCheckLogin()
```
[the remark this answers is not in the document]
ggriffiths: When I was trying to use the secretsManager, I noticed that these functions would just keep recursively calling each other. This would cause another bug unless I'm misunderstanding how these work.

cmd/osd/main.go
```diff
@@ -164,6 +165,10 @@ func main() {
 			Name:  "jwt-ecds-pubkey-file",
 			Usage: "JSON Web Token ECDS Public file path",
 		},
+		cli.StringFlag{
+			Name:  "jwt-system-shared-secret",
+			Usage: "JSON Web Token system shared secret",
```
lpabon: I know what this is, but other users could use a little more information. Something like "This shared secret is used by clusters to create tokens to communicate with eachother", something like that.

cmd/osd/main.go
```diff
@@ -431,6 +439,18 @@ func start(c *cli.Context) error {
 		if err != nil {
 			logrus.Fatalf("Failed to access TLS file information: %v", err)
 		}
+		if len(authenticators) > 0 {
+			// Initialize system token manager if an authenticator is setup
```
lpabon: probably want to do line 268 here.

either way, there must be a check for jwt-system-shared-secret not to be empty when oidc or share-secret has been set.

pkg/auth/tokengenerator.go
```diff
@@ -0,0 +1,44 @@
+package auth
```
lpabon: need apache license header

pkg/auth/tokengenerator.go
```diff
@@ -0,0 +1,44 @@
...
+	// GetToken returns a token which can be used for
+	// authentication and communication from node to node.
+	GetToken(opts *Options) (string, error)
+
+	// Issuer returns the token issuer for this generator necessary
+	// for registering the authenticator in the SDK.
+	Issuer() string
+
+	// GetAuthenticator returns an authenticator for this issuer used by the SDK
+	GetAuthenticator() (*JwtAuthenticator, error)
```
lpabon: I thought the return value should be auth.Authenticator, right?

cluster/manager/manager.go
```diff
@@ -1170,12 +1173,11 @@ func (c *ClusterManager) setupManagers(config *cluster.ClusterServerConfiguratio
 		c.objstoreManager = config.ConfigObjectStoreManager
 	}
 
-	if config.ConfigSecretManager == nil {
```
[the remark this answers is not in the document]
ggriffiths: Good catching, adding back in

pkg/auth/tokengenerator.go
ggriffiths: This has always been JwtAuthenticator:
https://github.com/portworx/porx/blob/sdk-master/pkg/auth/auth.go#L16

PR #883, opened 2019-02-13
Update snapshot API to take in a spec

harsh-px:
The existing snapshot API isn't sufficient if we want to add additional parameters for the snapshot like it's placement strategy. This PR adds a formal snapshot spec to the call. This is backward compatible since the request and response structures for the REST call doesn't change.

Closes #


harsh-px:
> adding only placement_strategy won't work ? looks ok anyways if rest (and other) APIs remain unchanged.

I guessing you are referring to just adding a new parameter placement_strategy to the  Snapshot call.

Yes, that would work but we might again be at this crossroads when we have to add another new param to the snapshot call. So I added a dedicated spec just like we have for volume creates. It's more extensible. What do you think?

ggriffiths:

PR #892, opened 2019-02-20
sdk: Add options to mountattach service RPCs

lpabon:
When we created the MountAttach service for the SDK, we removed support for
map[string]string because it was confusing for developers to use since the
values are not very clear. What we didn't realize is that drivers may use
this payload to add unique, opaque information.

This patch adds the option for driver specific values to be passed
in during the use of MountAttach service calls.


api/server/sdk/volume_node_ops.go
```diff
@@ -47,14 +45,11 @@ func (s *VolumeServer) Attach(
 		return nil, err
 	}
 
-	// Check if already attached
-	_, err := util.VolumeFromName(s.driver(), req.GetVolumeId())
```
lpabon: This code needed to be removed in a previous PR, but it was not. So I'm cleaning it up now, since it does nothing.

api/server/sdk/volume_node_ops.go
```diff
@@ -96,20 +91,17 @@ func (s *VolumeServer) Detach(
...
```
lpabon: This code needed to be removed in a previous PR, but it was not. So I'm cleaning it up now, since it does nothing.

PR #895, opened 2019-02-22
Fix issue where Connect() would timeout intermittently

### LINKED ISSUE #894: Connect() intermittent timeout issue

**What happened**:
Connect() timeout had issues intermittently when there was some lag between grpc.Dial and the timeout check.

No timeouts

Put a sleep inbetween grpc.Dial and the timeout check

**Environment**:
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

ggriffiths:

* Fixes timeout issue with `Connect` function

Closes #894

### DIFF
--- a/pkg/grpcserver/grpcutil.go
+++ b/pkg/grpcserver/grpcutil.go
@@ -16,12 +16,12 @@ limitations under the License.
 package grpcserver
 
 import (
-	"context"
 	"fmt"
 	"net"
 	"net/url"
 	"time"
 
+	"github.com/libopenstorage/openstorage/pkg/util"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/connectivity"
 )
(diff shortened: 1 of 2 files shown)


pkg/grpcserver/grpcutil.go
```diff
@@ -43,14 +42,40 @@ func Connect(address string, dialOptions []grpc.DialOption) (*grpc.ClientConn, e
...
+	}); err != nil {
+		return nil, fmt.Errorf("Connection timed out")
 	}
+
+	return conn, nil
+}
+
+// f() return <wait as bool, or err>.
+// waitFor stops when f() return (false) or err != nil
+func waitFor(timeout time.Duration, period time.Duration, f func() (bool, error)) error {
```
lpabon: This is great, can we move this function to pkg/util/wait.go and expose it and use it here. There may be other situations we use such a function.

ggriffiths: :+1: Yep good idea

PR #906, opened 2019-03-03
SDK to ignore the retain value for cloud-backup schedules.

talakad:
Retain value for Cloud backup schedule is separate field out of schedule format.

Closes #

### DIFF
--- a/api/server/sdk/utils.go
+++ b/api/server/sdk/utils.go
@@ -88,10 +88,6 @@ func sdkSchedToRetainInternalSpec(
 	req *api.SdkSchedulePolicyInterval,
 ) (*sched.RetainIntervalSpec, error) {
 
-	if req.GetRetain() < 1 {
-		return nil, status.Error(codes.InvalidArgument, "Must retain more than 0")
-	}
-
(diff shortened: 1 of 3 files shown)


api/server/sdk/utils.go
```diff
@@ -86,9 +86,10 @@ func timeWeekdayToSdkWeekly(t time.Weekday) api.SdkTimeWeekday {
 
 func sdkSchedToRetainInternalSpec(
 	req *api.SdkSchedulePolicyInterval,
+	checkRetain bool,
 ) (*sched.RetainIntervalSpec, error) {
 
-	if req.GetRetain() < 1 {
+	if checkRetain && req.GetRetain() < 1 {
```
lpabon: instead of adding a boolean to the function and checking retain here, let's move the check of it to the caller. This will simplify the arguments, and make the caller responsible for checking the retain value.

talakad: you mean move the check to create/update function of shed_policy.go or leave it unto the client( totally remove the check)?

talakad: Done. I moved it to cloudbjckup-schedule.go.

PR #912, opened 2019-03-08
SDK: Allow registering multiple VolumeDrivers with a single SDK server.

adityadani:

- Changed the UseVolumeDrivers API to take in a map of name to VolumeDrivers mapping
- Modified the SDK server, to inspect the context's metadata and select the correct
  VolumeDriver to use for the API.
- Modified the REST server, to parse the UserAgent in the incoming request and
  set it as a metadata key-value pair in the SDK grpc call's context.

### DIFF
--- a/api/server/sdk/server.go
+++ b/api/server/sdk/server.go
@@ -17,6 +17,7 @@ limitations under the License.
 package sdk
 
 import (
+	"context"
 	"fmt"
 	"io"
 	"os"
@@ -26,7 +27,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
-	"github.com/grpc-ecosystem/go-grpc-middleware"
+	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
 
 	"github.com/libopenstorage/openstorage/alerts"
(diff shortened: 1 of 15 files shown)


api/server/sdk/server.go
```diff
@@ -343,9 +348,12 @@ func newSdkGrpcServer(config *ServerConfig) (*sdkGrpcServer, error) {
 		name:            name,
 		log:             log,
 		clusterHandler:  config.Cluster,
-		driverHandler:   d,
-		alertHandler:    config.AlertsFilterDeleter,
-		policyServer:    config.StoragePolicy,
+		driverHandlers: map[string]volume.VolumeDriver{
```
lpabon: I'm not sure I understand this. A `map[string]volume.VolumeDriver` would mean that every value would have a DefaultDriverName. What if two values have two different defaultdrivernames?

Is something like this what you meant?:

```go
type VolumeDrivers struct {
  Drivers []string
  DefaultDriverName string
}
...
driverHandlers *VolumeDrivers,
...
```

lpabon: Or is it that `DefaultDriverName` is a special key where its value is the default driver?

adityadani: We need a map[string] VolumeDrivers, because the sdk client request is going to pass in a string which indicates which VolumeDriver to use.

DefaultDriver is required so that if the context does not have any requested drivername provided, then we will use the default one which is registered.

lpabon: I get it now, thanks!

PR #923, opened 2019-03-12
Require system.admin user for cluster get pair token

### LINKED ISSUE #922: GetPairToken should require system.admin user
BUG

**What happened**:
* Any user can access system.admin tokens with GetPairToken

* Only system.admin should be able to get cluster pair tokens when auth is enabled.

ggriffiths:

* Requires system.admin user for cluster pair token

Closes #922


ggriffiths:
Tested this locally with two tokens: `system.user` and `system.admin`.

System.user was denied with the error I've implemented.
System.admin returned a JWT System Token as expected.

pkg/role/sdkserviceapi.go
```diff
@@ -74,8 +74,8 @@ var (
 					"credentials",
 					"objectstore",
 					"schedulepolicy",
-					"volume",
 					"mountattach",
+					"migrate",
```
jrivera-px: Was this removed on purpose?

ggriffiths: Yeah, github hides it, but that slice of strings actually has `"volume"` twice in there

ggriffiths: we noticed it when looking through this code in the meeting. @lpabon said to remove it.

jrivera-px: Yeah ok cool,  yeah I remember now.   Was just checking.

PR #924, opened 2019-03-13
Add ownership to storage policy

ram-infrac:
This PR adds ownership to a storage policy.

```SdkStoragePolicy``` object is enhanced with the following -
- Force: bool -- If force enabled, then UC4.1 will be done, if false (default), then UC4.2 will be done
- AllowUpdate: bool -- If enabled, allows the volume to be updated even if it violates the policy. If false (default), user cannot update volume if the update violates the storage policy.
- SdkStoragePolicy have ownership

Closes #

### CODE REVIEW

File: api/api.proto [hunk 1]
```diff
@@ -327,7 +327,7 @@ message VolumeSpec {
   IoStrategy io_strategy = 30;
```
lpabon: sdk version needs to be updated

api/api.proto
```diff
@@ -591,14 +591,23 @@ message CapacityUsageInfo {
...
+// If storage policy is set default in OpenStoragePolicy service, VolumeSpecPolicy will be 
+// used before creating volume to validate volume specs or ensure minimum volume creation 
+// rules followed
 message SdkStoragePolicy {
   // Name of storage policy.
   string name = 1;
   // VolumeSpecs to apply while creating volume.
   VolumeSpecPolicy policy = 2;
+  // Force if set to true volume specs will be overwritten, otherwise
+  // volume creation will be failed if volSpecs is not inline with storage policy
```
lpabon: "will be failed" should be "will fail if the volume specification..."

api/api.proto
```diff
@@ -591,14 +591,23 @@ message CapacityUsageInfo {
...
+  bool force = 3;
+  // AllowUpdate if set to true volume can be updated without storage Policy
```
lpabon: Can't use `AllowUpdate` because that is golang only. In other languages it can be `allow_update`. So in the comment, you do not need to refer to the name. You can just say: "Used to determine if a policy... and so on"

api/api.proto
```diff
@@ -1807,21 +1816,21 @@ service OpenStoragePolicy {
       };
     }
 
-  // Enforces specified storage policy
-  rpc Enforce(SdkOpenStoragePolicyEnforceRequest)
-    returns (SdkOpenStoragePolicyEnforceResponse) {
+  // Set specified storage policy as default policy
+  rpc SetDefault(SdkOpenStoragePolicySetDefaultRequest)
```
lpabon: Changing and deleting an API is quite disruptive in the SDK. We need to make sure this doesn't happen again because non-openstorage developers of 2.0.x will see missing APIs when moving to 2.1.

This is ok for now, since it has been here a short time.

api/api.proto
lpabon: SDK_CHANGELOG.md must also be updated

api/server/sdk/volume_ops.go
```diff
@@ -652,199 +659,223 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
 	return current
 }
 
-// GetEnforcedVolSpecs returns volume spec merged with enforced policy applied if any
-func GetEnforcedVolSpecs(locator *api.VolumeLocator, spec *api.VolumeSpec) (*api.VolumeSpec, error) {
-	if locator != nil {
-		// if volume label is provided, override default policy
-		if val, ok := locator.VolumeLabels["enforce_disable"]; ok {
```
ram-infrac: we don't need this anymore, will add separate PR for administrative bypass

api/api.proto
ram-infrac: just curious does api.proto support to declare some API's are deprecated or identityManager is only way

lpabon: Good question. In April, we will be coming up with a process of deprecating APIs, and also creating new Alpha-Beta-GA releases for APIs.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -39,9 +40,9 @@ type SdkPolicyManager struct {
 }
 
 const (
-	policyPrefix = "/storage/policy"
+	policyPrefix = "storage/policy"
 	policyPath   = "/policies"
-	enforcePath  = "/storage/policy/enforce"
+	defaultPath  = "storage/policy/enforce"
```
lpabon: Just checking. Looks like you remove the leading `/`

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -87,82 +97,78 @@ func (p *SdkPolicyManager) Create(
...
-	volSpecs := &api.VolumeSpecPolicy{}
 	m := jsonpb.Marshaler{OrigName: true}
-	// marshal old policy to volSpecs
-	oldSpec, err := m.MarshalToString(pol.StoragePolicy.GetPolicy())
+	updateStr, err := m.MarshalToString(req.GetStoragePolicy())
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "Json Marshal failed for policy %s: %v", req.StoragePolicy.GetName(), err)
+		return nil, status.Errorf(codes.Internal, "Json Marshal failed for policy %s: %v", req.GetStoragePolicy().GetName(), err)
 	}
-	jsonpb.UnmarshalString(oldSpec, volSpecs)
```
lpabon: There is a missing check here. This call takes whatever ownership information they have updated and saves it as the new value, but this should not be. Instead Ownership should be checked if it was passed in and if it was, it needs to go through ownership.Update() so that it checks for the correct update rules, like, the system admin is the only one that can change the owner. If we allowed this call to go through as it is, for example, a collaborator with write access could send a new ownership object and become the owner.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -238,106 +262,163 @@ func (p *SdkPolicyManager) Enumerate(
...
 		&api.SdkOpenStoragePolicyInspectRequest{
 			Name: req.GetName(),
 		},
 	)
 	if err != nil {
-		return nil, status.Errorf(codes.NotFound, "Policy with name %s not found", req.GetName())
+		return nil, err
+	}
+
+	// only administrator or owner can set policy as default
```
lpabon: Now that I am thinking about it, this affects the entire cluster, not just a resource. Affecting the cluster requires system admin ownership o.IsAdminByUser(). I think we meant that only the admin of the system (not a user with admin rights for a resource) can do this. If not, then anyone (any owner of a storage policy) with the correct RBAC can affect the cluster. What do you think?

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -238,106 +262,163 @@ func (p *SdkPolicyManager) Enumerate(
...
 ) (*api.SdkOpenStoragePolicyReleaseResponse, error) {
-	// empty represents no policy enforcement is enabled
+
+	policy, err := p.DefaultInspect(ctx, &api.SdkOpenStoragePolicyDefaultInspectRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	// only administrator or owner of storage policy can remove storage policy restriction
+	if !policy.GetStoragePolicy().IsPermitted(ctx, api.Ownership_Admin) {
```
lpabon: See comment above about setting SetDefault()

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -238,106 +262,163 @@ func (p *SdkPolicyManager) Enumerate(
...
+		return nil, err
 	}
 
-	return &api.SdkOpenStoragePolicyEnforceInspectResponse{
+	return &api.SdkOpenStoragePolicyDefaultInspectResponse{
 		StoragePolicy: inspResp.GetStoragePolicy(),
 	}, nil
 }
+
+func volSpecToSdkStoragePolicy(inst *SdkPolicyManager) error {
```
lpabon: Does this function have unit tests?

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: yes I observed in etcd2 and etcd3 `/ ` is treated differently 
for eg. 
in etcd3 this is allowed -
```
root_path/clusterid//storage/policy
root_path/clusterid/storage/policy
root_path/clusterid/cluster/node
root_path/clusterid//cluster/role/...
```
so keeping inline to avoid unexpected errors

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: good catch, thanks will update

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: I guess that's right, setting/releasing default policy should be done by Admin user only. Anyone with write/Admin access can pass their storage policy for volume create/update

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: yes

api/server/sdk/volume_ops.go
```diff
@@ -652,199 +659,223 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+
+		// should we only allow policy to be used by owner who has write/admin access
+		if !customPolicy.GetStoragePolicy().IsPermitted(ctx, api.Ownership_Admin) {
+			return nil, status.Errorf(codes.PermissionDenied, "Cannot use storage policy %v", customPolicy.GetStoragePolicy().GetName())
+		}
+		policy = customPolicy.GetStoragePolicy()
+	} else {
+		// check if default storage policy is set
+		defPolicy, err := storPolicy.DefaultInspect(context.Background(), &api.SdkOpenStoragePolicyDefaultInspectRequest{})
+		if err == kvdb.ErrNotFound || defPolicy.GetStoragePolicy() == nil {
```
lpabon: We should not be returning kvdb.ErrNotFound in SDK calls. This code should check for `status.code == codes.NotFound`

api/server/sdk/volume_ops.go
```diff
@@ -652,199 +659,223 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+			// no default storage policy,
+			// should we add NA as volume is not created by any storage policy here ?
+			return spec, nil
+		}
+		// err means there is policy stored, but we are not able to retrive it
+		// hence we are not allowing volume create operation
+		if err != nil {
```
lpabon: use `else if` to make it clear that this is part of the previous `if` statement.

api/server/sdk/volume_ops.go
```diff
@@ -652,199 +659,223 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
 
-	// check if enforcement is enabled
-	policy, err := storPolicy.EnforceInspect(context.Background(), &api.SdkOpenStoragePolicyEnforceInspectRequest{})
-	if err == kvdb.ErrNotFound || policy.GetStoragePolicy() == nil {
-		// enforce is disabled, return original specs
+	// track volume created using storage policy
+	spec.StoragePolicy = policy.GetName()
+	// check if volume update request, if allowupdate is set
+	// return spec received as it is
+	if isUpdate && policy.GetAllowUpdate() {
```
lpabon: in `isUpdate` we should check for `write` access

api/server/sdk/volume_ops.go
```diff
@@ -163,10 +165,10 @@ func (s *VolumeServer) Create(
...
 	source := &api.Source{}
 
-	// Update given spec according to enforcement policy set
+	// Validate/Update given spec according to default storage policy set
 	// In case policy is not set, should fall back to default way
 	// of creating volume
-	spec, err := GetEnforcedVolSpecs(locator, req.GetSpec())
+	spec, err := GetDefaultVolSpecs(ctx, req.GetSpec(), false)
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Unable to update vol specs according to policy %v", err)
```
lpabon: Make GetDefaultVolSpecs() return gRPC error style. No need to reformat here. We need GetDefaultVolSpecs() to return gRPC error style because it is the one checking for ownership and it calls stp.Inspect() which also returns in gRPC format.

api/server/sdk/volume_ops.go
```diff
@@ -652,199 +659,223 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
 		if customErr != nil {
 			return nil, customErr
 		}
-		// update spec with custom storage policy
-		updatedSpec := mergeVolumeSpecsPolicy(spec, customPolicy.GetStoragePolicy().GetPolicy())
-		return updatedSpec, nil
+
+		// should we only allow policy to be used by owner who has write/admin access
+		if !customPolicy.GetStoragePolicy().IsPermitted(ctx, api.Ownership_Admin) {
+			return nil, status.Errorf(codes.PermissionDenied, "Cannot use storage policy %v", customPolicy.GetStoragePolicy().GetName())
```
lpabon: Good, but caller re-pad the error and change the code. Make sure all caller do not re-error.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -87,82 +98,92 @@ func (p *SdkPolicyManager) Create(
...
 
-	if req.StoragePolicy.GetPolicy() == nil {
+	if req.GetStoragePolicy().GetPolicy() == nil {
 		return nil, status.Error(codes.InvalidArgument, "Must supply Volume Specs")
 	}
 
-	// Get Existing details to merge
-	pol, err := p.Inspect(ctx,
+	// Get Existing details to check access
+	oldPolicy, err := p.Inspect(ctx,
```
lpabon: This checks read access, but it should then check for WRITE access in Update.

lpabon: oh i see. instead of doing it at line 162, it should be done right after this inspect.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -175,23 +196,40 @@ func (p *SdkPolicyManager) Delete(
...
 	if policy.GetStoragePolicy() != nil && policy.GetStoragePolicy().GetName() == req.GetName() {
 		_, err := p.Release(ctx, &api.SdkOpenStoragePolicyReleaseRequest{})
 		if err != nil {
-			return nil, status.Errorf(codes.Internal, "Disable enforcement failed with: %v", err)
+			return nil, status.Errorf(codes.Internal, "Removal of default policy failed with: %v", err)
 		}
 	}
+
+	// Only the owner or the admin can delete
+	if !inspResp.GetStoragePolicy().IsPermitted(ctx, api.Ownership_Admin) {
```
lpabon: this needs to be at line 208.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -238,106 +277,166 @@ func (p *SdkPolicyManager) Enumerate(
...
 	if err != nil {
-		return nil, status.Errorf(codes.NotFound, "Policy with name %s not found", req.GetName())
+		return nil, err
+	}
+
+	// Only administrator can set policy as default storage
+	// policy restriction
+	user, _ := auth.NewUserInfoFromContext(ctx)
+	if !policy.GetStoragePolicy().GetOwnership().IsAdminByUser(user) {
+		return nil, status.Errorf(codes.PermissionDenied, "Only admin can set storage policy as default %v", req.GetName())
```
lpabon: Can you change this to `Only the storage system admin...`?

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -175,23 +196,40 @@ func (p *SdkPolicyManager) Delete(
...
+	if !inspResp.GetStoragePolicy().IsPermitted(ctx, api.Ownership_Admin) {
+		return nil, status.Errorf(codes.PermissionDenied, "Cannot delete storage policy %v", req.GetName())
+	}
+
 	_, err = p.kv.Delete(prefixWithName(req.GetName()))
 	if err != kvdb.ErrNotFound && err != nil {
 		return nil, status.Errorf(codes.Internal, "Failed to delete Storage Policy %s: %v", req.GetName(), err)
 	}
 
+	logrus.Infof("Storage Policy %v is deleted", req.GetName())
```
lpabon: i like these logs 👍

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -238,106 +277,166 @@ func (p *SdkPolicyManager) Enumerate(
...
+
+	policy, err := p.DefaultInspect(ctx, &api.SdkOpenStoragePolicyDefaultInspectRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	// only administrator can remove storage policy restriction
+	user, _ := auth.NewUserInfoFromContext(ctx)
+	if !policy.GetStoragePolicy().GetOwnership().IsAdminByUser(user) {
+		return nil, status.Errorf(codes.PermissionDenied, "Only admin can remove storage policy restriction")
```
lpabon: same here: `storage system admin`

api/server/sdk/volume_ops.go
```diff
@@ -652,199 +659,223 @@ func setSpecBool(current, req bool, reqSet interface{}) bool {
...
+		customPolicy, customErr := storPolicy.Inspect(ctx, inspReq)
 		if customErr != nil {
 			return nil, customErr
 		}
-		// update spec with custom storage policy
-		updatedSpec := mergeVolumeSpecsPolicy(spec, customPolicy.GetStoragePolicy().GetPolicy())
-		return updatedSpec, nil
+
+		// should we only allow policy to be used by owner who has write/admin access
+		if !customPolicy.GetStoragePolicy().IsPermitted(ctx, api.Ownership_Admin) {
```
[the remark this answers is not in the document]
ram-infrac: so someone who has read access to storage policy can apply custom storage policy while creating volume

PR #925, opened 2019-03-13
Ownership support in credentials

### LINKED ISSUE #907: Credentials need ownership
Feature v2

**What happened**:
The credentials need ownership which provide more control on who can access which credential and who can see them.

# Design

Ownership will be added to the credentials and checked by the cloud backups APIs in the following manner:

Credential ACLs | Allowed Credential SDK APIs
------------ | -------------
Read | Enumerate, Inspect, (use for backups)
Write | Update(new, updates only ownership)
Admin | Can delete credential

lpabon:
Provides ownership and ACLS per credentials. This allows credential access check for both credential management and usage during cloud backup.

Closes #907


api/server/sdk/credentials.go
```diff
@@ -229,7 +237,15 @@ func (s *CredentialServer) Delete(
 		return nil, status.Error(codes.InvalidArgument, "Must provide credentials uuid")
 	}
 
-	err := s.driver(ctx).CredsDelete(req.GetCredentialId())
+	// Check ownership
+	_, err := s.Inspect(ctx, &api.SdkCredentialInspectRequest{
```
talakad: Seems like Inspect only checks for Read access. However Delete must be an admin access right?

PR #929, opened 2019-03-14
Set node status after listener start

jrivera-px:
The Node status was being set to OK ("operational") before the REST services were started via the listeners.  This can cause an issue with automation because automation would immediately issue requests to the node as soon as it becomes OK ("operational") and would receive errors cause the REST services were not avail yet.    So don't set the node status to OK until the listeners have started.

Closes #
PWX-7978


jrivera-px:
Closing cause having issues with build..

PR #936, opened 2019-03-14
Add authentication to REST Calls for Migration.

### LINKED ISSUE #930: Re-route Migration calls to SDK Migration code.
Currently we only check ownership on the SDK side.
For this to work for legacy projects we need to re-direct the Rest Server calls to the SDK where we check for ownership.

pault84:
Closes #930

When I added ownership checks on the SDK for migration I forgot to add the passthrough on the REST side.

### DIFF
--- a/volume/drivers/fake/fake.go
+++ b/volume/drivers/fake/fake.go
@@ -275,6 +275,18 @@ func (d *driver) Detach(volumeID string, options map[string]string) error {
 	return nil
 }
 
+func (d *driver) CloudMigrateStart(request *api.CloudMigrateStartRequest) (*api.CloudMigrateStartResponse, error) {
+	return &api.CloudMigrateStartResponse{}, nil
+}
+
+func (d *driver) CloudMigrateCancel(request *api.CloudMigrateCancelRequest) error {
+	return nil
+}
+
+func (d *driver) CloudMigrateStatus(request *api.CloudMigrateStatusRequest) (*api.CloudMigrateStatusResponse, error) {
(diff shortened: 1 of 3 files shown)


lpabon:
Needs tests.

pault84:
Oh right the sdk tests don't use this code.

volume/drivers/fake/fake.go
```diff
@@ -275,6 +275,18 @@ func (d *driver) Detach(volumeID string, options map[string]string) error {
 	return nil
 }
 
+func (d *driver) CloudMigrateStart(request *api.CloudMigrateStartRequest) (*api.CloudMigrateStartResponse, error) {
+	return &api.CloudMigrateStartResponse{}, nil
```
lpabon: Resturn the task id that was provided in the request, that way the tests can check if it got to this point.

volume/drivers/fake/fake.go
```diff
@@ -275,6 +275,18 @@ func (d *driver) Detach(volumeID string, options map[string]string) error {
...
+}
+
+func (d *driver) CloudMigrateCancel(request *api.CloudMigrateCancelRequest) error {
+	return nil
+}
+
+func (d *driver) CloudMigrateStatus(request *api.CloudMigrateStatusRequest) (*api.CloudMigrateStatusResponse, error) {
+	return &api.CloudMigrateStatusResponse{}, nil
```
lpabon: Return something here in the response the tests can check for.

PR #939, opened 2019-03-15
 Openstorage Cluster Manager with Cluster Domains.

adityadani:

Added following new APIs to the Cluster interface and the REST server
- UpdateSelfMetroDomain - updates the metro domains of this node in the cluster database
- UpdateMetroDomainsActiveMap - updates the map of metro domains to their active status
- ActivateMetroDomain - activates the given metro domain and updates the cluster database
- DeactivateMetroDomain - deactivates the given metro domain and updates the cluster

- ClusterManager.Start() takes in metro domain as an argument.
- It passes the metro domain info to gossip
- On a cluster database watch update, ClusterManager passes the new domain information to gossip.


adityadani:
Will post a corresponding SDK PR for the new APIs

lpabon:
We had a discussion offline and there will be SDK API instead of REST ones

adityadani:
Here is the PR for SDK APIs - > #950

cluster/manager/manager.go
```diff
@@ -327,27 +329,78 @@ func (c *ClusterManager) UpdateSchedulerNodeName(schedulerNodeName string) error
...
-	defer kvdb.Unlock(kvlock)
 
-	db, _, err := readClusterInfo()
-	if err != nil {
-		return err
+	return updateLockedDB("update-scheduler-name", c.selfNode.Id, updateCallbackFn)
+}
+
+func (c *ClusterManager) DeactivateClusterDomain(deactivateRequest *api.DeactivateClusterDomainRequest) error {
+	c.selfNodeLock.Lock()
```
sangleganesh: nit: probably don't need these locks.

cluster/manager/manager.go
```diff
@@ -327,27 +329,78 @@ func (c *ClusterManager) UpdateSchedulerNodeName(schedulerNodeName string) error
...
 
-	nodeEntry, ok := db.NodeEntries[c.selfNode.Id]
-	if !ok {
-		return fmt.Errorf("Node not found in cluster database")
+func (c *ClusterManager) ActivateClusterDomain(activateRequest *api.ActivateClusterDomainRequest) error {
+	c.selfNodeLock.Lock()
+	defer c.selfNodeLock.Unlock()
+
+	updateCallbackFn := func(db *cluster.ClusterInfo) (bool, error) {
+		isActive, _ := db.ClusterDomains.ActiveMap[activateRequest.ClusterDomain]
```
sangleganesh: if !db.ClusterDomains.ActiveMap[activateRequest.ClusterDomain] {
  db.ClusterDomains.ActiveMap[activateRequest.ClusterDomain]  = true  
 return true, nil
}
return false, nil
may be shorter.

api/api.go
```diff
@@ -241,6 +242,12 @@ type Cluster struct {
 
 	// TunnelConfig for the cluster [key, cert, endpoint]
 	TunnelConfig TunnelConfig
+
+	// ClusterDomainsActiveMap is a map of cluster domains to a boolean value
+	// indicating which domains are active or inactive.
+	// All the nodes from domains which are inactive do not
+	// participate in quorum decisions
+	ClusterDomainsActiveMap types.ClusterDomainsActiveMap
```
sangleganesh: nit: suggest to use an enum for state, it is more explicit and allows expansion in future. (state = active, state = inactive).

cluster/manager/manager.go
```diff
@@ -327,27 +329,78 @@ func (c *ClusterManager) UpdateSchedulerNodeName(schedulerNodeName string) error
...
+			return false, nil
+		}
+		db.ClusterDomains.ActiveMap[activateRequest.ClusterDomain] = true
+		return true, nil
+	}
+	return updateLockedDB("deactivate-cluster-domain", c.selfNode.Id, updateCallbackFn)
+}
+
+func (c *ClusterManager) UpdateSelfClusterDomain(selfClusterDomain string) error {
+	if c.gossip != nil {
```
sangleganesh: this should be done in the watch - the update below may fail gossip would be updated, so the states would not be in sync.

cluster/manager/manager.go
```diff
@@ -327,27 +329,78 @@ func (c *ClusterManager) UpdateSchedulerNodeName(schedulerNodeName string) error
...
+		c.gossip.UpdateSelfClusterDomain(selfClusterDomain)
+	}
+
```
sangleganesh: does it have to be self ? can it be accepting node ids instead ?

cluster/manager/manager.go
```diff
@@ -492,6 +546,15 @@ func (c *ClusterManager) watchDB(key string, opaque interface{},
 
 	c.gossip.UpdateCluster(c.getNonDecommisionedPeers(db))
 
+	if len(db.ClusterDomains.ActiveMap) > 0 {
+		// Copy over the active map
+		c.activeMap = make(types.ClusterDomainsActiveMap)
```
sangleganesh: (below when starting gossip we don't make a copy, I would suggest gossip always making a copy of this instead of osd passing a copy).

cluster/manager/manager.go
```diff
@@ -1132,6 +1227,14 @@ func (c *ClusterManager) initListeners(
 			return ErrNewNodeOverMaxCapacity(clusterMaxSize)
 		}
 
+		if len(c.selfClusterDomain) > 0 {
+			// Add an entry of our cluster domain if it does not exists
```
sangleganesh: this can race with someone else setting the domain to inactive ? don't think we need it here ..

cluster/manager/manager.go
adityadani: gossip always makes a copy of the active map received through the API.
I wanted to make a copy here as well, because DB is being used elsewhere in this function as well.

cluster/manager/manager.go
adityadani: after our yesterday's conversation, do you think this is still true?

sangleganesh: 👍

cluster/manager/manager.go
adityadani: Added an enum ClusterDomainState in gossip.

api/api.go
adityadani: Added an enum ClusterDomainState in gossip.

cluster/manager/manager.go
adityadani: done

cluster/manager/manager.go
adityadani: In the current model, the cluster domain is passed to each node only at startup. I wanted to add a provision for future, to update the cluster domain on the fly if required for self node. We can change the API to take in nodes, if it is required.

PR #940, opened 2019-03-15
If the input driver name is not in the map use the default driver

jrivera-px:
…driver.

Clients just using the API request do not specify which driver is to be used.   Need a default.   So if the driver name is not found in the map just use the default.  

Closes #
PWX-8054

### DIFF
--- a/api/server/sdk/server.go
+++ b/api/server/sdk/server.go
@@ -23,13 +23,8 @@ import (
 	"os"
 	"sync"
 
-	"github.com/sirupsen/logrus"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/credentials"
-
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_auth "github.com/grpc-ecosystem/go-grpc-middleware/auth"
-
 	"github.com/libopenstorage/openstorage/alerts"
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/spec"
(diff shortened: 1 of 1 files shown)


api/server/sdk/server.go
```diff
@@ -481,7 +479,7 @@ func (s *sdkGrpcServer) useAlert(a alerts.FilterDeleter) {
 // Accessors
 func (s *sdkGrpcServer) driver(ctx context.Context) volume.VolumeDriver {
 	driverName := grpcserver.GetMetadataValueFromKey(ctx, ContextDriverKey)
-	if driverName == "" {
+	if s.driverHandlers[driverName] == nil {
```
lpabon: This works, but normally it is written as follows:

```go
if handler, ok := s.driverHandlers[driverName]; ok {
    return handler
} else {
    return s.driverHandlers[DefaultDriverName]
}
```

lpabon: Also just realized that the code as it is may not be what the caller wanted. Since they asked for a driver, and that driver, maybe, had been initialized to `nil` by the caller to the SDK, then instead the caller would be getting the Default driver. With the snippet in the comment above, the caller would be getting whatever value was placed in the driver handler for the driver requested. I'm assuming the latter is what we want, right?

adityadani: @lpabon You are right!
@jrivera-px Lets modify it to what Luis has suggested.

jrivera-px: Yeah definitely ... will do it  now..

PR #943, opened 2019-03-15
Add support to pass security token through token_secret in DC/OS

### LINKED ISSUE #928: DCOS secrets need to be supported in Docker Volume to get auth token
FEATURE

**What happened**:
Like Kubernetes, OpenStorage should have the support for pointing to a secret to fetch the auth token.

ggriffiths:

* Add supports to pass token through token_secret in dc/os

Closes #928

### DIFF
--- a/api/server/docker.go
+++ b/api/server/docker.go
@@ -13,6 +13,7 @@ import (
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/spec"
 	"github.com/libopenstorage/openstorage/config"
+	osecrets "github.com/libopenstorage/openstorage/pkg/auth/secrets"
 	"github.com/libopenstorage/openstorage/pkg/grpcserver"
 	"github.com/libopenstorage/openstorage/pkg/options"
 	"github.com/libopenstorage/openstorage/pkg/util"
@@ -35,6 +36,8 @@ type driver struct {
 	sdkUds string
 	conn   *grpc.ClientConn
 	mu     sync.Mutex
+
+	secretsStore osecrets.Auth
 }
 
 type handshakeResp struct {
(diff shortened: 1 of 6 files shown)


api/server/docker.go
```diff
@@ -218,30 +222,60 @@ func (d *driver) handshake(w http.ResponseWriter, r *http.Request) {
...
-		return ctx, ""
+
+	// get token from opts
+	tokenFromOpts := request.Opts[api.Token]
+	if tokenFromOpts != "" {
+		return addTokenMetadata(ctx, tokenFromOpts), tokenFromOpts
 	}
 
+	// get token secret
+	secret, context, ok := d.GetTokenSecretFromString(request.Name)
```
adityadani: Check if request.Opts has a api.TokenSecret ?

api/server/docker.go
```diff
@@ -218,30 +222,60 @@ func (d *driver) handshake(w http.ResponseWriter, r *http.Request) {
...
+	if ok && d.secretsStore != nil {
+		token, err := d.secretsStore.GetToken(secret, context)
```
adityadani: I think we should return an error when secret store is not initialized, but a token_secret was provided. In this way the user would get the correct error.

In the current format, user would get "not authorized" error instead of "secret store not initialized"

api/server/docker.go
```diff
@@ -218,30 +222,60 @@ func (d *driver) handshake(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			return ctx, ""
```
adityadani: same comment as above

api/server/docker.go
```diff
@@ -218,30 +222,60 @@ func (d *driver) handshake(w http.ResponseWriter, r *http.Request) {
...
+	tokenFromName, tokenInName := d.GetTokenFromString(request.Name)
+	if tokenInName {
+		return addTokenMetadata(ctx, tokenFromName), tokenFromName
 	}
 
-	md := metadata.New(map[string]string{
-		"authorization": "bearer " + token,
-	})
-	return metadata.NewOutgoingContext(ctx, md), token
+	// get token secret
```
adityadani: nit: moving the GetTokenFromString and GetTokenSecretFromString to addTokenMetadata function, will reduce code duplication

ggriffiths: Ended up adding a `parseTokenInput` that's shared by `attachTokenMount` and `attachToken` to reduce code duplication

ggriffiths:
Addressed PR comments @adityadani

PR #944, opened 2019-03-15
Implementation of storageops for Azure Disks

piyush-nimbalkar:
This PR adds some Azure managed disk operations to implement storageops interface.
Here we are using an older version of sdk due to some grpc conflicts.

Closes #


pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,808 @@
...
+	if !ok {
+		return nil, storageops.NewStorageError(
+			storageops.ErrVolInval,
+			"Invalid volume template given",
+			a.instance,
+		)
+	}
+
+	ctx := context.Background()
+	future, err := a.disksClient.CreateOrUpdate(
```
disrani-px: Is the behavior for Create() different from `aws` because here it will also update the volume if it already exists?. We should check if it already exists and return an error if that is the case.

pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,808 @@
...
+				},
+				DiskSizeGB: d.DiskProperties.DiskSizeGB,
+			},
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("cannot create disk: %v", err)
+	}
+
+	t := func() (interface{}, bool, error) {
```
disrani-px: This snippet to wait for completion duplicated in all the backgrounded calls. All the future structs implement `azure.Future`. So maybe you can pass in the future variables to a `waitForCompletion`  API and wait for future.Done() to return true.

pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,808 @@
...
+		a.resourceGroupName,
+		diskName,
+	)
+	if err != nil {
+		return "", fmt.Errorf("cannot get disk %v: %v", diskName, err)
+	}
+
+	nextLun := nextAvailableLun(*vm.StorageProfile.DataDisks)
+	if nextLun < 0 {
+		return "", fmt.Errorf("No LUN available to attach the disk")
```
disrani-px: Can we return the number of attached luns here to help debug in case of issues?

pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,808 @@
...
+		return "", err
+	}
+
+	if vm.StorageProfile == nil || vm.StorageProfile.DataDisks == nil {
+		return "", fmt.Errorf("instance does not have any disks attached")
+	}
+
+	for _, d := range *vm.StorageProfile.DataDisks {
+		if *d.Name == diskName {
+			devPath, err := lunToBlockDevPathWithRetry(*d.Lun)
```
disrani-px: Can you add a comment as to why a retry is required here and not in other cases?

pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,808 @@
...
+
+	return "", storageops.NewStorageError(
+		storageops.ErrVolAttachedOnRemoteNode,
+		fmt.Sprintf("disk %s is not attached on: %s (Attached on: %v)",
+			diskName, a.instance, disk.ManagedBy),
+		a.instance,
+	)
+}
+
+func (a *azureOps) Snapshot(diskName string, readonly bool) (interface{}, error) {
```
disrani-px: Looks like read-write snapshots aren't supported for Azure, so setting `readonly` to false should return an error.

pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,808 @@
...
+		clientSecret, env.ResourceManagerEndpoint)
+	if err != nil {
+		return nil, err
+	}
+
+	authorizer := autorest.NewBearerAuthorizer(token)
+
+	disksClient := compute.NewDisksClient(subscriptionID)
+	disksClient.Authorizer = authorizer
+	disksClient.PollingDelay = clientPollingDelay
```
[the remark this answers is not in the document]
piyush-nimbalkar: Some of the calls happen on a lock, sometimes multiple of those happen. We don't want to hold the locks for such a long time as the default timeout period. The given time of roughly 1 min seems good enough for the calls make.

pkg/storageops/azure/azure.go
piyush-nimbalkar: As discussed offline `azure.Future` is a struct and cannot be used to implement a generic `waitForCompletion()`
In the next version of sdk we will be using the `future.WaitForCompletion` which is built.

pkg/storageops/azure/azure.go
piyush-nimbalkar: Just printing the number of data disks, which indirectly gives the same info.

pkg/storageops/test/storageops.go
```diff
@@ -28,7 +28,6 @@ func RunTest(
 			disk := create(t, d, template)
```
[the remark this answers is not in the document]
piyush-nimbalkar: Will do it in some other PR. For not created a issue to track it - https://github.com/libopenstorage/openstorage/issues/958

pkg/storageops/azure/azure.go
```diff
@@ -0,0 +1,844 @@
...
+		storageops.ErrVolAttachedOnRemoteNode,
+		fmt.Sprintf("disk %s is not attached on: %s (Attached on: %v)",
+			diskName, a.instance, disk.ManagedBy),
+		a.instance,
+	)
+}
+
+func (a *azureOps) Snapshot(diskName string, readonly bool) (interface{}, error) {
+	if !readonly {
+		return nil, fmt.Errorf("read-write snapshots are not supported")
```
disrani-px: Mention they are not supported for this provider ie Azure

PR #954, opened 2019-03-18
in-tree: Add check for empty secret/context on create/delete 

### LINKED ISSUE #953: in-tree: Error message for no secret/context should be more descriptive
FEATURE

**What happened**:
Users were getting a confusing error message when they didn't pass the secret name/context name. 

The error should be clear that an secretName/contextName was not found.

ggriffiths:

* Check for empty secret/context on create/delete 
* Return respective error messages.

Closes #953

### DIFF
--- a/api/server/middleware_auth.go
+++ b/api/server/middleware_auth.go
@@ -73,6 +73,15 @@ func (a *authMiddleware) createWithAuth(w http.ResponseWriter, r *http.Request, 
 		json.NewEncoder(w).Encode(&dcRes)
 		return
 	}
+	if secretName == "" {
+		errorMessage := "Access denied, no secret found in the annotations of the persistent volume claim"
+		a.log(locator.Name, fn).Error(errorMessage)
+		dcRes.VolumeResponse = &api.VolumeResponse{Error: errorMessage}
+		json.NewEncoder(w).Encode(&dcRes)
+		w.WriteHeader(http.StatusUnauthorized)
+		return
+	}
+
 	token, err := a.provider.GetToken(secretName, secretContext)
 	if err != nil {
 		a.log(locator.Name, fn).WithError(err).Error("failed to get token")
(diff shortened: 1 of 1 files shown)


api/server/middleware_auth.go
```diff
@@ -73,6 +73,15 @@ func (a *authMiddleware) createWithAuth(w http.ResponseWriter, r *http.Request,
 		json.NewEncoder(w).Encode(&dcRes)
 		return
 	}
+	if secretName == "" || secretContext == "" {
```
adityadani: It is ok for secretContext to be empty. So you can remove it from the check

api/server/middleware_auth.go
```diff
@@ -206,6 +215,14 @@ func (a *authMiddleware) deleteWithAuth(w http.ResponseWriter, r *http.Request,
 		json.NewEncoder(w).Encode(volumeResponse)
 		return
 	}
+	if secretName == "" || secretContext == "" {
+		errorMessage := "Internal error, no secret found in the annotations of the persistent volume claim"
```
adityadani: @lpabon If a cluster goes from no auth -> auth, then all the Kubernetes requests are going to get this error. And we will not be able to delete a volume through Kubernetes. Is it ok to just doc this edge case?

api/server/middleware_auth.go
```diff
@@ -206,6 +215,14 @@ func (a *authMiddleware) deleteWithAuth(w http.ResponseWriter, r *http.Request,
 		json.NewEncoder(w).Encode(volumeResponse)
 		return
 	}
+	if secretName == "" || secretContext == "" {
```
adityadani: again no need to check for secretContext

api/server/middleware_auth.go
lpabon: offline discussion we decided that it is ok to fail the PVC delete request coming from Kubernetes, but provide the SDK as the method of cleaning up the volumes in this situation.

PR #957, opened 2019-03-19
CloudBackup enumerate status conversion.

talakad:
Also fix bug in the default creds path for enumerate.

Status conversion to string while enumerating cloud backups.
Closes #

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -857,6 +857,29 @@ func CloudBackupStatusTypeToSdkCloudBackupStatusType(
 	}
 }
 
+func SdkCloudBackupStatusTypeToCloudBackupStatusString(
+	t SdkCloudBackupStatusType,
+) string {
+	switch t {
+	case SdkCloudBackupStatusType_SdkCloudBackupStatusTypeNotStarted:
+		return string(CloudBackupStatusNotStarted)
+	case SdkCloudBackupStatusType_SdkCloudBackupStatusTypeDone:
+		return string(CloudBackupStatusDone)
+	case SdkCloudBackupStatusType_SdkCloudBackupStatusTypeAborted:
+		return string(CloudBackupStatusAborted)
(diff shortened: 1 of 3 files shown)


api/api.go
```diff
@@ -857,6 +857,29 @@ func CloudBackupStatusTypeToSdkCloudBackupStatusType(
 	}
 }
 
+func SdkCloudBackupStatusTypeToCloudBackupStatusString(
+	t SdkCloudBackupStatusType,
+) string {
+	switch t {
+	case SdkCloudBackupStatusType_SdkCloudBackupStatusTypeNotStarted:
```
disrani-px: I think this is a problem with the SDK types. This fixes the issue with the status for the REST API but the SDK status strings still are "SdkCloudBackupStatusTypeDone". Why do users care about what interface was being used in the status string? And the status is already stored in the Backup Status field, so it doesn't need to have redundant information. The status string should just be "Done"

lpabon: Stings cannot be stored or described in gRPC proto file, only enums that is why it is called SdkCloudBackupStatusTypeDone.

disrani-px: That just means the variable names need to be fixed. Please don't merge PRs when changes are requested on them

disrani-px: You can look at the CloudMigrate status and phase strings on how to do this correctly: https://github.com/libopenstorage/openstorage/blob/master/api/api.pb.go#L992

disrani-px: And this needs to be fixed for all the other SDK strings too

lpabon: @disrani-px the method you have above works well. Please write an issue so that we can fix all the values in the SDK to use that model. In the mean time, release-6.0 needs to have the current model.

lpabon: Actually, I'll create it.

lpabon: Filed #962

PR #964, opened 2019-03-21
Add TaskId back to MigrationStart

### LINKED ISSUE #930: Re-route Migration calls to SDK Migration code.

pault84:

I broke the tracking of migrations by removing it.
Also added back the osterrors check for idempotency.

Closes #930 

### DIFF
--- a/api/server/migrate.go
+++ b/api/server/migrate.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/libopenstorage/openstorage/api"
+	ost_errors "github.com/libopenstorage/openstorage/api/errors"
 )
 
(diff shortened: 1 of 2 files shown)


api/server/migrate.go
```diff
@@ -27,11 +28,13 @@ func (vd *volAPI) cloudMigrateStart(w http.ResponseWriter, r *http.Request) {
 	conn, err := vd.getConn()
 	if err != nil {
 		vd.sendError(vd.name, method, w, err.Error(), http.StatusInternalServerError)
+
```
disrani-px: Extra newline

PR #967, opened 2019-03-21
Fix Idempotency for migration starts

pault84:

Handle the error coming back from the driver and retain the old logic of idempotency.

### DIFF
--- a/api/server/migrate.go
+++ b/api/server/migrate.go
@@ -5,7 +5,8 @@ import (
 	"net/http"
 
 	"github.com/libopenstorage/openstorage/api"
-	ost_errors "github.com/libopenstorage/openstorage/api/errors"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 func (vd *volAPI) cloudMigrateStart(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


api/server/migrate.go
```diff
@@ -58,7 +59,8 @@ func (vd *volAPI) cloudMigrateStart(w http.ResponseWriter, r *http.Request) {
 
 	resp, err := migrations.Start(ctx, migrateRequest)
 	if err != nil {
-		if _, ok := err.(*ost_errors.ErrExists); ok {
+		serverError, _ := status.FromError(err)
```
lpabon: err should be checked so that serverError.Code() is not called

api/server/sdk/volume_migrate.go
```diff
@@ -86,6 +87,10 @@ func (s *VolumeServer) volumeGroupMigrate(
 	}
 	resp, err := s.driver(ctx).CloudMigrateStart(request)
 	if err != nil {
+		if _, ok := err.(*ost_errors.ErrExists); ok {
```
lpabon: 👍

api/server/migrate.go
```diff
@@ -58,8 +59,13 @@ func (vd *volAPI) cloudMigrateStart(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		if _, ok := err.(*ost_errors.ErrExists); ok {
-			w.WriteHeader(http.StatusConflict)
+		if serverError, ok := status.FromError(err); ok {
+			if serverError.Code() == codes.AlreadyExists {
+				w.WriteHeader(http.StatusConflict)
+				return
+			}
+
+			vd.sendError(method, startReq.TargetId, w, err.Error(), http.StatusInternalServerError)
```
lpabon: This is the same as line 71. You can probably combine them.

PR #986, opened 2019-03-23
Set SystemTokenManager have internal issuer - system_token_issuer

### LINKED ISSUE #985: Cloud migration failing due to systemTokenManager mismatched issuers

**What happened**:
* Cluster is getting permission denied when cloud migration start happens.
* Because the issuers do not match.

* Clusters should be able to migrate between each other

ggriffiths:

* Sets the system token issuer to be a set issuer, allowing clusters system managers to communicate with each other.

Closes #985


ggriffiths:
Closing this for now as it's not the right solution

PR #991, opened 2019-03-26
Anonymize token entry in the logrequest ID entry.

jrivera-px:
Token entries should not be written to logs.  When logging server requests replace ID 'token'
entry with '*******' to hide the token string.

Closes # PWX-8174


api/server/server.go
```diff
@@ -21,6 +21,20 @@ type Route struct {
...
+type AnonRegex struct {
+	anonymizeRegx *regexp.Regexp
+	replaceString string
+}
+
+// anonTokenRegxes[] used to hide strings when logging.
+var anonRegxes = []AnonRegex{
+	{
+		// Anonymize the token=.*, replace with token=***..*,
+		anonymizeRegx: regexp.MustCompile(`token=.*,`),
```
ggriffiths: should this be `token=.*,*` in case there's no comma after? i.e. a log statement `could not find token=abcd`. 

Maybe not, just unfamiliar with what all logs look like.

api/server/server.go
```diff
@@ -197,11 +211,23 @@ type restBase struct {
 	name    string
 }
 
+func regxAnonString(inputStr string) string {
```
ggriffiths: maybe add a quick unit test for this function. To make sure the regex is working as expected

api/server/server.go
```diff
@@ -21,6 +21,20 @@ type Route struct {
 	fn   func(http.ResponseWriter, *http.Request)
 }
 
+type AnonRegex struct {
```
ggriffiths: nit: Maybe add quick comment on what this struct is for

api/server/server.go
jrivera-px: This actually is specific to the token in the 'ID=' part of the output log string.    So the format I followed would be for that one.   Which is described in the bug.   However I can see that if you wanted to capture everything included after the token you could add that regex but then having both regex's one could overdue the other.  I guess this should not be this generic and should be relegated to the types of fields..   Like this one should be use for IDs and if we wanted we could create another []AnonRegex for Msgs.. I guess.   I will rename this for now to anonIDRegxes...

api/server/server.go
jrivera-px: Yeah sounds good..

PR #993, opened 2019-03-26
Handle spec update for nodiscard, io_strategy

prabirpaul:
nodiscard and io_strategy weren't being handled for spec update

Closes # PWX-8217


api/api.proto
```diff
@@ -371,6 +371,11 @@ message VolumeSpecUpdate {
   // `ownership` message is an empty string then the value of `owner` in
   // the `VolumeSpec.Ownership.owner` will not be updated.
   Ownership ownership = 26;
+  // Nodiscard specifies if the volume will be mounted with discard support disabled.
```
lpabon: These values look good. You will need to update the SdkVersion and the SDK_CHANGELOG.md

prabirpaul: Updated.

PR #995, opened 2019-03-26
Add view storage policy rule for system.user (master)

ram-infrac:
same as #994 , against master branch
Closes #

### DIFF
--- a/pkg/role/sdkserviceapi.go
+++ b/pkg/role/sdkserviceapi.go
@@ -83,6 +83,14 @@ var (
 				Services: []string{"identity"},
 				Apis:     []string{"*"},
 			},
+			&api.SdkRule{
+				Services: []string{"policy"},
+				Apis: []string{
+					"*enumerate*",
+					// This will allow system.user to view default policy also
+					"*inspect*",
+				},
+			},
 		},
 	}
 )


pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -439,3 +444,37 @@ func volSpecToSdkStoragePolicy(inst *SdkPolicyManager) error {
...
+	newPolStr, err := m.MarshalToString(newPolicy.GetPolicy())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "Json Marshal failed for volSpec %s: %v", newPolicy.GetName(), err)
+	}
+	err = jsonpb.UnmarshalString(newPolStr, updatedSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	updatedPolicy.Name = newPolicy.GetName()
```
lpabon: I don't think we should be changing the name. The name is the identifier. If the caller wants to change the name, they should create a new policy and delete the old one. We do not allow, for example, changing the id of a volume, since it is the identifier of the volume. Same with roles.

pkg/storagepolicy/sdkstoragepolicy.go
```diff
@@ -128,8 +128,6 @@ func (p *SdkPolicyManager) Create(
 }
 
 // Update Storage policy
-// TODO: Decide whether update storage policy request should update ownership at all
-// We can check if user is admin then only allow update ownership
 func (p *SdkPolicyManager) Update(
```
lpabon: I don't think this is necessary. Update should just take the object given as the new value just as it did before. The caller should do a ReadModifyWrite to update accordingly. This is how Volume ownership, and roles work.

lpabon:
i think only the roles change is necessary in this PR.

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: Does similar behaviour should be followed for `volumePolicySpec`

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: changed it

pkg/storagepolicy/sdkstoragepolicy.go
ram-infrac: after discussion with @lpabon in the offline call, SDK should not try to understand changes requested client should do ReadModifyWrite

PR #999, opened 2019-03-27
Fix Inspect and Cloud Backup Create

pault84:

Fix Cloud Backup Create and Inspect.


api/server/sdk/cloud_backup.go
```diff
@@ -74,6 +74,9 @@ func (s *CloudBackupServer) Create(
 		Labels:         req.GetLabels(),
 	})
 	if err != nil {
+		if err == volume.ErrInvalidName {
+			return nil, status.Errorf(codes.AlreadyExists, "Failed to create backup: %v", err)
```
lpabon: Instead of using the same message, it should be something like "Backup already exists or invalid name"...

api/server/volume.go
```diff
@@ -774,12 +774,20 @@ func (vd *volAPI) enumerate(w http.ResponseWriter, r *http.Request) {
 	for i, s := range ids {
 		vol, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{VolumeId: s})
 		if err != nil {
-			e := fmt.Errorf("Failed to inspect volumeID: %s", err.Error())
-			vd.sendError(vd.name, method, w, e.Error(), http.StatusBadRequest)
+			if s, ok := status.FromError(err); ok && s.Code() != codes.NotFound {
```
lpabon: i think the logic here is a little confusing because if it is ever not `ok` then it will fall through to line 783.

I think you want `ok && s.Code() == codes.NotFound` then move lines 784-785 to inside the if, the move 778-779 out of the if scope

PR #1007, opened 2019-03-28
Remove unused variables to fix go 1.11 vet errors in unit tests

rodrigc:

This removes unused variables in the unit tests.  In go 1.11 and higher, *go test*
runs *go vet*.  go 1.11's *go vet* fails because of unused variables in the unit tests.
This fix causes the unit tests to succeed on go 1.11.

### DIFF
--- a/pkg/sanity/volume.go
+++ b/pkg/sanity/volume.go
@@ -54,7 +54,6 @@ var _ = Describe("Volume [Volume Tests]", func() {
 		var (
 			volumeID         string
 			numVolumesBefore int
-			numVolumesAfter  int
 		)
 
 		BeforeEach(func() {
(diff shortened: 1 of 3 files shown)


pkg/sanity/volume.go
```diff
@@ -775,9 +758,8 @@ var _ = Describe("Volume [Volume Tests]", func() {
 			err = volumedriver.Delete(volumeID)
 			Expect(err).ToNot(HaveOccurred())
 
-			volumes, err := volumedriver.Enumerate(&api.VolumeLocator{}, nil)
+			_, err = volumedriver.Enumerate(&api.VolumeLocator{}, nil)
 			Expect(err).ToNot(HaveOccurred())
-			numVolumesAfter = len(volumes)
```
harsh-px: correct fix should be to use `numVolumesAfter` and testing for it instead of ignoring them.

The previous test was not doing so.

rodrigc: ok

PR #1014, opened 2019-03-29
Add DisasterRecoveryMode option to ClusterPair requests

disrani-px:

Closes #


api/api.proto
```diff
@@ -3414,6 +3414,8 @@ message ClusterPairCreateRequest {
   string remote_cluster_token = 3;
   // Set the new pair as the default
   bool set_default = 4;
+  // If the pair is being created for DR mode
+  bool disaster_recovery_mode = 5;
```
talakad: Suggestion:Should this be a enum rather than bool,  to allow any flavors of disaster recovery?

disrani-px: Yeah, that's a good point. What should the types be though? Should the name be Mode, and different modes be Default and DisasterRecovery for now?

talakad: Yeah that sounds great. default would be disaster_recovery_mode_off.

disrani-px: Updated. Can you take a look again?

talakad: gook great!.

PR #1017, opened 2019-03-30
Return proper error msg for pxctl cloudmigrate.

jrivera-px:
When running 'pxctl cloudmigrate' on a cluster not configured for cloud migration.
Should return proper messages to indicate its unconfigured.

Closes #  PWX-8204

### DIFF
--- a/cluster/manager/pair.go
+++ b/cluster/manager/pair.go
@@ -197,12 +197,20 @@ func (c *ClusterManager) GetPair(
 	if id == "" {
 		id, err = getDefaultPairId()
 		if err != nil {
-			return nil, err
+			if err == kvdb.ErrNotFound {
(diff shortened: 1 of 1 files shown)


cluster/manager/pair.go
```diff
@@ -197,12 +197,20 @@ func (c *ClusterManager) GetPair(
 	if id == "" {
 		id, err = getDefaultPairId()
 		if err != nil {
-			return nil, err
+			if err == kvdb.ErrNotFound {
+				return nil, fmt.Errorf("No default cluster pair id set.")
```
talakad: we could take out "id" ? i mean just say "No default cluster pair found"

jrivera-px: K will do.   Thx

PR #1031, opened 2019-04-02
Fix DCOS token secret regex to be more flexible for full dcos secret path

ggriffiths:

* Regex was not general enough - missed case of no context and multiple levels of slashes. 
* Changed to use only secretName for entire secretPath, instead of secretName + secretContext.

### DIFF
--- a/api/spec/spec_handler_test.go
+++ b/api/spec/spec_handler_test.go
@@ -242,25 +242,35 @@ func TestGetTokenFromString(t *testing.T) {
 func TestGetTokenSecretFromString(t *testing.T) {
 	s := NewSpecHandler()
 
-	secret := "token1"
-	context := "team1"
+	tt := []struct {
+		InputSecret    string
+		ExpectedSecret string
+		Successful     bool
+	}{
+		{
+			"/px/secrets/alpha/token1",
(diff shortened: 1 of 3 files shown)


api/spec/spec_handler_test.go
```diff
@@ -242,25 +242,24 @@ func TestGetTokenFromString(t *testing.T) {
...
 
-	sample := "name=abcd,token_secret=/px/secrets/alpha/" + context + "/" + secret + ",token="
-	sample2 := "name=abcd,token_secret=padsasdax/secr//e/sdasda//ds/asda///sts/al//pha/" + context + "/" + secret + ",token="
+	sample := "name=abcd,token_secret=" + secret + ",token="
+	sample2 := "name=abcd,token_secret=" + secret2 + ",token="
 
-	secretParsed, contextParsed, ok := s.GetTokenSecretFromString(sample)
-	require.Equal(t, secret, secretParsed)
-	require.Equal(t, context, contextParsed)
+	secretParsed, ok := s.GetTokenSecretFromString(sample)
```
piyush-nimbalkar: May be add a test case where the is no nested path - `token_secret=token1`

PR #1035, opened 2019-04-03
Add support for volume creation with secret from storage class

### LINKED ISSUE #1033: Upgrading from noauth to an auth cluster will cause issues for k8s in-tree

**What happened**:
* When upgrading from no-auth to auth, if no secret info is provided in the PVC, it will fail.

* We should be able to upgrade from noauth->auth

ggriffiths:

Closes #1033

### DIFF
--- a/api/server/middleware_auth.go
+++ b/api/server/middleware_auth.go
@@ -19,6 +19,7 @@ import (
 	osecrets "github.com/libopenstorage/secrets"
 	"github.com/portworx/sched-ops/k8s"
 	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


api/server/middleware_auth.go
```diff
@@ -313,24 +314,43 @@ func (a *authMiddleware) parseSecret(
...
+		scName := pvc.ObjectMeta.Annotations[v1.BetaStorageClassAnnotation]
+		sc, err := k8s.Instance().GetStorageClass(scName)
+		if err != nil {
+			return "", "", err
+		}
+
+		// Get secret name from PVC first. If does not exists, get from SC.
+		secretName, nameFound := pvc.ObjectMeta.Annotations[secrets.SecretNameKey]
+		if !nameFound {
+			secretName, nameFound = sc.ObjectMeta.Annotations[secrets.SecretNameKey]
```
adityadani: I thought the secret name will be provided as a storage class parameter and not as storage class annotations. @lpabon Do we want to read annotations from storage class like this?

ggriffiths: As discussed offline, this info will be translated from storage class params to spec.VolumeLabels, so this PR is not needed.

PR #1036, opened 2019-04-03
Fix for volume group migration

pault84:
The code was using labels and not the Group object


api/api.proto
```diff
@@ -2483,6 +2485,8 @@ message SdkVolumeEnumerateWithFiltersRequest {
   map<string, string> labels = 3;
   // (optional) Ownership to match
   Ownership ownership = 4;
+  // (optional) Group to match
+  Group group = 5;
 }
```
lpabon: Need new SdkVersion and update to the SDK_CHANGELOG.md

volume/drivers/common/default_store_enumerator.go
```diff
@@ -184,6 +184,11 @@ func match(
 	if locator == nil {
 		return hasSubset(v.Locator.VolumeLabels, volumeLabels)
 	}
+	if locator.GetGroup() != nil {
+		if v.GetSpec().GetGroup() == nil || v.GetSpec().GetGroup() != locator.GetGroup() {
```
lpabon: I'm not sure this is correct. It is comparing the addresses of the Group objects. Instead it should be comparing the Id of the group object.

I think you should create a new file called `api/group.go` and in it add a receiver to `(* Group)` called `IsMatch` like the  one done for Ownership. Then it can do the comparison according to the values in the object.

pault84: gotcha makes sense.

volume/drivers/common/default_store_enumerator.go
```diff
@@ -184,6 +184,18 @@ func match(
 	if locator == nil {
 		return hasSubset(v.Locator.VolumeLabels, volumeLabels)
 	}
+	if locator.GetGroup() != nil {
```
lpabon: I think this code needs to be removed, right?

PR #1045, opened 2019-04-08
Continue on an error during Snap Enumerate

pault84:

We shouldn't return an error if inspect fails during a snapEnumerate
We can delete volumes while an enumerate is happening. and that results into a not found error.

### DIFF
--- a/api/server/volume.go
+++ b/api/server/volume.go
@@ -1029,9 +1029,13 @@ func (vd *volAPI) snapEnumerate(w http.ResponseWriter, r *http.Request) {
 	for _, s := range resp.GetVolumeSnapshotIds() {
 		vol, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{VolumeId: s})
 		if err != nil {
-			continue
+			if s, ok := status.FromError(err); ok && s.Code() != codes.NotFound {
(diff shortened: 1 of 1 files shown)


api/server/volume.go
```diff
@@ -1025,15 +1025,13 @@ func (vd *volAPI) snapEnumerate(w http.ResponseWriter, r *http.Request) {
...
-	snaps := make([]*api.Volume, len(resp.GetVolumeSnapshotIds()))
-	for i, s := range resp.GetVolumeSnapshotIds() {
+	snaps := make([]*api.Volume, 0)
+	for _, s := range resp.GetVolumeSnapshotIds() {
 		vol, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{VolumeId: s})
 		if err != nil {
-			e := fmt.Errorf("Failed to inspect volumeID: %s", err.Error())
-			vd.sendError(vd.name, method, w, e.Error(), http.StatusBadRequest)
-			return
+			continue
```
harsh-px: Instead of blinding ignoring all errors, this should just ignore the volume not found error.

harsh-px: Need to maintain the previous behavior where enumerate doesn't report errors for not found, but returns for other errors.

PR #1050, opened 2019-04-08
Fixes for handling unmounts on deleted volumes

harsh-px:

**What this PR does / why we need it**: This fixes an issue where unmount was not handling volumes that are deleted.

### DIFF
--- a/api/server/sdk/errors.go
+++ b/api/server/sdk/errors.go
@@ -1,0 +1,16 @@
+package sdk
+
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// IsErrorNotFound returns if the given error if due to not found
+func IsErrorNotFound(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	s, ok := status.FromError(err)
+	return ok && s.Code() == codes.NotFound
+}
--- a/api/server/sdk/errors.go
+++ b/api/server/sdk/errors.go
@@ -5,7 +5,7 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// IsErrorNotFound returns if the given error if due to not found
+// IsErrorNotFound returns if the given error is due to not found
 func IsErrorNotFound(err error) bool {
 	if err == nil {
 		return false
--- a/api/server/sdk/volume_node_ops.go
+++ b/api/server/sdk/volume_node_ops.go
@@ -89,7 +89,8 @@ func (s *VolumeServer) Detach(
 	}
 
 	// Get access rights
-	if err := s.checkAccessForVolumeId(ctx, req.GetVolumeId(), api.Ownership_Write); err != nil {
+	err := s.checkAccessForVolumeId(ctx, req.GetVolumeId(), api.Ownership_Write)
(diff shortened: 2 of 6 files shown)


harsh-px:
UT fixed. @lpabon , @adityadani can you take a look?

api/server/sdk/errors.go
```diff
@@ -0,0 +1,16 @@
+package sdk
```
lpabon: I like this 👍

api/server/sdk/volume_node_ops.go
```diff
@@ -89,7 +89,8 @@ func (s *VolumeServer) Detach(
 	}
 
 	// Get access rights
-	if err := s.checkAccessForVolumeId(ctx, req.GetVolumeId(), api.Ownership_Write); err != nil {
+	err := s.checkAccessForVolumeId(ctx, req.GetVolumeId(), api.Ownership_Write)
```
lpabon: Suggestion: You may want to pull in the implementation of `checkAccessForVolumeId` in case the implementation changes.

If you do not want to do it now, just add an issue to openstorage and/or a comment on that function of the expected return values.

api/server/volume.go
```diff
@@ -362,62 +330,85 @@ func (vd *volAPI) volumeSet(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if req.Options["WAIT_BEFORE_DELETE"] == "true" {
+		unmountOptions.NoDelayBeforeDeletingMountPath = false
+	} else {
+		unmountOptions.NoDelayBeforeDeletingMountPath = true
+	}
+
+	if req.Locator != nil || req.Spec != nil {
+		// Only update spec if spec and locator are not nil.
+		vol, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{
```
lpabon: You have to be careful on the use of `err`. Here, this `err` is a new one, but on line 379 and on, it is the one on the previous scope.

Then `err` from the top scope it is evaluated on 423.

harsh-px: Yeah that's okay. This scope is just for updating the volumes.

If the Inspect fails and the err is other than not found, it returns with error anyways.

Line 423 should be using `err` from the top scope as that reflects the actual attach, mount, detach and unmount operations.

api/server/sdk/volume_node_ops.go
harsh-px: added function comment

api/server/volume.go
```diff
@@ -362,62 +330,85 @@ func (vd *volAPI) volumeSet(w http.ResponseWriter, r *http.Request) {
...
 				})
 			}
-			if err != nil {
-				break
-			}
 		}
-		break
 	}
 
 	resVol, err2 := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{
```
adityadani: Don't we need the ErrNotFound error check here?

harsh-px: not needed here since inspect failure is saved in `err2`. The actual response in the REST response will be `err` which is for the set calls.

harsh-px:
@lpabon I'll proceeding with merge. If you have any further comments post-merge, let me know and I'll send a follow up PR.

PR #1052, opened 2019-04-09
Do not hide errors in enumerate

lpabon:
Need to return errors found during inspect in the enumerate handler of the old REST server


api/server/volume.go
```diff
@@ -774,6 +774,11 @@ func (vd *volAPI) enumerate(w http.ResponseWriter, r *http.Request) {
 		vol, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{VolumeId: s})
 		if err == nil {
 			vols = append(vols, vol.GetVolume())
+		} else if sdk.IsErrorNotFound(err) {
```
pault84: should we use this in snapEnumerate too?

lpabon: Yeah, i'll update it.

PR #1054, opened 2019-04-09
Add option for token creation to subtact a duration from IAT

ggriffiths:

* Adds an option to provide a time duration offset for IAT

### DIFF
--- a/pkg/auth/token.go
+++ b/pkg/auth/token.go
@@ -28,6 +28,12 @@ import (
 type Options struct {
 	// Expiration time in Unix format as per JWT standard
 	Expiration int64
+
+	// IATSubtract is the time duration you would like to remove from
+	// the token IAT (Issue At Time). This is useful as a guard against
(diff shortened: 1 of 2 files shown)


pkg/auth/token.go
```diff
@@ -89,7 +95,7 @@ func Token(
 		"email": claims.Email,
 		"name":  claims.Name,
 		"roles": claims.Roles,
-		"iat":   time.Now().Unix(),
+		"iat":   time.Now().Add(-options.IATSubtract).Unix(),
```
jrivera-px: What is the default value of time.Duration?   Is this going to be ok: https://stackoverflow.com/questions/23051973/what-is-the-zero-value-for-time-time-in-go?

ggriffiths: Time.Duration is just an int64 alias, so it should be 0. So it should have no effect if it's not.

jrivera-px: Ah ok..  cool..

PR #1056, opened 2019-04-10
SDK InspectWithFilters and Deep.Inspect

lpabon:
Inspect may be a high latency call in drivers, therefore doing
an sdk Enumerate/Inspect combination may be problematic. Instead,
this patch introduces a "deep" option for SDK Inspect() which makes it
act like it used to, but now instead of calling driver.Inspect(),
it calls the driver's Enumerate() calls, which may be cache friendly.

**Notes to reviewers**
There are many changes to unit tests files, fyi. You may want to skip those.


api/server/volume.go
```diff
@@ -751,35 +755,41 @@ func (vd *volAPI) enumerate(w http.ResponseWriter, r *http.Request) {
...
+		for _, s := range v {
+			// They asked for inspect of specific volumes. We must honor with deep inspects
+			resp, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{
+				VolumeId: string(s),
+				Options: &api.VolumeInspectOptions{
+					Deep: true,
+				},
+			})
+			if err == nil {
+				vols = append(vols, resp.GetVolume())
```
sangleganesh: can Inspect() return no error and no volumes in response (for ex. if permission is denied?)  ?

lpabon: Afaik, yes. And I think there is a lot of old code to handle that situation when calling Inspect(). For example: https://github.com/libopenstorage/stork/blob/master/drivers/volume/portworx/portworx.go#L381

sangleganesh: oh I mean respond.GetVolumes() may return nil, do we need to check ?

lpabon: I'll add a check for that, good point.

PR #1057, opened 2019-04-12
Change mount permissions check to match attach

ggriffiths:
Unmount uses a different helper function `checkAccessForVolumeId` that handles inspect/NotFound better

### DIFF
--- a/api/server/sdk/volume_node_ops.go
+++ b/api/server/sdk/volume_node_ops.go
@@ -134,18 +134,10 @@ func (s *VolumeServer) Mount(
 		return nil, status.Error(codes.InvalidArgument, "Invalid Mount Path")
 	}
 
-	// Get volume information
-	resp, err := s.Inspect(ctx, &api.SdkVolumeInspectRequest{
-		VolumeId: req.GetVolumeId(),
-	})
+	// Get access rights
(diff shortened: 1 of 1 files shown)


api/server/sdk/volume_node_ops.go
```diff
@@ -134,18 +134,10 @@ func (s *VolumeServer) Mount(
...
-		return nil, status.Errorf(codes.NotFound, err.Error())
-	}
-	vol := resp.GetVolume()
-
-	// Checks for ownership
-	if !vol.IsPermitted(ctx, api.Ownership_Write) {
-		return nil, status.Errorf(codes.PermissionDenied, "Access denied to volume %s", vol.GetId())
+	// Get access rights
+	err := s.checkAccessForVolumeId(ctx, req.GetVolumeId(), api.Ownership_Write)
+	if err != nil && !IsErrorNotFound(err) {
```
harsh-px: Since attach and mount are setup calls, let's not ignore not found errors.

For unmounts, during teardown, the volumes could already be deleted. So we are allowing users to proceed.

For setup calls, we want to volumes to exist.

ggriffiths: Makes sense. Changed it back to check for not found errors.

PR #1062, opened 2019-04-15
Increase buffer size to allow for many volumes

lpabon:
Need to increase the maximum buffer size when requesting data from thousands of volumes

### DIFF
--- a/api/server/volume.go
+++ b/api/server/volume.go
@@ -26,7 +26,14 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-const schedDriverPostFix = "-sched"
+const (
+	schedDriverPostFix = "-sched"
+
+	// We set it to 128Mi to support large number of volumes. Before, the client
+	// was using 4Mi and it would not allow the support of over 5k volumes.
+	// We increased to a very large value to support over 100k volumes.
+	maxMsgSize = 128 * 1024 * 1024
+)
 
 type volAPI struct {
 	restBase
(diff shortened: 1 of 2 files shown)


api/server/volume_test.go
```diff
@@ -2618,3 +2619,79 @@ func TestStorkVolumeInspect(t *testing.T) {
...
+		},
+	}
+
+	labels := make(map[string]string)
+	for i := 0; i < 50; i++ {
+		labels[fmt.Sprintf("label %d", i)] = fmt.Sprintf("label value %d", i)
+	}
+	req.GetLocator().VolumeLabels = labels
+
+	numvols := 10 * 1000
```
pault84: this is 10k not 100k

lpabon: oh yeah, i should change the test name

api/server/volume.go
```diff
@@ -63,7 +66,10 @@ func (vd *volAPI) getConn() (*grpc.ClientConn, error) {
 		var err error
 		vd.conn, err = grpcserver.Connect(
 			vd.sdkUds,
-			[]grpc.DialOption{grpc.WithInsecure()})
+			[]grpc.DialOption{
+				grpc.WithInsecure(),
+				grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMsgSize)),
```
ggriffiths: Should we also set the `MaxCallSendMsgSize`?

K8's fix for this also includes Send:
https://github.com/kubernetes/kubernetes/pull/63894/files

api/server/volume_test.go
lpabon: done

api/server/volume.go
lpabon: I'm not sure if we ever send that much data into the server. Let me know if you know

lpabon: Also, that is the server, we are affecting the client. I'd like not to affect the server unless needed.

ggriffiths: I interpreted it as the grpc server sending the results of `pxctl v l` over the grpc steam back to the client. But I just looked at the `MaxCallRecvMsgSize` documentation. Looks good.

api/server/volume.go
```diff
@@ -26,7 +26,10 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-const schedDriverPostFix = "-sched"
+const (
+	schedDriverPostFix = "-sched"
+	maxMsgSize         = 128 * 1024 * 1024
```
adityadani: nit: add a comment, why this is required for future readers.

PR #1066, opened 2019-04-18
AWS: Add tags to the volume on creation

adityadani:
Updates the vendor only for openstorage so that we can use AWS SDK, that allows specifying tags on creation.

Closes #


pkg/storageops/aws/aws.go
```diff
@@ -611,6 +621,7 @@ func (s *ec2Ops) Create(
 		return nil, s.rollbackCreate(*resp.VolumeId, err)
 	}
 	if len(labels) > 0 {
+		// Apply the provided labels as tags
```
piyush-nimbalkar: Instead of making two different calls to apply tags (create and apply tags), why not combine the labels and vol.Tags and let it add those in create?
This will fix the problem where the ebs vol is created but failed to add tags. Ideally this whole operation should be atomic.

adityadani: done

adityadani: done

PR #1070, opened 2019-04-19
Add new field RetentionDays to cloudbackup schedule.

talakad:
This allows users to separately specify when to force a full (maxBackups) and
for how long to retain the scheduled backups.

Closes #


api/api.proto
```diff
@@ -3026,11 +3026,15 @@ message SdkCloudBackupScheduleInfo{
...
   repeated SdkSchedulePolicyInterval schedules = 3;
-  // MaxBackups are the maximum number of backups retained
-  // in cloud.Older backups are deleted
+  // MaxBackups indicates when to force full backup to cloud. If RetentionDays
+  // is not specified or is 0 (older scheme), this is also the maximum number 
+  // of scheduled backups retained in the cloud. Older backups are deleted
   uint64 max_backups = 4;
+  // Number of days that Scheduled CloudBackups will be kept after which they
+  // are deleted
+ uint32 retention_days = 5;
```
disrani-px: We shouldn't change the order of the fields, it'll cause backward compatibility issues

talakad: thanks!. Will add this to the end

PR #1072, opened 2019-04-19
Support disk provisioning on VM scale sets in Azure

piyush-nimbalkar:

In azure the VM apis in scale set are different from the ones using by normal VMs. Adding support to attach/detach disk to both kinds of VMs.
- If the scale set name is non empty we assume the VM is running inside a scale set. The instance id passed in such a case is the just the index (instance id) of the VM.
- For non-scale set VMs the scale set name will be empty and instance id will have the name of the VM.

Closes #

### DIFF
--- a/pkg/storageops/azure/vmsclient.go
+++ b/pkg/storageops/azure/vmsclient.go
@@ -1,0 +1,36 @@
+package azure
+
+import (
+	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-06-01/compute"
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// vmsClient is an interface for azure vm client operations
+type vmsClient interface {
+	// describe returns the VM instance object
+	describe(instanceID string) (interface{}, error)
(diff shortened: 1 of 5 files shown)


pkg/storageops/azure/vmsclient.go
```diff
@@ -0,0 +1,36 @@
...
+type vmsClient interface {
+	// describe returns the VM instance object
+	describe(instanceID string) (interface{}, error)
+	// getDataDisks returns a list of data disks attached to the given VM
+	getDataDisks(instanceID string) ([]compute.DataDisk, error)
+	// updateDataDisks update the data disks for the given VM
+	updateDataDisks(instanceID string, dataDisks []compute.DataDisk) error
+}
+
+func NewVMsClient(
```
disrani-px: This doesn't need to be exposed outside the package, right?

piyush-nimbalkar:
Merged in cloudops repo - https://github.com/libopenstorage/cloudops/pull/3

PR #1082, opened 2019-04-24
Add fields for filtering cloudbackup enumerate.

talakad:
Also add support for pagination of enumerate output.

Closes #


api/api.go
```diff
@@ -343,6 +343,10 @@ type CloudBackupGenericRequest struct {
 	CredentialUUID string
 	// All if set to true, backups for all clusters in the cloud are processed
 	All bool
+	// StatusFilter indicates backups based on status
+	StatusFilter CloudBackupStatusType
+	// TagFilter indicates backups based on tag
```
disrani-px: Not sure if we need tag here, it is an implementation detail in portworx that isn't exposed in openstorage

talakad: Perhaps i should rename this to metadataType, to allow implementations to have their own filter based on metadata?

talakad: Changed this to metadata filters.

api/server/sdk/cloud_backup.go
```diff
@@ -220,14 +220,22 @@ func (s *CloudBackupServer) EnumerateWithFilters(
...
 			CredentialUUID: credId,
 			All:            req.GetAll(),
+			MetadataFilter: req.MetadataFilter,
 		},
-	})
+		ContinuationToken: req.ContinuationToken,
+		MaxBackups:        req.MaxBackups,
+	}
+	if req.StatusFilter != api.SdkCloudBackupStatusType_SdkCloudBackupStatusTypeUnknown {
+		enumerateReq.StatusFilter = api.CloudBackupStatusType(api.SdkCloudBackupStatusTypeToCloudBackupStatusString(req.StatusFilter))
```
disrani-px: Can we use a bigger API name for `SdkCloudBackupStatusTypeToCloudBackupStatusString`?

PR #1088, opened 2019-04-26
CSI 1.1 Volume Expand API implementation

### LINKED ISSUE #1086: CSI 1.1 Support
We need to support CSI 1.1 resize and other API changes.

ggriffiths:
* Adds CSI 1.1 support

Closes #1086


csi/controller.go
```diff
@@ -266,6 +276,17 @@ func (s *OsdCsiServer) CreateVolume(
 		return nil, status.Error(codes.InvalidArgument, e)
 	}
 
+	// Get parent ID from request: snapshot or volume
+	if req.GetVolumeContentSource() != nil {
```
ggriffiths: A test in csi-test was added to check that this is implemented. That was failing, so I added the implementation in this PR.

csi/controller.go
```diff
@@ -371,6 +392,63 @@ func (s *OsdCsiServer) DeleteVolume(
...
+
+	// Get secret if any was passed
+	ctx = s.setupContextWithToken(ctx, req.GetSecrets())
+
+	// If the new size is greater than the current size, a volume update
+	// should be issued. Otherwise, no operation should occur.
+	volumes := api.NewOpenStorageVolumeClient(conn)
+
+	// Update volume with new size
+	_, err = volumes.Update(ctx, &api.SdkVolumeUpdateRequest{
```
ggriffiths: Originally, I was checking if the current volume size is less than the new size (in order to verify an expand was needed), but `volumes.Update` checks for this and provides a good error message:

```
 pxctl v update 1050284155922952603 -s 100
Update Volume: Failed to update volume: Error updating size on remote node: Shrinking of volumes is not supported. Current size for volume 6df869038d7f39d899f998f6abb0a553b634018d58e7bec1996046d14284c32c is 101 GiB
```

csi/csisanity_test.go
```diff
@@ -127,8 +127,13 @@ func TestCSISanity(t *testing.T) {
 	}
 
 	// Start CSI Sanity test
+	targetPath := "/tmp/mnt/csi"
 	sanity.Test(t, &sanity.Config{
 		Address:    server.Address(),
-		TargetPath: "/mnt",
+		TargetPath: targetPath,
+		CreateTargetDir: func(p string) (string, error) {
+			os.MkdirAll(p+"/target", os.FileMode(0755))
```
ggriffiths: The CSI mock driver no longer creates this for us, so we need to do this manually with this flag. 

See release notes:
https://github.com/kubernetes-csi/csi-test/releases/tag/v2.0.0

csi/controller.go
```diff
@@ -371,6 +392,63 @@ func (s *OsdCsiServer) DeleteVolume(
...
+	} else if req.GetCapacityRange().GetRequiredBytes() < 0 || req.GetCapacityRange().GetLimitBytes() < 0 {
+		return nil, status.Error(codes.InvalidArgument, "Capacity ranges values cannot be negative")
+	}
+
+	// Get parameters
+	spec := &api.VolumeSpecUpdate{}
+
+	// Get Size
+	newSize := uint64(req.GetCapacityRange().GetRequiredBytes())
+	if newSize != 0 {
```
lpabon: What happens if it is `0`? The code continues, without anything to update. Should it instead check for zero in line 404?

ggriffiths: This was a typo I think. We don't need to check for 0.

Since RequiredBytes and LimitBytes are both optional per the spec, if you don't pass one, it'll use the zero value.

PR #1091, opened 2019-04-30
 Add GCP and AWS cloud ops for instance group support

harsh-px:

This adds support for InspectSelf and InspectSelfInstanceGroup.


pkg/cloudops/aws/aws.go
```diff
@@ -0,0 +1,257 @@
...
+		instInfo: &cloudops.InstanceInfo{},
+	}
+	err := ops.populateBasicInfo()
+	if err != nil {
+		return nil, err
+	}
+
+	ops.instInfo.Region = ops.instInfo.Zone[:len(ops.instInfo.Zone)-1]
+
+	ops.ec2 = ec2.New(
```
adityadani: We need to have a New(), that takes in a Credentials object, so that we support instance privileges as well and not just Env variables. We do something similar in storageops.

pkg/cloudops/aws/aws.go
```diff
@@ -0,0 +1,257 @@
...
+			id, len(out.Reservations[0].Instances))
+	}
+	return out.Reservations[0].Instances[0], nil
+}
+
+func labelsFromTags(input interface{}) map[string]string {
+	labels := make(map[string]string)
+	ec2Tags, ok := input.([]*ec2.Tag)
+	if ok {
+		for _, tag := range ec2Tags {
```
adityadani: nit: you can have a function to loop over tags, to reduce code duplication

pkg/cloudops/aws/aws.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+
+			if len(result.AutoScalingGroups) != 1 {
+				return nil, fmt.Errorf("DescribeAutoScalingGroups (%v) returned %v groups, expect 1",
+					value, len(result.AutoScalingGroups))
+			}
+
+			group := result.AutoScalingGroups[0]
+			if group.MinSize == nil || group.MaxSize == nil {
+				return nil, fmt.Errorf("Autoscaling group: %s does not have min or max set", value)
```
adityadani: we should have MinSize and MaxSize as pointers to int and the callers can decide what to do if they are nil.
Lets not return an error here.

pkg/cloudops/gcp/gcp.go
```diff
@@ -0,0 +1,227 @@
...
+	if len(gkeClusterName) == 0 || len(instanceTemplate) == 0 || len(clusterLocation) == 0 {
+		return nil, fmt.Errorf("API is currently only supported on the GKE platform.")
+	}
+
+	for _, item := range meta.Items {
+		if item == nil {
+			continue
+		}
+
+		if item.Key == kubeLabelsKey {
```
adityadani: I am assuming the kubeLabelsKey will be set only in GKE, so can we not just check on this Key and avoid the loop above and do it in a single loop.

pkg/cloudops/gcp/gcp.go
```diff
@@ -0,0 +1,227 @@
...
+			}
+
+			kubeLabels, err := parser.LabelsFromString(*item.Value)
+			if err != nil {
+				return nil, err
+			}
+
+			for labelKey, labelValue := range kubeLabels {
+				if labelKey == nodePoolKey {
+					nodePoolPath := fmt.Sprintf("projects/%s/locations/%s/clusters/%s/nodePools/%s", "portworx-eng", clusterLocation, gkeClusterName, labelValue)
```
adityadani: nit: split the fmt.Sprintf into multiple lines. It is just easier to read

pkg/cloudops/gcp/gcp.go
```diff
@@ -0,0 +1,227 @@
...
+					nodePool, err := g.containerService.Projects.Locations.Clusters.NodePools.Get(nodePoolPath).Do()
+					//nodePool, err := g.containerService.Projects.Locations.Clusters.NodePools.Get(labelValue).Do()
```
adityadani: nit: probably remove this comment?

pkg/cloudops/aws/aws.go
harsh-px: the tag types are different (ec2.Tag vs autoscaling.TagDescriptio). Hence the duplicate code to process them.

pkg/cloudops/cloudops.go
```diff
@@ -0,0 +1,43 @@
...
+// CloudObjectMeta encapsulates generic metadata information about a cloud object
+type CloudObjectMeta struct {
+	// Name is the name of the object
+	Name string
+	// ID is the object ID
+	ID string
+	// Labels are labels attached with the object
+	Labels map[string]string
+	// Zone is the cloud zone the object resides in
+	Zone string
```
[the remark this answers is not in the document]
harsh-px: Instance groups also have zones and regions where they belong, just like instances. Hence using a common meta structure for both.

pkg/cloudops/gcp/gcp.go
harsh-px: done

pkg/cloudops/gcp/gcp.go
harsh-px: removed

pkg/cloudops/gcp/gcp.go
harsh-px: good suggestion. Cleaned up.

pkg/cloudops/aws/aws.go
harsh-px: in clouds, I'm updating NewAWSOps to be consistent with how GCE works. This will use the credentials package from libopenstorage/secrets.

pkg/cloudops/aws/aws.go
harsh-px: done

harsh-px:
https://github.com/libopenstorage/cloudops/pull/6 supercedes this

PR #1092, opened 2019-04-30
Remove the cluster max size check from openstorage

adityadani:
- Let the Storage Listeners decide the maximum no. of nodes that can be
  added to the cluster.
- Add a new StorageListener API CanNodeJoin that gets invoked before joining a node
  to a cluster.


zoxpx:
@adityadani -- the travis-build is failing, looks like `cluster/mock/cluster.mock.go` will need update

adityadani:
@zoxpx I have updated the PR

cluster/manager/manager.go
```diff
@@ -1135,25 +1112,21 @@ func (c *ClusterManager) initListeners(
...
-		if clusterMaxSize > 0 && numNodes > clusterMaxSize {
-			// throw a slightly different error depending if node was already
-			// in the cluster.
-			if exist {
-				return ErrOldNodeOverMaxCapacity(clusterMaxSize)
+	initFunc := func(clusterInfo *cluster.ClusterInfo) error {
+		// Irrespective of whether the node is doing an Init or is
+		// already in cluster, check with listeners if it is OK to join
+		// this cluster.
+		for e := c.listeners.Front(); e != nil; e = e.Next() {
```
sangleganesh: there is already a finalizeCB callback, so not sure why we want to callback listeners again ?

adityadani: We want the clusterMaxSize check to be called during Init as well as Join. This piece of code (initFunc) gets called during both. While in Init finalizeCbs is populated, but in Join it is an empty array.

I have added a comment below about finalizeCbs being nil

PR #1099, opened 2019-05-01
Check for ownership on migrate cancel and status

### LINKED ISSUE #1098: Add ownership to migration status and cancel
FEATURE REQUEST

**What happened**:
All users can see migration status and cancel migrations

Users should only see status for volumes that they have access to. Also, they should only be able to cancel migrations that have full access to.

ggriffiths:

* This enables ownership for cancel and status.

Closes #1098

* This PR introduces a lot of Volume Inspects on behalf of the user, to check for access. 
* Open to suggestions on optimizing this to reduce the stress on the system.

### DIFF
--- a/api/server/sdk/volume_migrate.go
+++ b/api/server/sdk/volume_migrate.go
@@ -162,6 +162,28 @@ func (s *VolumeServer) volumeMigrate(
 	}, nil
 }
 
+func (s *VolumeServer) checkMigrationPermissions(ctx context.Context, taskId string) error {
+	// Inspect migration to get VolumeIds
+	resp, err := s.driver(ctx).CloudMigrateStatus(&api.CloudMigrateStatusRequest{
+		TaskId: taskId,
+	})
+	if err != nil {
+		return status.Errorf(codes.Internal, "Failed to get migration information : %v", err)
+	}
+
+	// Check that a user has access to all volumes being migrated
+	for _, cluster := range resp.Info {
+		for _, migrateInfo := range cluster.List {
+			if err := checkAccessFromDriverForVolumeId(ctx, s.driver(ctx),
+				migrateInfo.GetLocalVolumeId(), api.Ownership_Read); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
 // Cancel or stop a ongoing migration
 func (s *VolumeServer) Cancel(
 	ctx context.Context,
@@ -170,11 +192,15 @@ func (s *VolumeServer) Cancel(
 	if s.cluster() == nil || s.driver(ctx) == nil {
 		return nil, status.Error(codes.Unavailable, "Resource has not been initialized")
 	}
-
 	if req.GetRequest() == nil {
 		return nil, status.Errorf(codes.InvalidArgument, "Must supply valid request")
 	} else if len(req.GetRequest().GetTaskId()) == 0 {
 		return nil, status.Errorf(codes.InvalidArgument, "Must supply valid Task ID")
+	}
+
+	// Check if the user has access to all volumes associated with the TaskID
+	if err := s.checkMigrationPermissions(ctx, req.GetRequest().GetTaskId()); err != nil {
+		return nil, err
 	}
 	err := s.driver(ctx).CloudMigrateCancel(req.GetRequest())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


api/server/sdk/volume_migrate.go
```diff
@@ -184,6 +210,23 @@ func (s *VolumeServer) Cancel(
 	return &api.SdkCloudMigrateCancelResponse{}, nil
 }
 
+func (s *VolumeServer) filterStatusResponseForPermissions(
+	ctx context.Context,
+	resp *api.CloudMigrateStatusResponse) *api.CloudMigrateStatusResponse {
+	// Filter out volumes the user doesn't have access to
+	for _, cluster := range resp.Info {
+		for j, migrateInfo := range cluster.List {
+			if err := checkAccessFromDriverForVolumeId(ctx, s.driver(ctx),
```
ggriffiths: Just thought - instead of doing an `Inspect` for every Volume in the MigrateInfo, is it instead acceptable to do an `AllVolumes := Enumerate` to get all volumes the user has read access to, and compare `MigrateInfo` against that? Less driver calls. Do you think this would be a better way to check for Access for many volumes? @lpabon

ggriffiths: @lpabon addressed this feedback.

api/server/sdk/volume_migrate.go
```diff
@@ -197,6 +269,13 @@ func (s *VolumeServer) Status(
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "Cannot get status of migration : %v", err)
 	}
+	fmt.Println("*", *resp)
```
lpabon: You probably want to remove this.

api/server/sdk/volume_migrate.go
```diff
@@ -184,6 +211,51 @@ func (s *VolumeServer) Cancel(
...
+// that we have access to. While it seems too complicated, it minimizes the number of driver calls.
+func (s *VolumeServer) filterStatusResponseForPermissions(
+	ctx context.Context,
+	resp *api.CloudMigrateStatusResponse) (*api.CloudMigrateStatusResponse, error) {
+	var allVolIds []string
+
+	// get all volume ids to inspect
+	for _, cluster := range resp.Info {
+		for _, migrateInfo := range cluster.List {
+			allVolIds = append(allVolIds, migrateInfo.GetLocalVolumeId())
```
lpabon: this may work, but most of the time you want to `make` a slice of `0`

api/server/sdk/volume_migrate.go
```diff
@@ -184,6 +211,51 @@ func (s *VolumeServer) Cancel(
...
+		if !vol.IsPermitted(ctx, api.Ownership_Read) {
+			volAccessPermitted[vol.Id] = true
+		}
+	}
+
+	// remove volumes we don't have access to
+	for _, cluster := range resp.Info {
+		for j, migrateInfo := range cluster.List {
+			if found := volAccessPermitted[migrateInfo.GetLocalVolumeId()]; !found {
+				cluster.List = append(cluster.List[:j], cluster.List[j+1:]...)
```
lpabon: You should not adjust the same variable you are using for the for loop. It's looping is undefined. Instead just create a new list:

```go
filtered := make([]string, 0, len(allVols))
...
// if permitted:
filtered = append(filtered, migrateInfo.GetLocalVolumeId())
```

Then return the `resp` as needed

api/server/sdk/volume_migrate.go
```diff
@@ -162,6 +162,28 @@ func (s *VolumeServer) volumeMigrate(
...
+	})
+	if err != nil {
+		return status.Errorf(codes.Internal, "Failed to get migration information : %v", err)
+	}
+
+	// Check that a user has access to all volumes being migrated
+	for _, cluster := range resp.Info {
+		for _, migrateInfo := range cluster.List {
+			if err := checkAccessFromDriverForVolumeId(ctx, s.driver(ctx),
+				migrateInfo.GetLocalVolumeId(), api.Ownership_Read); err != nil {
```
talakad: Shouldn't this be api.OwnershipWrite? since the check is being used for cancel?

ggriffiths: I can change it to that, but anyone with Read Access can start a migration for that volume, so I figured that delete/status should match that. 

Let me see what Bhavana and Luis think.

talakad: ok

ggriffiths:
Just re-tested this on two px clusters with the latest changes - works as expected.

PR #1101, opened 2019-05-02
Add support for REST server extensions

### LINKED ISSUE #1100: SDK REST Gateway server cannot be extended

**What happened**:
The SDK Server rest server cannot be extended

It should be able extendable, similar to the gRPC server itself.

ggriffiths:

* Makes the gRPC SDK Rest Gateway server extendable. 

Closes #1100

### DIFF
--- a/api/server/sdk/rest_gateway.go
+++ b/api/server/sdk/rest_gateway.go
@@ -147,6 +147,9 @@ func (s *sdkRestGateway) restServerSetupHandlers() (http.Handler, error) {
 		api.RegisterOpenStoragePolicyHandler,
 	}
 
+	// REST Gateway extensions
+	handlers = append(handlers, s.config.RestServerExtensions...)
+
 	// Register the REST Gateway handlers
 	for _, handler := range handlers {
 		err := handler(context.Background(), gmux, conn)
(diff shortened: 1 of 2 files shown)


api/server/sdk/rest_gateway.go
```diff
@@ -155,6 +168,11 @@ func (s *sdkRestGateway) restServerSetupHandlers() (http.Handler, error) {
 		}
 	}
 
+	// Register extra services provided
+	if err := s.registerServerExtensions(gmux, conn); err != nil {
```
lpabon: If we do it this way, the caller will have to setup a function with a for loop to register their functions. During registering of the gRPC function, the single function the caller registers does not need a loop.

I think it would be cleaner to accept an array of handlers that take `(ctx, gmux, conn)` that way they can be appended to the `handlers` on line 164 and automatically registered.

What do you think?

ggriffiths: The works too - that's one of the ways I considered setting it up for gRPC. I was thinking that the way we set them up should be similar for both gRPC and REST, as but as long as we document the behavior it should be fine.

I can change it to accept a slice of handlers

ggriffiths: @lpabon I've changed it to just be a slice of handlers, and they're started the same way as the core handlers are created.

PR #1117, opened 2019-05-14
Update volume placement spec

harsh-px:

**What this PR does / why we need it**: This updates the proto definition of placement strategy. Note that this is not compatible with the previous definition. Breaking that is fine since noone is using placement rules.

Closes #


api/api.proto
```diff
@@ -3687,52 +3687,62 @@ message LocateResponse {
...
 message VolumePlacementStrategy {
-  // Rules defines a list of rules as part of the placement spec. All the rules specified will
-  // be applied for volume placement.
-  // Rules that have enforcement as "required" are strictly enforced while "preferred" are best effort.
-  // In situations, where 2 or more rules conflict, the weight of the rules will dictate which wins.
-  repeated VolumePlacementRule rules = 1;
+  // ReplicaAffinity defines affinity rules between replicas within a volume
+  repeated ReplicaPlacementSpec replica_affinity = 1;
+  // ReplicaAntiAffinity defines anti-affinity rules between replicas within a volume
+  repeated ReplicaPlacementSpec replica_anti_affinity = 2;
```
sangleganesh: put the affinity type inside the ReplicaPlacementSpec and VolumePlacementSpec ?

harsh-px: I kept it like this so that it's consistent with the k8s CRD. 

https://github.com/portworx/talisman/pull/88/files#diff-1f7c7c432e4b1db77dc8fa77eedcf8c2R40

`ReplicaPlacementSpec` here is a common type that can be used for affinity or anti-affinity.

When parsing, we will already know what it is using the field names replica_affinity and replica_anti_affinity.

harsh-px:
https://github.com/libopenstorage/openstorage/pull/1122 was approved and merged. So I'm just going to cherry-pick that to master.

PR #1123, opened 2019-05-23
Add new io profile applicable to shared volumes.

talakad:

Add new io profile for shared volumes
Closes #

I have not bumped any versions as this only add new value for io profile and does not change any api.


adityadani:
@lpabon Do we need to bump up the versions even if there is no RPC/Service changes?

PR #1125, opened 2019-05-27
Added a new field in the Node struct to store the hardware type

sharma-tapas:
- Also changed the test cases to test for this new field
 - Ran the make docker-test


api/api.proto
```diff
@@ -175,6 +175,15 @@ enum OperationFlags {
   OP_FLAGS_DETACH_FORCE = 2;
 }
 
+enum HardwareType {
+  // When we do not know the machine type alsp the default
+  Machine_Type_Unknown = 0;
```
disrani-px: Remove underscores from the names

sharma-tapas: Done.

PR #1130, opened 2019-05-30
Add cloud group backup API to SDK

### LINKED ISSUE #1129: Add cloud group-backup in SDK
FEATURE REQUEST 

**What happened**:
* Cloud group backup is not supported in the SDK

* This should be implemented.

ggriffiths:
This PR adds support for Cloud group backup in the SDK Server

Closes #1129

I have made the following two assumptions regarding ownership:
* If a user passes in a list of volumes, they must have access to ALL volumes in that list. If the user does not have access to a single volumes, the call fails.
* If a user passes in a groupID, they must have access to ALL volumes in that group. My reason for this:
If userA has access to 25 volumes inside a `GroupA`, and asks userB to backup `GroupA`, but userB only has access to 5 out of the 25 volumes, it would be dangerous to only backup those 5 volumes. This is because UserA asked for a backup of all of GroupA, but userB only has partial access to the volumes in that group. Returning successfully in this case would give UserB false confidence that the entirety of GroupA has been backed up.


api/server/sdk/cloud_backup.go
```diff
@@ -86,6 +86,67 @@ func (s *CloudBackupServer) Create(
...
+	ctx context.Context,
+	req *api.SdkCloudBackupGroupCreateRequest,
+) (*api.SdkCloudBackupGroupCreateResponse, error) {
+
+	if s.driver(ctx) == nil {
+		return nil, status.Error(codes.Unavailable, "Resource has not been initialized")
+	}
+	credId := req.GetCredentialId()
+	var err error
+	if len(req.GetGroupId()) == 0 && len(req.GetVolumeIds()) == 0 {
```
talakad: users could request group cloudsnaps for volume's with matching labels.

api/server/sdk/cloud_backup.go
```diff
@@ -86,6 +86,67 @@ func (s *CloudBackupServer) Create(
...
+	if len(req.GetCredentialId()) == 0 {
+		credId, err = s.defaultCloudBackupCreds(ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// Check ownership for all volume IDs
+	if len(req.GetVolumeIds()) > 0 {
+		if err := checkAccessFromDriverForVolumeIds(ctx, s.driver(ctx), req.GetVolumeIds(), api.Ownership_Read); err != nil {
```
talakad: these checks need to consider labels and group id as well.

api/server/sdk/cloud_backup.go
```diff
@@ -86,6 +86,67 @@ func (s *CloudBackupServer) Create(
...
+			return nil, err
+		}
+	}
+
+	// Check ownership for group ID
+	if len(req.GetGroupId()) > 0 {
+		if err := checkAccessFromDriverForGroupId(ctx, s.driver(ctx), req.GetGroupId(), api.Ownership_Read); err != nil {
```
talakad: This should include labels and must be done only len(req.GetVolumeIds() ==0.

api/server/sdk/cloud_backup.go
```diff
@@ -86,6 +86,67 @@ func (s *CloudBackupServer) Create(
...
+			return nil, err
+		}
+	}
+
```
talakad: If both groupID and volume list is not specified and only labels are provided, permissions perhaps should be checked for those volumes with matching labels?

api/server/sdk/cloud_backup.go
```diff
@@ -86,6 +86,101 @@ func (s *CloudBackupServer) Create(
...
+	}
+	if len(req.GetCredentialId()) == 0 {
+		credId, err = s.defaultCloudBackupCreds(ctx)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	switch {
+	// VolumeIDs and at least a GroupID or Labels are provided. Get filtered volumes based on GroupID and/or Labels,
```
talakad: Nice!

PR #1139, opened 2019-06-05
Add ability to look for the presence of any label in a set

jvinod:
Add ability to look for the presence of any label in a set
Add ability to match only label key and skip value


pkg/parser/labels.go
```diff
@@ -64,11 +64,36 @@ func MergeLabels(old map[string]string, new map[string]string) map[string]string
...
+func hasAnyLabel(set, subset map[string]string, matchValue bool) bool {
+	for k, v1 := range subset {
+		if v2, ok := set[k]; ok && (!matchValue || v1 == v2) {
+			return true
+		}
+	}
+	return false
+}
+
+func HasAnyLabel(set, subset map[string]string) bool {
```
adityadani: nit: a doc for this function or may be rename it to HasAtleastOneMatchingLabel makes it more explicit ?

jvinod: OK

PR #1147, opened 2019-06-10
Provide pool and disk information in Node Inspect

lpabon:
Node inspect call should go to a listener to update the node
status object if needed. This will allow the cluster listener
to update the node status with pool and disk information.


pault84:
LGTM

PR #1148, opened 2019-06-12
Allow cloudbackup schedule update

talakad:

Allows user to update existing cloudsnap with credential or schedule without having to recreate it.
Closes #


api/api.go
```diff
@@ -572,6 +574,11 @@ type CloudBackupGroupSchedCreateRequest struct {
 	MaxBackups uint
 	// Full indicates if scheduled backups must be full always
 	Full bool
+	// RetentionDays is the number of days that the scheduled backups will be kept
+	// and after these number of days it will be deleted
+	RetentionDays uint32
+	// Update indicates if existing schedule is being updated
+	Update bool
```
disrani-px: We should create new APIs for update instead of overloading the Create APIs

talakad: OK. Added new API for update.

api/api.proto
```diff
@@ -1864,6 +1864,15 @@ service OpenStorageCloudBackup {
       };
     }
 
+  // Update existing cloud backup schedule
+  rpc SchedUpdate(SdkCloudBackupSchedUpdateRequest)
+    returns (SdkCloudBackupSchedUpdateResponse) {
+      option(google.api.http) = {
+        post: "/v1/cloudbackups/schedules"
```
disrani-px: `post` is for create operations. `put` should be used for updates.

talakad: I missed it. Thanks! Done

PR #1153, opened 2019-06-19
Remove depricated config fields 'LoggingURL' & 'TunnelConfig'

jrivera-px:

Removing these fields to reduce which are not being used.


api/api.go
```diff
@@ -233,17 +223,11 @@ type Cluster struct {
 	// array of all the nodes in the cluster.
 	Nodes []Node
 
-	// Logging url for the cluster.
-	LoggingURL string
-
 	// Management url for the cluster
 	ManagementURL string
```
pault84: I think these 2 can go aswell but let's verify.

PR #1154, opened 2019-06-19
Fix api client request body length 0

pault84:

Moving the creation of the request inside of the for-loop.
After doing some research on the web regarding
` http: ContentLength=xxx  with Body length 0`

It seems that all these issues happen when the request's body buffer has been read.
So we're now creating a new request for every retry and filling it with the body buffer from the original call. (this one has not been read)

### DIFF
--- a/api/client/client_test.go
+++ b/api/client/client_test.go
@@ -1,12 +1,20 @@
 package client
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 	"time"
 
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+)
+
+var (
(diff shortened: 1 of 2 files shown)


pault84:
> re-sending the request cleared the buffer to len of 0?

So it seems like a race condition. but in the old code, the buffer from the request body gets read and then when we retry it and it throws the body 0 error.

In the changed code we just shove the body back in a new request and shoot out the new request.
I did notice a significant speed increase too with the new code.

api/client/client_test.go
```diff
@@ -1,21 +1,30 @@
...
 	"testing"
 	"time"
 
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+var (
+	attemptLock sync.Mutex
```
lpabon: Move this to the function

PR #1162, opened 2019-06-20
Support CSI Cloning

lpabon:
The code to support volume cloning is in the CSI driver, but
it did not advertise that it supported it. This patch informs
the caller that the driver supports volume cloning.

Cherry-pick of #1159


lpabon:
Closing, no need to patch release-6.0 with this fix.

PR #1163, opened 2019-06-20
Get PVC name, namespace, and labels/annotations on Provision

ggriffiths:
* Allows PVC labels/annotations to be added to the locator.VolumeLabels

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -17,6 +17,7 @@ limitations under the License.
 package csi
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/portworx/kvdb"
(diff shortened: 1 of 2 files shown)


csi/controller_test.go
```diff
@@ -2115,3 +2116,22 @@ func TestControllerDeleteSnapshot(t *testing.T) {
 	})
 	assert.NoError(t, err)
 }
+
+func TestGetPVCMetadata(t *testing.T) {
```
lpabon: awesome

csi/controller.go
```diff
@@ -260,6 +272,37 @@ func osdVolumeContext(v *api.Volume) map[string]string {
...
 }
 
+func getPVCMetadata(params map[string]string) (map[string]string, error) {
+	metadata := make(map[string]string)
+
+	metadata[intreePvcNameKey] = params[osdPvcNameKey]
+	metadata[intreePvcNamespaceKey] = params[osdPvcNamespaceKey]
+	decodedLabels := make(map[string]string)
+	decodedAnnotations := make(map[string]string)
+	if len(params[osdPvcAnnotationsKey]) > 0 || len(params[osdPvcLabelsKey]) > 0 {
```
lpabon: Why have an OR in the if. If we are going to use an `if` then it should be for each one (a function maybe?)

ggriffiths: Broke this down into a function.

PR #1199, opened 2019-08-19
SDK will now return a volume only if it is ready

### LINKED ISSUE #1180: During volume create, SDK should not return existing volume not in up state
**BUG REPORT**:

SDK returned an existing volume which will still in down state.

**What happened**:

- 2 volume creates came in at the same time for the same volume name
- While the first create call will still creating the volume, the 2nd create returned the volume as it existed in the inspect response. The volume was still in down state. This caused the k8s PVC to bind to a down volume. The test then created a snapshot in down state since the parent volume was in down state.

https://github.com/libopenstorage/openstorage/blob/master/api/server/sdk/volume_ops.go#L45

Return volume only if it is in up state.

1. Trigger 1st create volume call and it should put volume in down state before the 2nd call starts. Don't return from here
2. Trigger 2nd create call for the same name. In current code, it will return a down state volume.

**Environment**:
- Container Orchestrator and version: k8s on openshift (can be reproduced outside of k8s using above)
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

lpabon:
With this fix, any call creating a volume and noticing it already
exists now checks to make sure that the volume is UP and ready
before returning.

Closes #1180

* Updated the test port randomizer 
* Added a new wait call to `pkg/util` called `WaitForWithContext` which will automatically use the passed default value or the Deadline (WithTimeout) value passed in the context by the caller.

### DIFF
--- a/api/server/sdk/volume_ops.go
+++ b/api/server/sdk/volume_ops.go
@@ -19,6 +19,7 @@ package sdk
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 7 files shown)


api/server/sdk/volume_ops.go
```diff
@@ -42,12 +101,26 @@ func (s *VolumeServer) create(
 	// Check if the volume has already been created or is in process of creation
 	volName := locator.GetName()
 	v, err := util.VolumeFromName(s.driver(ctx), volName)
-	if err == nil {
+
+	// If the volume is still there but it is being delete, then wait until it is removed
+	if err == nil && v.GetState() == api.VolumeState_VOLUME_STATE_DELETED {
+		if err = s.waitForVolumeRemoved(ctx, volName); err != nil {
+			return "", err
```
harsh-px: can we throw a more user friendly error message here? Looks like currently it will throw a generic timed out error message from the util package. Instead it should be something along the lines

```
volume with same name is in the process of being deleted. Timed out waiting for deletion to complete.
```

api/server/sdk/volume_ops.go
```diff
@@ -42,12 +101,26 @@ func (s *VolumeServer) create(
...
 		// Check ownership
 		if !v.IsPermitted(ctx, api.Ownership_Admin) {
 			return "", status.Errorf(codes.PermissionDenied, "Volume %s already exists and is owned by another user", volName)
 		}
 
+		// Wait until ready
+		v, err = s.waitForVolumeReady(ctx, volName)
+		if err != nil {
+			return "", err
+		}
```
harsh-px: Same as above comment.

The error message could be
```
Timed out waiting for volume <name> to be in ready state.
```

PR #1214, opened 2019-09-04
Update vendor and use secrets instance for all GetToken calls

ggriffiths:


pkg/auth/secrets/secrets.go
```diff
@@ -118,18 +73,14 @@ func (a *Auth) GetToken(tokenSecretContext *api.TokenSecretContext) (string, err
...
-func (a *Auth) requestToContext(request *api.TokenSecretContext) map[string]string {
+func requestToContext(request *api.TokenSecretContext) map[string]string {
 	context := make(map[string]string)
 
 	// Add namespace for providers that support it.
-	switch a.ProviderType {
-	case TypeK8s:
+	switch lsecrets.Instance().String() {
+	case lsecrets.TypeK8s:
 		context[k8s.SecretNamespace] = request.SecretNamespace
```
ggriffiths: @piyush-nimbalkar @adityadani unrelated to this PR, but should the key inside of the secret always be `auth-token`? 

Currently it's `auth-token` for k8s and whatever the secret name is for all other providers. I can change that here though if needed.

api/server/middleware_auth.go
```diff
@@ -31,22 +31,11 @@ const (
...
-) (*authMiddleware, error) {
-	provider, err := secrets.NewAuth(
-		authType,
-		s,
-	)
-	if err != nil {
-		return nil, err
-	}
-	return &authMiddleware{provider}, nil
+func NewAuthMiddleware() (*authMiddleware, error) {
```
piyush-nimbalkar: Does this have to return an error now?

pkg/auth/secrets/secrets.go
piyush-nimbalkar: You mean the key will be the same as the name of the k8s secret?
Doesn't really matter as long as we have documented it properly. Also won't this break for existing users if they are using a key called `auth-token`?

api/server/middleware_auth.go
ggriffiths: No I can remove the error

pkg/auth/secrets/secrets.go
ggriffiths: I'd prefer if they're all `auth-token`. That's how it's currently documented and behaves for k8s.

For DCOS, I need to check if there's a reason for it using the secret name. There doesn't appear to be a `auth-token=value` in our documented example:
`dcos security secrets create --value=<auth-token> pwx/secrets/user1-token`

ggriffiths:
> LGTM
> Build seems to be broken though

Yeah need to fix a bunch of mock secrets errors since we're checking `mockDriver.String()` twice per GetToken call now.

pkg/auth/secrets/secrets.go
piyush-nimbalkar: That's because in k8s we have two layers of information to get a secret value - the secret name and then the key as it is a map of key-value pairs.
In dc/os, vault, kvdb we just need the name of the secret and the value of it is what we need. So there is no key as such, just the secret name.

piyush-nimbalkar:
You can use `.AnyTimes()` in the mock for `String()` method to get around it.

pkg/auth/secrets/secrets.go
ggriffiths: perfect, that explains it

PR #1226, opened 2019-09-11
CSI: Add req.Secrets encryption info to locator.VolumeLabels

ggriffiths:
* For CSI secrets compatibility, this adds all req.Secrets to locator.VolumeLabels.

Closes #

### DIFF
--- a/csi/csi.go
+++ b/csi/csi.go
@@ -21,6 +21,7 @@ import (
 	"sync"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/libopenstorage/openstorage/pkg/options"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/metadata"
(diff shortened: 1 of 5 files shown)


ggriffiths:
@adityadani, Can you re-review? Changed the behavior of this PR. Instead of the entirety of req.Secrets(), I'm now adding just `SECRET_NAME`, `SECRET_CONTEXT`, and `SECRET_KEY`.

I think this will avoid accidentally adding a plaintext password or auth token into the volume labels, without having to dynamically strip labels based on the `SECRET_KEY` provided.

I realized this was a problem when I was using the secret for all of the following (which someone may do...)
1. auth-token=...
2. SECRET_NAME=...
3. SECRET_CONTEXT=...
4. SECRET_KEY=encryptionkey
5. encryptionkey=...

csi/csi.go
```diff
@@ -125,11 +126,35 @@ func (s *OsdCsiServer) setupContextWithToken(ctx context.Context, csiSecrets map
...
 		return metadata.NewOutgoingContext(ctx, md)
 	}
+
 	return ctx
 }
 
+// addEncryptionInfoToLabels adds the needed secret encryption
+// fields to locator.VolumeLabels.
+func (s *OsdCsiServer) addEncryptionInfoToLabels(labels, csiSecrets map[string]string) map[string]string {
+	if csiSecrets == nil || len(csiSecrets) == 0 {
```
adityadani: nit: len(map) == 0 checks both

PR #1227, opened 2019-09-12
CSI: Default to Sharedv4

ggriffiths:
* Default to sharedv4 for shared volume creation

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -197,7 +197,7 @@ func (s *OsdCsiServer) ValidateVolumeCapabilities(
 		mode := capability.GetAccessMode()
 		switch {
 		case mode.Mode == csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER:
-			if v.Spec.Shared {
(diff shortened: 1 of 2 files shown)


csi/controller.go
```diff
@@ -378,7 +379,11 @@ func (s *OsdCsiServer) CreateVolume(
 	var newVolumeId string
 	if source.Parent == "" {
 		// Get Capabilities and Size
-		spec.Shared = csiRequestsSharedVolume(req)
+		if spec.Shared && !spec.Sharedv4 {
+			spec.Shared = csiRequestsSharedVolume(req)
+		} else {
+			spec.Sharedv4 = csiRequestsSharedVolume(req)
+		}
```
ggriffiths: Added legacy support for `shared` volumes if `shared=true` is set.  @adityadani 

`shared` volume are only used if 
`shared=true` and `sharedv4=false (or sharedv4 is empty)`  

All other cases result in sharedv4:
* default: no SC params but ReadWriteMany mode set
* or, when sharedv4=true

adityadani: perfect!

PR #1240, opened 2019-09-17
Don't redirect detach requests by default

harsh-px:
https://portworx.atlassian.net/browse/PWX-9974 has more details

By default, we don't need to redirect detach requests. They should be targeted to specific nodes. Redirection has an issue where detach requests could criss cross and lead to cyclic dependencies. So node1 sends request to node2 and node2 sends it back to node1. The request will get stuck on node1 since Portworx has a per volume lock that will first wait for the first request to complete.

Closes #


harsh-px:
Updated the PR to handle redirect detach option, if provided by user since we have use cases where users like pxctl or delete would want to redirect detach.

PR #1245, opened 2019-09-23
Change volume migration to use Enumerate instead of Inspect

ggriffiths:

Volume migration is taking a while, stuck `filterStatusResponseForPermissions`.

This change makes volume migration use Enumerate over inspect, which should be more efficient.

Closes #

### DIFF
--- a/api/server/sdk/volume_migrate.go
+++ b/api/server/sdk/volume_migrate.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/libopenstorage/openstorage/api"
 	ost_errors "github.com/libopenstorage/openstorage/api/errors"
+	"github.com/libopenstorage/openstorage/pkg/auth"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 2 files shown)


api/server/sdk/volume_migrate.go
```diff
@@ -229,18 +230,37 @@ func (s *VolumeServer) filterStatusResponseForPermissions(
...
+			Acls: &api.Ownership_AccessControl{
+				Groups: groupAcls,
+				Collaborators: map[string]api.Ownership_AccessType{
+					userinfo.Username: api.Ownership_Read,
+				},
+			},
+		},
+	}
+
+	allVols, err := s.driver(ctx).Enumerate(&locator, nil)
```
lpabon: The code should use the sdk Inspect call and let it decide what to do

api/server/sdk/volume_migrate.go
```diff
@@ -281,9 +284,11 @@ func (s *VolumeServer) Status(
 	}
 
 	// Filter out volumes we don't have access to
-	resp, err = s.filterStatusResponseForPermissions(ctx, resp)
-	if err != nil {
-		return nil, err
+	if auth.Enabled() {
```
ggriffiths: Since this call is expensive, we don't need to Enumerate all volumes only to permit all of them when auth is disabled.

api/server/sdk/volume_migrate.go
ggriffiths: Spoke with Luis and we agreed switching to driver.Enumerate() will have better performance as it will not perform a deep inspect.

PR #1251, opened 2019-09-26
New parameter to cloudbackup enumerate API

talakad:
Allows users to specify backup ID to enumerate.

Allows users to specify the backup ID that they want enumerate/check.
Closes #


api/server/sdk/cloud_backup_test.go
```diff
@@ -493,6 +493,69 @@ func TestSdkCloudBackupEnumerateWithFiltersBadArguments(t *testing.T) {
...
+	s := newTestServer(t)
+	defer s.Stop()
+
+	uuid := "uuid"
+	backupId := "one"
+	req := &api.SdkCloudBackupEnumerateWithFiltersRequest{
+		CloudBackupId: backupId,
+		CredentialId:  uuid,
+	}
+	list := &api.CloudBackupEnumerateResponse{
```
disrani-px: You should test with multiple backups with different IDs. Without a backupID in the request it should return both, with a backupID it should return only one. Even then, the test really depends on the implementation

talakad: There is already a test for testing with out backupID( TestSdkCloudBackupEnumerateWithFilters). And you are right, the test really depends on the backend implementation. Intention here was to show, only one backup being returned.

PR #1255, opened 2019-10-01
Handle nil migratestatus request

ram-infrac:
Make sure `CloudMigrateStatus` does not get nil `CloudMigrateStatusRequest`
Closes #

### DIFF
--- a/api/api.proto
+++ b/api/api.proto
@@ -1606,7 +1606,8 @@ service OpenStorageMigrate {
       };
     }
 
-  // Inspect the status of migration operation
+// Status for migration operation.
+// If status request is empty, status for all migration operation will be returned.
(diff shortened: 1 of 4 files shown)


api/api.proto
```diff
@@ -1606,7 +1606,9 @@ service OpenStorageMigrate {
       };
     }
 
-  // Inspect the status of migration operation
+  // Status for migration operation
+  // if status request is empty, status for all migration
```
lpabon: This is markdown, and this would look weird in the output to https://libopenstorage.github.io. It should be:

```
// Status for migration operation.
// If status request is empty, status for all migration operation will be returned.
```

I added a `.` and capitalized the second sentence.

PR #1258, opened 2019-10-01
Update cloud migrate group backup access denied message

### LINKED ISSUE #1257: Cloud backup group create access denied error
BUG REPORT

**What happened**:
Cloud backup group create should not mention volume ID which the user does not have access to. 

Instead, it should just return "Access denied to at least one volume in group id %s"

ggriffiths:

Updates the error message for cloud backup group create when a user does not have access to a volume in a group.

Closes #1257

### DIFF
--- a/api/server/sdk/cloud_backup.go
+++ b/api/server/sdk/cloud_backup.go
@@ -18,6 +18,7 @@ package sdk
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/volume"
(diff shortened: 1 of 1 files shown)


api/server/sdk/cloud_backup.go
```diff
@@ -133,7 +134,11 @@ func (s *CloudBackupServer) GroupCreate(
 
 		// Check ownership for this intersection
 		if err := checkAccessFromDriverForVolumeIds(ctx, s.driver(ctx), volumesToCheck, api.Ownership_Read); err != nil {
-			return nil, err
+			if err == kvdb.ErrNotFound {
```
talakad: checkAccessFromDriverForVolume returns "codes.PermissionDenied" for access errors.  Perhaps we should check for that and return other errors as is?

ggriffiths: checkAccessFromDriverForVolumeIds is used by other functions, so I don't want to return the error about GroupIDs there.

But yeah I'll change the caller to check for codes.NotFound instead.

PR #1259, opened 2019-10-02
Fix Storage Pools and deprecate id

lpabon:
SDK developers where using the ID instead of the uuid of the storage pool. Also fixed the copying of the pools into the SDK object.

### DIFF
--- a/SDK_CHANGELOG.md
+++ b/SDK_CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Releases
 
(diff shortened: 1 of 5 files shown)


SDK_CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Releases
 
+### v0.64.0 - Tech Preview (9/26/2019)
+
+* Deprecated StoragePool.ID. Please StoragePool.Uuid
```
adityadani: nit: Please use

PR #1273, opened 2019-10-09
Fix cloud migration status auth filtering

ggriffiths:

* driver.Enumerate, when given a list of volume Ids, will return
NotFound and an empty list if any volumes does exist. This was problematic, 
because we were building a list of which volumes the user has access to based on this response.
* If a volume ID did not exist during a cloud migration status call, this would cause all volume IDs to be perceived as access denied, resulting in an empty cloud migration status response.

Now, we just enumerate each volume and only include volumes which the user has access to.

### DIFF
--- a/api/server/sdk/volume_migrate.go
+++ b/api/server/sdk/volume_migrate.go
@@ -216,35 +216,6 @@ func (s *VolumeServer) Cancel(
 func (s *VolumeServer) filterStatusResponseForPermissions(
 	ctx context.Context,
 	resp *api.CloudMigrateStatusResponse) (*api.CloudMigrateStatusResponse, error) {
-	allVolIds := make([]string, 0)
-
-	// get all volume ids to inspect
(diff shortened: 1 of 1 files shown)


api/server/sdk/volume_migrate.go
```diff
@@ -255,12 +226,15 @@ func (s *VolumeServer) filterStatusResponseForPermissions(
 		filteredCluster.List = make([]*api.CloudMigrateInfo, 0)
 
 		for _, migrateInfo := range cluster.List {
-			if found := volAccessPermitted[migrateInfo.GetLocalVolumeId()]; found {
+			err := checkAccessFromDriverForLocator(ctx, s.driver(ctx), &api.VolumeLocator{
```
ggriffiths: checkAccessFromDriverForLocator performs a driver enumerate.

ggriffiths:
Tested this fix on my local cluster and it works as expected.

PR #1279, opened 2019-10-16
Use compare and set and resubscribe watch

sangleganesh:
Two changes:
1. Remove Put kvdb operation and use CompareAndSet instead
CompareAndSet is more strict and works even if the lock is released. Added a method to retry the update operation using CAS if the version got changed (due to lock release).

2. Watch changes
- Instead of querying db when there is a watch update, unmarshal data in watch update to get the db state.
- Instead of exiting on watch errors, try to resubscribe.

There needs to be one more change to exit watch if required version is compacted, will do that separately since it requires kvdb package update.

Closes #


sangleganesh:
PX automation BVT/FVT passed.

PR #1290, opened 2019-10-29
Permissive mode support

ggriffiths:

* Adds permissive mode core features

**In this PR is the following:**
* Permissive mode core
    * Create public user role
    * Replace empty token with public user info
    * Create public resources when public user
* Support for different modes
    * Disabled (no auth configured)
    * Permissive (auth configured)
    * Strict (via CLI/API or configuration)

**The following features will be implemented in separate PRs:**
* tooling for admin to see which vols are public
* be able to set volumes as public


api/server/volume_test.go
```diff
@@ -2415,37 +2415,6 @@ func TestMiddlewareVolumeDeleteSuccess(t *testing.T) {
 
 }
 
-func TestMiddlewareVolumeDeleteFailure(t *testing.T) {
```
ggriffiths: This test is no longer applicable. It was failing because a token was not provided.

api/api.go
```diff
@@ -1098,6 +1098,11 @@ func (v *VolumeSpec) GetCloneCreatorOwnership(ctx context.Context) (*Ownership,
 
 	// If there is user information, then auth is enabled
 	if userinfo, ok := auth.NewUserInfoFromContext(ctx); ok {
+		// if a public user clones a volume, we should not update
```
lpabon: Remove this and let `OwnershipSetUsernameFromContext(ctx, nil)` handle returning the appropriate ownership object, which could be a public one or a `nil`

api/api.proto
```diff
@@ -597,6 +597,10 @@ message Ownership {
     Admin = 2;
   }
 
+  message PublicAccessControl {
```
lpabon: This is comments

api/ownership.go
```diff
@@ -34,9 +34,14 @@ const (
 // a volume. It takes an ownership value if passed in by the user, then
 // sets the `owner` value to the user name referred to in the user context
 func OwnershipSetUsernameFromContext(ctx context.Context, srcOwnership *Ownership) *Ownership {
+
```
lpabon: delete this line

api/ownership.go
```diff
@@ -34,9 +34,14 @@ const (
...
 	// Check if the context has information about the user. If not,
 	// then security is not enabled.
 	if userinfo, ok := auth.NewUserInfoFromContext(ctx); ok {
+		// Public users cannot provide ownership
+		if userinfo.Public {
+			return srcOwnership
```
lpabon: Return nil to be clear

api/ownership.go
```diff
@@ -182,7 +187,8 @@ func (o *Ownership) HasAnOwner() bool {
 
 // IsPublic returns true if there is no ownership in this resource
 func (o *Ownership) IsPublic() bool {
-	return !o.HasAnOwner()
+	return (o.Acls != nil && o.Acls.Public != nil &&
```
lpabon: Simplify the code by adding it as a function `IsAccessPermittedByPublic(accessType)`

api/ownership.go
```diff
@@ -182,7 +187,8 @@ func (o *Ownership) HasAnOwner() bool {
 
 // IsPublic returns true if there is no ownership in this resource
 func (o *Ownership) IsPublic() bool {
```
lpabon: Adjust it to support accessType

api/server/sdk/server_interceptors.go
```diff
@@ -54,8 +54,22 @@ func (s *sdkGrpcServer) rwlockIntercepter(
 
 // Authenticate user and add authorization information back in the context
 func (s *sdkGrpcServer) auth(ctx context.Context) (context.Context, error) {
+	var token string
+	var err error
+
+	// public call attempted, add system.public user
+	if auth.IsPublic(ctx) {
+		return auth.ContextSaveUserInfo(ctx, &auth.UserInfo{
```
lpabon: Create a function in auth to return the UserInfo for public with the appropriate RBAC.

Pass in the value of the PublicVolumeCreationDisabled bool into the new function and let it set the rules for the system.public role accordingly.

api/server/sdk/server_interceptors.go
```diff
@@ -141,6 +155,12 @@ func (s *sdkGrpcServer) authorizationServerInterceptor(
 		"method":   info.FullMethod,
 	})
 
+	// If a public create call is attempted, but public volume creation is disabled, then deny
+	if auth.IsPublic(ctx) && s.config.Security.PublicVolumeCreationDisabled &&
```
lpabon: We can remove this and let the RBAC system take care of it

cmd/osd/main.go
```diff
@@ -483,9 +483,10 @@ func start(c *cli.Context) error {
 			Cluster:       cm,
 			StoragePolicy: sp,
 			Security: &sdk.SecurityConfig{
-				Role:           rm,
-				Tls:            tlsConfig,
-				Authenticators: authenticators,
+				Role:                         rm,
+				Tls:                          tlsConfig,
+				Authenticators:               authenticators,
+				PublicVolumeCreationDisabled: c.Bool("publicvolumecreationdisabled"),
```
lpabon: `--public-volume-create-allowed=true` and set the default to `true`, then use `!c.Bool('...')`

pkg/auth/token.go
```diff
@@ -109,3 +111,7 @@ func Token(
 
 	return signedtoken, nil
 }
+
+func IsPublic(ctx context.Context) bool {
+	return metautils.ExtractIncoming(ctx).Get("authorization") == ""
```
lpabon: Use a constant for the string and investigate that the key is case-insensitive

pkg/role/sdkserviceapi.go
```diff
@@ -91,6 +91,19 @@ var (
 				},
 			},
 		},
+
+		// system:public role is used for any unauthenticated user.
+		// They can only use standard volume lifecycle commands.
+		"system.public": {
```
lpabon: Add support for `!` not allowed RBAC. `!*inspect*`. Remove `!` then send the string to `match(r[1:], s)`

```		"system.public": {
			&api.SdkRule{
				Services: []string{"volume"},
				Apis:     []string{"!create", "*"},
			},
			&api.SdkRule{
				Services: []string{"mountattach", "cloudbackup", "migrate"},
				Apis:     []string{"*"},
			},
			&api.SdkRule{
				Services: []string{"identity"},
				Apis:     []string{"version"},
			},
		},
```

Update the documentation in the Role API in the api.proto to support this, and show examples.

pkg/auth/token.go
ggriffiths: it does toLower: https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/util/metautils/single_key.go#L16

pkg/role/sdkserviceapi.go
ggriffiths: This will happen in a separate PR.

api/server/sdk/server_interceptors.go
ggriffiths: We decided to check PublicVolumeCreationDisabled on server startup and updating the system.public user RBAC accordingly.

api/api.go
```diff
@@ -1131,7 +1130,7 @@ func (s *SdkStoragePolicy) IsPermitted(ctx context.Context, accessType Ownership
 }
 
 func (s *SdkStoragePolicy) IsPermittedFromUserInfo(user *auth.UserInfo, accessType Ownership_AccessType) bool {
-	if s.IsPublic() {
+	if s.IsPublic(Ownership_Read) {
```
ggriffiths: I chose to still pass in Ownership_Read for the Public checks, since Public is always Ownership_Read, right?

ggriffiths:
Addressed feedback @lpabon , but have a question about the storage policy IsPublic function.

api/api.go
ggriffiths: Updated all of these to pass in accessType instead.

api/ownership.go
```diff
@@ -37,6 +37,10 @@ func OwnershipSetUsernameFromContext(ctx context.Context, srcOwnership *Ownershi
 	// Check if the context has information about the user. If not,
 	// then security is not enabled.
 	if userinfo, ok := auth.NewUserInfoFromContext(ctx); ok {
+		// Public users cannot provide ownership
+		if userinfo.Public {
```
lpabon: This probably should be an accessor func

api/server/sdk/credentials.go
```diff
@@ -500,7 +500,7 @@ func (s CredentialServer) isPermitted(
 	}
 
 	// If ownership is missing then it is also public
-	if ownership == nil || ownership.IsPublic() {
+	if ownership == nil || ownership.IsPublic(api.Ownership_Read) {
```
lpabon: I think here we need to pass accessType

api/ownership.go
ggriffiths: Added UserInfo.IsPublic()

api/server/sdk/credentials.go
ggriffiths: Done

api/server/sdk/server_interceptors.go
ggriffiths: This will be refactored once we have support for negation in the RBAC system.

PR #1299, opened 2019-11-08
CSI: Unmount to not fail when volume is not found or deleted

ggriffiths:
Before, we were erroring out if unmount could not find the volume ID. 

Closes #

### DIFF
--- a/csi/node.go
+++ b/csi/node.go
@@ -21,7 +21,7 @@ import (
 	"os"
 
 	"github.com/libopenstorage/openstorage/api"
-	"github.com/libopenstorage/openstorage/pkg/util"
+	"github.com/portworx/kvdb"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi"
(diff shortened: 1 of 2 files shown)


csi/node.go
```diff
@@ -159,9 +159,8 @@ func (s *OsdCsiServer) NodeUnpublishVolume(
 	// Get volume information
 	vol, err := util.VolumeFromName(s.driver, req.GetVolumeId())
 	if err != nil {
-		return nil, status.Errorf(codes.NotFound, "Volume id %s not found: %s",
-			req.GetVolumeId(),
-			err.Error())
+		logrus.Infof("Volume %s was deleted or cannot be found: %s", req.GetVolumeId(), err.Error())
```
harsh-px: instead of ignoring all errors, let's only ignore volume not found errors.

For all other errors, we should fail as before.

ggriffiths: got it - will check for kvdb.ErrNotFound

PR #1308, opened 2019-11-14
Add label key for pool location constraint

harsh-px:

Will be using this to enumerate volumes based on pool uuids.

Closes #

### DIFF
--- a/volume/volume.go
+++ b/volume/volume.go
@@ -59,6 +59,9 @@ const (
 const (
 	// LocationConstaint is a label that specifies data location constraint.
 	LocationConstraint = "LocationConstraint"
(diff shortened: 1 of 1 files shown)


sangleganesh:
why not use existing LocationContraint ?

harsh-px:
> why not use existing LocationContraint ?

Since that is already being processed for node IDs.  I can overload LocationContraint to contain both node IDs or pool UIDs and compare the string specified in LocationContraint with both node IDs and the volumes pool IDs. But with that, the parsing code won't look clean. Opted for a more specific check.

Let me know if you have a strong preference.

sangleganesh:
we can continue to LocationConstraint, since it does not specify node id only.

harsh-px:
Using existing LocationConstraint instead of adding a new key

PR #1327, opened 2019-11-22
Add volumeSpec and locator cloudbackupRestore api

talakad:

Allows  more user control while restoring cloudbackups ( w.r.t to volume provisioning) 
Closes #


talakad:
Some of fields in volume spec will not applicable for restoring cloud backups( such as FS, block size etc) and hence users cannot specify them. Since it would be confusing for users as to what fields they can provide in volume spec, decided to create a new spec for this purpose. It would be a subset of volume spec and include only applicable/allowed fields.  Hence closing this PR.

PR #1329, opened 2019-11-23
Convert cred handlers to use SDK calls

stgleb:
Use SDK instead of objects

### DIFF
--- a/api/server/creds.go
+++ b/api/server/creds.go
@@ -3,6 +3,7 @@ package server
 import (
 	"encoding/json"
 	"net/http"
+	"strconv"
 
 	"github.com/gorilla/mux"
 	"github.com/libopenstorage/openstorage/api"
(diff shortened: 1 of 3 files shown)


api/server/creds.go
```diff
@@ -11,61 +12,155 @@ import (
...
+
+	if err != nil {
+		vd.sendError(vd.name, method, w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	credentials := make(map[string]interface{})
+
+	for _, credentialId := range resp.CredentialIds {
+		credResp, err := client.Inspect(r.Context(), &api.SdkCredentialInspectRequest{
```
lpabon: @stgleb I don't think this is correct. The context in line 15 is what you want. That has the Auth token. Did this work in your tests?

api/server/creds.go
```diff
@@ -11,61 +12,155 @@ import (
...
+	conn, err := vd.getConn()
 	if err != nil {
 		vd.sendError(vd.name, method, w, err.Error(), http.StatusInternalServerError)
 		return
 	}
-	json.NewEncoder(w).Encode(creds)
+
+	client := api.NewOpenStorageCredentialsClient(conn)
+	resp, err := client.Enumerate(ctx, &api.SdkCredentialEnumerateRequest{})
+
```
lpabon: nit: Remove empty line

api/server/creds.go
```diff
@@ -11,61 +12,155 @@ import (
...
+	if len(input.InputParams[api.OptCredProxy]) != 0 {
+		useProxy, err = strconv.ParseBool(input.InputParams[api.OptCredProxy])
+
+		if err != nil {
+			vd.sendError(vd.name, method, w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+
+	ownerShip, err := getOwnershipFromString(input.InputParams[api.OptCredOwnership])
```
lpabon: nit: ownership is one word

api/server/creds.go
```diff
@@ -11,61 +12,155 @@ import (
...
+
+	if err != nil {
```
lpabon: FYI, i see that this is your style, but most of the golang code I have see, the `if err != nil` is just below the function call, without an empty line

api/server/creds.go
```diff
@@ -11,61 +12,155 @@ import (
...
+	// Get gRPC connection
+	conn, err := vd.getConn()
 	if err != nil {
 		vd.sendError(vd.name, method, w, err.Error(), http.StatusInternalServerError)
 		return
 	}
-	json.NewEncoder(w).Encode(response)
+
+	client := api.NewOpenStorageCredentialsClient(conn)
+	resp, err := client.Create(ctx, sdkReq)
```
lpabon: err is lost here. it is replaced by line 120. Please return if error is found

api/server/creds.go
lpabon: You can remove this from the REST call. This was not supported through the old REST server. It is only for the SDK

api/server/util.go
```diff
@@ -0,0 +1,28 @@
+package server
```
lpabon: You can remove this entire file. No need to support ownership in the old REST api.

api/server/creds.go
stgleb: Thank you, I've missed that. I'll test it again.

api/server/creds.go
stgleb: Thats weird, in mu IDE there is no blank line.

api/server/creds.go
stgleb: Yes, I've missed that case since I did not create any credentials before I test enumerate. My fault.

api/server/creds.go
stgleb: gotcha

PR #1344, opened 2019-12-10
Add support for filesystem-trim and filesystem-check

vjs-px:
This changes adds support for filesystem-trim and filesystem-check
The change include Service definition, API definition and Struct defintions.

Signed-off-by: vjs-px <vjs@portworx.com>

Closes #


api/api.proto
```diff
@@ -1203,6 +1203,83 @@ service OpenStorageRole {
...
+        post: "/v1/filesystem-trim/start"
+        body: "*"
+      };
+    }
+
+  // Get Status of a filesystem Trim background operation on a mounted volume, if any
+  rpc GetStatus(SdkFilesystemTrimGetStatusRequest)
+    returns (SdkFilesystemTrimGetStatusResponse){
+      option(google.api.http) = {
+        get: "/v1/filesystem-trim/getstatus/query/{volume_id}/{mount_path}"
```
lpabon: This should probably be `get: "/v1/filesystem-trim/getstatus"`. The caller would then send `"GET" /v1/filesystem-trim/getstatus?volume_id=myid&mount_path=mypath`

api/api.proto
```diff
@@ -1203,6 +1203,83 @@ service OpenStorageRole {
...
+    }
+  }
+
+// OpenStorageFilesystemCheckService service provides methods to manage a 
+// filesystem check operation on a volume. This operation is run in the 
+// background on an unmounted volume. If the volumes are mounted, then these
+// APIs return error.
+service OpenStorageFilesystemCheckService {
+
+  // Get a report of issues found on the filesystem. This operation works on an unmounted volume.
```
lpabon: Should there be a comment here saying to use `Fix()` first?

api/api.proto
```diff
@@ -1203,6 +1203,83 @@ service OpenStorageRole {
...
+        post: "/v1/filesystem-check/fix"
+        body: "*"
+      };
+    }
+
+  // Get Status of a filesystem check background operation on an unmounted volume, if any
+  rpc GetStatus(SdkFilesystemCheckGetStatusRequest)
+    returns (SdkFilesystemCheckGetStatusResponse){
+      option(google.api.http) = {
+        get: "/v1/filesystem-check/getstatus/query/{volume_id}"
```
lpabon: This can be `/v1/filesystem-check/getstatus/{volume_id}`

api/api.proto
```diff
@@ -3454,6 +3531,110 @@ message SdkRoleUpdateResponse {
...
 
+// SdkFilesystemTrimStartRequest defines a request to start a background filesystem trim operation
+message SdkFilesystemTrimStartRequest {
+  // Id of the volume
+  string volume_id = 1;
+  // Path where the volume is mounted
+  string mount_path = 2;
+}
+
+// SdkFilesystemTrimStartResponse defines the response for a SdkFilesystemTrimStartRequest. If case of error, this response struct is not valid
```
lpabon: I think you can remove all the `If case of error...` documentation for all of these messages since these are not RPCs, just messages.

api/api.proto
```diff
@@ -3454,6 +3531,110 @@ message SdkRoleUpdateResponse {
...
+// SdkFilesystemTrimStopRequest defines a request to stop a background filesystem trim operation
+message SdkFilesystemTrimStopRequest {
+  // Id of the volume
+  string volume_id = 1;
+  // Path where the volume is mounted
+  string mount_path = 2;
+}
+
+// SdkFilesystemTrimStopResponse defines the response for a SdkFilesystemTrimStopRequest. If case of error, this response struct is not valid
+message SdkFilesystemTrimStopResponse{
```
lpabon: Why would the stop response return information? Shouldn't it just be error or nil?

api/api.proto
```diff
@@ -3454,6 +3531,110 @@ message SdkRoleUpdateResponse {
...
+  // Id of the volume
+  string volume_id = 1;
+}
+
+// SdkFilesystemCheckGetStatusResponse defines the response for a SdkFilesystemCheckGetStatusRequest. If case of error, this response struct is not valid
+message SdkFilesystemCheckGetStatusResponse{
+  // Id of the volume
+  string volume_id = 1;
+  // Status of the background filesystem check GetStatus request
+  string status = 2;
```
lpabon: `status` should be some type of enum. Maybe add a `message` for a string returned from the code.

api/api.proto
```diff
@@ -3454,6 +3531,110 @@ message SdkRoleUpdateResponse {
...
+}
+
+// SdkFilesystemCheckStopRequest defines a request to stop a background filesystem check operation
+message SdkFilesystemCheckStopRequest {
+  // Id of the volume
+  string volume_id = 1;
+}
+
+// SdkFilesystemCheckStopResponse defines the response for a SdkFilesystemCheckStopRequest. If case of error, this response struct is not valid
+message SdkFilesystemCheckStopResponse{
```
lpabon: This probably should be error or nil

api/api.proto
lpabon: enum displaying job status

api/api.proto
```diff
@@ -3454,6 +3531,110 @@ message SdkRoleUpdateResponse {
...
+  string status = 2;
+}
+
+// SdkFilesystemCheckFixRequest defines a request to fix all the filesystem consistency issues reported in the response to OpenStorageFilesystemCheckService.Report() request.
+message SdkFilesystemCheckFixRequest {
+  // Id of the volume
+  string volume_id = 1;
+}
+
+// SdkFilesystemCheckFixResponse defines the response for a SdkFilesystemCheckFixRequest. If case of error, this response struct is not valid
```
lpabon: Add a comment that the developer should call GetStatus until completion

api/api.proto
```diff
@@ -3454,6 +3531,110 @@ message SdkRoleUpdateResponse {
...
+message SdkFilesystemCheckFixResponse{
```
lpabon: err / nil

api/api.proto
vjs-px: changed it a to post request

api/api.proto
vjs-px: fixing it to return a status code

api/api.proto
disrani-px: As I've pointed out in previous PRs (where it is also done incorrectly), we shouldn't be using PUT or POST for read APIs.
This is clearly explained here: https://www.w3schools.com/tags/ref_httpmethods.asp

Also, you shouldn't have the verb in the path itself. It should be a `GET`  on `/v1/filesystem-check/{volume_id}/status`

api/api.proto
```diff
@@ -1203,6 +1203,130 @@ service OpenStorageRole {
...
+        body: "*"
+      };
+    }
+
+  // Get Status of a filesystem Trim background operation on a mounted
+  // volume, if any
+  rpc GetStatus(SdkFilesystemTrimGetStatusRequest)
+    returns (SdkFilesystemTrimGetStatusResponse){
+      option(google.api.http) = {
+        post: "/v1/filesystem-trim/get_status"
```
vjs-px: replace get_status to get-status

api/api.proto
```diff
@@ -3454,6 +3578,165 @@ message SdkRoleUpdateResponse {
   SdkRole role = 1;
 }
 
+// FilesystemTrimStatus represents the status codes returned from
+// OpenStorageFilesystemTrim service APIs()
+enum FilesystemTrimStatus {
+  // Filesystem Trim operation is not running for the specified volume
+  FS_TRIM_NOT_RUNNING = 0;
```
vjs-px: Add UNKNOWN option and never start any valid value from zero

api/api.proto
```diff
@@ -3454,6 +3578,165 @@ message SdkRoleUpdateResponse {
...
+  // Filesystem Trim operation started for the specified volume
+  FS_TRIM_STARTED = 1;
+  // Filesystem Trim operation is in progress for the specified volume
+  FS_TRIM_INPROGRESS = 2;
+  // Filesystem Trim operation was aborted for the specified volume
+  FS_TRIM_ABORTED = 3;
```
vjs-px: STOPPED

api/api.proto
```diff
@@ -1203,6 +1203,143 @@ service OpenStorageRole {
...
+// This operation is run in the background on an **unmounted volume**.
+// If the volume is mounted, then these APIs return error.
+//
+// Once the filesystem check operation(either CheckHealth() or FixAll()) is
+// started, the clients have to poll for the status of the background operation
+// using the `OpenStorageFilesystemcheck.CheckHealthGetStatus()` rpc request or
+// `OpenStorageFilesystemCheck.FixAllGetStatus()` rpc request.
+//
+// **Note: CheckHealth() and FixAll() cannot run in parallel for the same volume**
+
```
lpabon: need a `//` here to show that this is still part of the same documentation. If not, the documentation will start with line 1276. We can fix this after. Please feel free to merge.

api/api.proto
vjs-px: I have removed "get" verb from the REST API path

There are 2 parameters expected for this API
1. volumeid
2. mountpath
Having mountpath as a query parameter for get would require the client to encode the request
we decided it would be easier for the client and also from the server to deal with mountpath as parameter given as part of the POST request

Let me know if you still insist that we need this API to use the GET verb

If you s

vjs-px: discussed with dinesh. Finally we concluded to use GET request instead of POST for all the status APIs

PR #1360, opened 2020-01-02
R7.0 fscheck fstrim

vjs-px:
This PR adds support to do run fsck and fstrim on portworx volumes

Closes #


lpabon:
You need to create a PR against release-7.0, not master. that is why it looks incorrect

PR #1371, opened 2020-01-20
start watch from index 0.

sangleganesh:
if we get entry from kvdb, watch index should be 0 to start
watching new updates.

Signed-off-by: ganesh <ganesh@portworx.com>

Closes #


sangleganesh:
> Starting the watch on 0 means we will be missing out on watch updates, which could mean we might miss out on a node Decommission.
iirc there is no way to identify missed updates if we cannot start from the index which has been compacted.
Since on the next update we will reconcile the db again, there is a very small window when it would be lagging behind (watch returning error, getting entire db, starting watch again) until the next update.

PR #1387, opened 2020-01-27
Add RestoreVolume spec for cloudbackup restore api

talakad:

Closes #
Allows users to modify some of the volume spec related fields for the volume being restored from cloudbackup. (Current scheme does not support things like specifying HA level different than original volume)


api/api.proto
```diff
@@ -4450,3 +4456,57 @@ enum EnforcementType {
...
+  // Aggregation level Specifies the number of parts the volume can be aggregated from.
+  uint32 aggregation_level = 7;
+  // SnapshotSchedule a well known string that specifies when snapshots should be taken.
+  string snapshot_schedule = 8;
+  // Sticky volumes cannot be deleted until the flag is removed.
+  RestoreParamBoolType sticky = 9;
+  // Group identifies a consistency group
+  Group group = 10;
+  // GroupEnforced is true if consistency group creation is enforced.
+  bool group_enforced = 11;
```
disrani-px: Should be of type `RestoreParamBoolType`?

api/api.proto
```diff
@@ -4450,3 +4456,57 @@ enum EnforcementType {
...
+	PARAM_TRUE = 2;
+}
+
+// RestoreSpec allows some of the restore volume properties of to be modified
+// while restoring cloud baackup.
+message RestoreVolumeSpec {
+  // HaLevel specifies the number of copies of data.
+  int64 ha_level = 1;
+  // Cos specifies the relative class of service.
+  CosType cos = 2;
```
disrani-px: Needs a value similar to io_profile to pick up from backup source?

api/api.proto
```diff
@@ -4450,3 +4456,57 @@ enum EnforcementType {
...
+  // ReplicaSet is the desired set of nodes for the volume data.
+  ReplicaSet replica_set = 6;
+  // Aggregation level Specifies the number of parts the volume can be aggregated from.
+  uint32 aggregation_level = 7;
+  // SnapshotSchedule a well known string that specifies when snapshots should be taken.
+  string snapshot_schedule = 8;
+  // Sticky volumes cannot be deleted until the flag is removed.
+  RestoreParamBoolType sticky = 9;
+  // Group identifies a consistency group
+  Group group = 10;
```
disrani-px: How do you differentiate if this should be picked from the backup or not used?

api/api.proto
```diff
@@ -4450,3 +4456,57 @@ enum EnforcementType {
...
+  uint32 queue_depth = 14;
+  // Nodiscard specifies if the volume will be mounted with discard support disabled.
+  // i.e. FS will not release allocated blocks back to the backing storage pool.
+  RestoreParamBoolType nodiscard = 15;
+  // IoStrategy preferred strategy for I/O.
+  IoStrategy io_strategy = 16;
+  // PlacementStrategy specifies a spec to indicate where to place the volume.
+  VolumePlacementStrategy placement_strategy = 17;
+  // StoragePolicy if applied/specified while creating volume
+  string storage_policy = 18;
```
disrani-px: How do you differentiate if this should be picked from the backup or not used?

disrani-px: Same for ownership and export_spec

api/api.proto
talakad: No.  group_enforced is applicable  only if group is specified. And if specified, it overwrites the backup's group field

api/api.proto
talakad: this field has None value that it can be set to if user wishes to fallback whats in backup's spec.

api/api.proto
talakad: if specified, it will be used. if not backup's spec is used.

api/api.proto
talakad: all strings if non empty will be used to overwrite backup spec's respective fields.  Similarly ownershop and export_spec will used if provided. None must be provided if backup spec must be used.

api/api.proto
disrani-px: But what if the backup spec has a group and a user doesn't want to use that for the restore?

talakad: in that case user specifies a group in RestoreSpec. where that Group spec is enaforced or not is specified by enforce_group parameter. IOW, if group is specified in restore spec, it will be used rather than backup spec's group field.

talakad: to clarfy, if enforce_group is set to yes, then restore may fail if the group requirements cant be met.

disrani-px: Same question for ownership and storage_policy too then. BackupSpec has a storagepolicy, if we don't want to use a storage policy during restore how will a user specify that? They can't leave it empty because it'll pick up the one from the Backup

talakad: I see.  what are you asking is ignore fields for storage policy and ownership. I will add those. enforce_group would be honored only when group is specified.   From this discussion, I see the need for ignoreGroup as well. All the ignore fields will only be honored when user does not specify the relevant fields in restore spec.

talakad: Done. Added ignore fields for the above.

disrani-px: I think this is error prone. Can those be made pointers? If it is nil then you can use the value from the Backup. If it is not nil then use the value from the variable (it can be empty or have a valid value)

talakad: Ok. I reverted the changes. Those fields are pointers.

PR #1395, opened 2020-01-29
Add api to sched package

talakad:
Returns number of intervals per day for given schedule items.

Adds new api to sched package that returns the number intervals per day.
Useful  for estimating number of times the schedule would be exercised per day.
Closes #

### DIFF
--- a/pkg/sched/intervals.go
+++ b/pkg/sched/intervals.go
@@ -13,6 +13,7 @@ import (
 const (
 	scheduleSeparator    = ";"
 	nonYamlTypeSeparator = "="
+	MinutesPerDay        = float64(60 * 24)
 	DailyType            = "daily"
 	MonthlyType          = "monthly"
 	WeeklyType           = "weekly"
@@ -314,20 +315,18 @@ func ScheduleIntervalSummary(items []Interval, policyTags *PolicyTags) string {
 	return summary
 }
 
-func ScheduleIntervalsPerDay(items []RetainInterval, policyTags *PolicyTags) uint32 {
-	if len(items) == 0 {
-		return 0
-	}
+func MaxPerDayInstances(items []RetainInterval) uint32 {
 	var intervalsPerDay uint32
 	for _, iv := range items {
 		var perDay uint32
 		switch iv.Spec().Freq {
 		case PeriodicType:
+			perDay = uint32(1)
 			period := time.Duration(iv.Spec().Period).Minutes()
-			if period < float64(60*24) {
-				perDay = uint32(float64(60*24) / period)
-			} else {
-				perDay = uint32(1)
+			if period == 0 {
+				perDay = 0
+			} else if period < MinutesPerDay {
+				perDay = uint32(MinutesPerDay / period)
 			}
 		case DailyType:
 			perDay = uint32(1)
@@ -341,10 +340,7 @@ func ScheduleIntervalsPerDay(items []RetainInterval, policyTags *PolicyTags) uin
 	return intervalsPerDay
 }
 
-func ScheduleTotalRetains(items []RetainInterval, policyTags *PolicyTags) uint32 {
-	if len(items) == 0 {
-		return 0
-	}
+func ScheduleRetainSum(items []RetainInterval) uint32 {
 	var totalRetains uint32
 	for _, iv := range items {
 		totalRetains += iv.RetainNumber()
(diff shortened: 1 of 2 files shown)


pkg/sched/intervals.go
```diff
@@ -314,6 +314,44 @@ func ScheduleIntervalSummary(items []Interval, policyTags *PolicyTags) string {
...
+func ScheduleIntervalsPerDay(items []RetainInterval, policyTags *PolicyTags) uint32 {
+	if len(items) == 0 {
+		return 0
+	}
+	var intervalsPerDay uint32
+	for _, iv := range items {
+		var perDay uint32
+		switch iv.Spec().Freq {
+		case PeriodicType:
+			period := time.Duration(iv.Spec().Period).Minutes()
```
sangleganesh: period can be zero if Period < 60 seconds.

pkg/sched/intervals.go
```diff
@@ -314,6 +314,44 @@ func ScheduleIntervalSummary(items []Interval, policyTags *PolicyTags) string {
...
+			perDay = uint32(1)
+		case MonthlyType:
+			perDay = uint32(1)
+		}
+		intervalsPerDay += perDay
+	}
+	return intervalsPerDay
+}
+
+func ScheduleTotalRetains(items []RetainInterval, policyTags *PolicyTags) uint32 {
```
sangleganesh: policyTags is not used ?
May be just call RetainSum()

pkg/sched/intervals.go
```diff
@@ -314,6 +314,44 @@ func ScheduleIntervalSummary(items []Interval, policyTags *PolicyTags) string {
...
+	if len(items) == 0 {
```
sangleganesh: its ok to iterate over nil/0 Len range, so can skip this check

pkg/sched/intervals.go
```diff
@@ -314,6 +314,44 @@ func ScheduleIntervalSummary(items []Interval, policyTags *PolicyTags) string {
 	return summary
 }
 
+func ScheduleIntervalsPerDay(items []RetainInterval, policyTags *PolicyTags) uint32 {
```
sangleganesh: May be call this MaxPerDayInstances - since this is the max and not per day thing ? 
policyTags is not used ? can skip len(items) and do the range always

pkg/sched/intervals.go
```diff
@@ -314,6 +314,44 @@ func ScheduleIntervalSummary(items []Interval, policyTags *PolicyTags) string {
...
+	if len(items) == 0 {
+		return 0
+	}
+	var intervalsPerDay uint32
+	for _, iv := range items {
+		var perDay uint32
+		switch iv.Spec().Freq {
+		case PeriodicType:
+			period := time.Duration(iv.Spec().Period).Minutes()
+			if period < float64(60*24) {
```
sangleganesh: nit: can store float64(60*24) as a const.

pkg/sched/sched_test.go
```diff
@@ -183,10 +183,35 @@ func TestScheduleStrings(t *testing.T) {
...
+		case WeeklyType:
+			require.Equal(t, perDay, uint32(1), "Incorrect Weekly intervals per day")
+		case MonthlyType:
+			require.Equal(t, perDay, uint32(1), "Incorrect Monthly intervals per day")
+		}
 	}
+	cumulativeSched := dockSched[0] + scheduleSeparator +
+		dockSched[1] + scheduleSeparator +
+		dockSched[2] + scheduleSeparator +
+		dockSched[3]
```
sangleganesh: nit: use strings.Join(dockSched, scheduleSeparator)

pkg/sched/intervals.go
talakad: Thanks. Assumed Period will at least a a minute. Fixed now.

pkg/sched/intervals.go
talakad: Done

pkg/sched/intervals.go
talakad: Done

pkg/sched/intervals.go
talakad: Done

pkg/sched/sched_test.go
talakad: Done.

PR #1399, opened 2020-01-30
NFS Raw block support

lpabon:

For CSI Raw block testing

Volume Creation:
Will attach to the NFS host, then create a file of the requested
size on the NFS share. It will then use losetup to mount the file
and format it according to the format in the request. If not format
was requested, then it will be left as RAW.

Snapshots are done by copying the file.

### DIFF
--- a/volume/drivers/nfs/nfs.go
+++ b/volume/drivers/nfs/nfs.go
@@ -107,7 +107,7 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 
 	//mount each nfs server
 	for _, v := range inst.nfsServers {
-		nfsServer := v + src
+		nfsServer := src
 		nfsMountPoint := nfsMountPath + v
 		inst.mounter.Reload(nfsServer)
 
--- a/volume/drivers/nfs/nfs.go
+++ b/volume/drivers/nfs/nfs.go
@@ -118,25 +118,38 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 		}
 		if !mounted {
 			if server != "" {
-				err = inst.mounter.Mount(
-					0,
-					nfsServer,
-					nfsMountPoint,
+				/*
+					err = inst.mounter.Mount(
+						0,
+						nfsServer,
+						nfsMountPoint,
+						"nfs",
+						0,
+						"user_xattr,nolock,addr="+v,
+						0,
+						nil)
+				*/
+				err = syscall.Mount(
+					src,
+					nfsMountPath+v,
 					"nfs",
 					0,
-					"user_xattr,nolock,addr="+v,
-					0,
-					nil)
+					"nolock,addr="+v,
+				)
+
 			} else {
-				err = inst.mounter.Mount(
-					0,
-					nfsServer,
-					nfsMountPoint,
-					"",
-					syscall.MS_BIND,
-					"user_xattr",
-					0,
-					nil)
+				err = syscall.Mount(src, nfsMountPath+v, "", syscall.MS_BIND, "")
+				/*
+					err = inst.mounter.Mount(
+						0,
+						nfsServer,
+						nfsMountPoint,
+						"",
+						syscall.MS_BIND,
+						"user_xattr",
+						0,
+						nil)
+				*/
 			}
 			if err != nil {
 				logrus.Errorf("Unable to mount %s:%s at %s (%+v)",
--- a/volume/drivers/nfs/nfs.go
+++ b/volume/drivers/nfs/nfs.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path"
 	"strconv"
 	"syscall"
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+	losetup "gopkg.in/freddierice/go-losetup.v1"
 
 	"math/rand"
 	"strings"
@@ -19,6 +22,7 @@ import (
 	"github.com/libopenstorage/openstorage/config"
 	"github.com/libopenstorage/openstorage/pkg/mount"
 	"github.com/libopenstorage/openstorage/pkg/seed"
+	"github.com/libopenstorage/openstorage/pkg/util"
 	"github.com/libopenstorage/openstorage/volume"
 	"github.com/libopenstorage/openstorage/volume/drivers/common"
 	"github.com/pborman/uuid"
@@ -27,10 +31,12 @@ import (
 
 const (
 	Name         = "nfs"
-	Type         = api.DriverType_DRIVER_TYPE_FILE
 	NfsDBKey     = "OpenStorageNFSKey"
 	nfsMountPath = "/var/lib/openstorage/nfs/"
 	nfsBlockFile = ".blockdevice"
+
+	// Set to block, but it will handle size 0 as file based
+	Type = api.DriverType_DRIVER_TYPE_BLOCK
 )
 
 // Implements the open storage volume interface.
@@ -60,7 +66,6 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 	} else {
 		logrus.Printf("NFS driver initializing with %s:%s ", server, path)
 	}
-
 	//support more than one server using CSV
 	//TB-FIXME: modify driver params flow to support map[string]struct/array
 	servers := strings.Split(server, ",")
@@ -71,6 +76,7 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 		logrus.Warnf("Failed to create mount manager for server: %v (%v)", server, err)
 		return nil, err
 	}
+
 	inst := &driver{
 		IODriver:              volume.IONotSupported,
 		StoreEnumerator:       common.NewDefaultStoreEnumerator(Name, kvdb.Instance()),
@@ -89,10 +95,11 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 	//make directory for each nfs server
 	for _, v := range servers {
 		logrus.Infof("Calling mkdirAll: %s", nfsMountPath+v)
-		if err := os.MkdirAll(nfsMountPath+v, 0744); err != nil {
+		if err := os.MkdirAll(nfsMountPath+v, 0755); err != nil {
 			return nil, err
 		}
 	}
+
 	src := inst.nfsPath
 	if server != "" {
 		src = ":" + inst.nfsPath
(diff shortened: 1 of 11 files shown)


lpabon:
# How it was tested

## Starting the server

I created an NFS share on the development host that exposed `/nfs` share. I then ran the following command to start OSD:

```
make install
sudo $GOPATH/bin/osd -d --driver=name=nfs,server=127.0.0.1,path=/nfs --sdkrestport=9910
```

## Tests

Then I went on another terminal and ran the following commands:

### Create

#### RAW format

```
curl -X POST "http://localhost:9910/v1/volumes" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"name\":\"nfs1\",\"spec\":{\"size\":\"123412341234\"}}"
```

#### EXT4

```
curl -X POST "http://localhost:9910/v1/volumes" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"name\":\"nfs-formatted\",\"spec\":{\"format\":\"FS_TYPE_EXT4\",\"size\":\"12345432\"}}"
```

Then I save the volume id on an environment variable called `vid`

### Attach

```
curl -X POST "http://localhost:9910/v1/mountattach/attach" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"volume_id\":\"$vid\"}"
```

### Mount

```
curl -X POST "http://localhost:9910/v1/mountattach/mount" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"mount_path\":\"/mnt/nfs\",\"volume_id\":\"$vid\"}"

```

### Unmount

```
curl -X POST "http://localhost:9910/v1/mountattach/unmount" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"mount_path\":\"/mnt/nfs\",\"volume_id\":\"$vid\"}"
```

### Detach

```
curl -X POST "http://localhost:9910/v1/mountattach/detach" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"volume_id\":\"$vid\"}"
```

lpabon:
If you are curious, this is what the NFS mount could look like with a few volumes:

```
$ ls /nfs
21684c25-b98a-40b2-861b-35e467538853.blockdevice  50a5ca92-1c94-4fb4-8ab6-c9c009859c4c.blockdevice  5bba393c-04db-49be-9308-49e997383981.blockdevice  b007dd9b-ac8f-472c-afcf-0d02bfeef12b.blockdevice
2e4565f4-5b23-4a9a-8766-f8b73a25f9a5.blockdevice  514dbf8c-7c98-49de-8c50-8537964a51b6.blockdevice  aa7c21e0-236b-435a-8f1c-7dcfbcc15525.blockdevice
```

lpabon:
Now it does not allow mounting of raw volumes:

```
$ curl -X POST "http://70.0.0.129:9910/v1/mountattach/mount" -H "accept: application/json" -H "Content-Type: application/json" -d "{\"mount_path\":\"/mnt/nfs\",\"volume_id\":\"$vid\"}"        {"error":"Failed to mount volume 643d49a2-d48b-43ae-a358-405291a41387: Volume of raw format cannot be mounted","code":13}
```

volume/drivers/nfs/nfs.go
```diff
@@ -101,10 +106,8 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 	//mount each nfs server
 	for _, v := range inst.nfsServers {
 		// If src is already mounted at dest, leave it be.
-		mountExists, err := mounter.Exists(src, nfsMountPath+v)
-		if !mountExists {
-			// Mount the nfs server locally on a unique path.
-			syscall.Unmount(nfsMountPath+v, 0)
+		o, err := exec.Command("/bin/mount", "-t", "nfs").CombinedOutput()
```
ggriffiths: should we be checking this error?

volume/drivers/nfs/nfs.go
```diff
@@ -285,31 +292,121 @@ func (d *driver) Create(
...
+		v = common.NewVolume(
+			volumeID,
+			api.FSType_FS_TYPE_NFS,
+			locator,
+			source,
+			spec,
+		)
+
+		// Create a directory on the NFS server with this UUID.
+		err := os.MkdirAll(volPath, 0744)
```
ggriffiths: I've seen 0750 used for directory permissions in k8s.. not sure why above 744 but just a thought:
https://github.com/kubernetes/kubernetes/blob/master/pkg/volume/csi/csi_block.go#L166

volume/drivers/nfs/nfs.go
```diff
@@ -285,31 +292,121 @@ func (d *driver) Create(
...
+				return "", err
+			}
+		}
 
-	if err := d.CreateVol(v); err != nil {
-		return "", err
+		// Set to ready
+		v.State = api.VolumeState_VOLUME_STATE_AVAILABLE
+		if err := d.UpdateVol(v); err != nil {
+			d.Delete(v.GetId())
```
ggriffiths: if this fails - are we okay with ignoring the error? Maybe log that cleanup failed?

volume/drivers/nfs/nfs.go
lpabon: great point!

volume/drivers/nfs/nfs.go
```diff
@@ -285,31 +292,121 @@ func (d *driver) Create(
...
+		v.State = api.VolumeState_VOLUME_STATE_PENDING
+		if err := d.CreateVol(v); err != nil {
+			return "", err
+		}
+
+		// Check if this from as a clone
+		if source != nil && len(source.GetParent()) != 0 {
+			// Need to clone
+			if err := d.clone(volumeID, source.GetParent()); err != nil {
+				d.Delete(v.GetId())
```
ggriffiths: Same as above, maybe have a cleanupVolume helper if there's multiple occurrences of deleting after failed operations.

volume/drivers/nfs/nfs.go
```diff
@@ -285,31 +292,121 @@ func (d *driver) Create(
...
+				return "", err
+			}
+
+			// Format
+			if spec.GetFormat() != api.FSType_FS_TYPE_NONE {
+				dev, err := losetup.Attach(blockFile, 0, false)
+				if err != nil {
+					return "", err
+				}
+				defer dev.Detach()
```
ggriffiths: Can detach fail? Maybe have
```
defer func(){
  if err = dev.Detach(); err != nil {
    ..
  }

}()
```

volume/drivers/nfs/nfs.go
```diff
@@ -319,14 +416,18 @@ func (d *driver) Delete(volumeID string) error {
...
+		return fmt.Errorf("Volume is still attached and cannot be deleted")
+	}
 
 	nfsVolPath, err := d.getNFSVolumePath(v)
 	if err != nil {
 		return err
 	}
 
+	// Delete the simulated block volume
+	os.Remove(nfsVolPath + nfsBlockFile)
```
ggriffiths: Maybe check for errors on all os.Remove*? For pesky issues with permissions or the file system in general

volume/drivers/nfs/nfs.go
```diff
@@ -285,31 +294,131 @@ func (d *driver) Create(
...
+					logrus.Warnf("Failed to run command %s %s: %v\nOutput: %s", cmd, dev.Path(), err, string(o))
+					return "", err
+				}
+			}
+		}
+
+		// Set to ready
+		v.State = api.VolumeState_VOLUME_STATE_AVAILABLE
+		if err := d.UpdateVol(v); err != nil {
+			d.Delete(v.GetId())
```
ggriffiths: Can log delete error like others

ggriffiths:
Just a question on the 744 directory permissions, and the d.Delete err checks.

volume/drivers/nfs/nfs.go
```diff
@@ -285,31 +294,131 @@ func (d *driver) Create(
...
+				}
+				defer func() {
+					if err := dev.Detach(); err != nil {
+						logrus.Errorf("Failed to detach %s", dev)
+					}
+				}()
+				logrus.Infof("Formatting %s with %v", dev, spec.Format)
+
+				// Get mkfs
+				cmd, err := mkfsFormatTypeCmd(spec.Format.SimpleString())
```
adityadani: I thought we decide to make Mount operation as no-op for block devices. So we will never format the volume and treat a volume with FsType set as a regular volume.

volume/drivers/nfs/nfs.go
```diff
@@ -356,30 +476,54 @@ func (d *driver) Mount(volumeID string, mountpath string, options map[string]str
...
+		// Block access
+		if d.isRaw(v) {
+			return fmt.Errorf("Volume of raw format cannot be mounted")
+		}
+		if v.GetState() != api.VolumeState_VOLUME_STATE_ATTACHED {
+			return fmt.Errorf("Voume %s is not attached", volumeID)
+		}
+		mountExists, _ := d.mounter.Exists(v.DevicePath, mountpath)
+		if !mountExists {
+			if err := syscall.Mount(v.DevicePath, mountpath, v.Spec.Format.SimpleString(), 0, ""); err != nil {
```
adityadani: Mount should be no-op for raw block volumes.

volume/drivers/nfs/nfs.go
```diff
@@ -101,10 +106,10 @@ func Init(params map[string]string) (volume.VolumeDriver, error) {
 	//mount each nfs server
 	for _, v := range inst.nfsServers {
 		// If src is already mounted at dest, leave it be.
-		mountExists, err := mounter.Exists(src, nfsMountPath+v)
-		if !mountExists {
-			// Mount the nfs server locally on a unique path.
-			syscall.Unmount(nfsMountPath+v, 0)
+
+		// ignore the error and let it retry
+		o, _ := exec.Command("/bin/mount", "-t", "nfs").CombinedOutput()
```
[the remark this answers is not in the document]
lpabon: good question. i'll see if i can change it.

volume/drivers/nfs/nfs.go
lpabon: I added the log to d.Delete() itself so that all callers will have it log on error

volume/drivers/nfs/nfs.go
```diff
@@ -27,10 +31,12 @@ import (
 
 const (
 	Name         = "nfs"
-	Type         = api.DriverType_DRIVER_TYPE_FILE
 	NfsDBKey     = "OpenStorageNFSKey"
 	nfsMountPath = "/var/lib/openstorage/nfs/"
 	nfsBlockFile = ".blockdevice"
+
+	// Set to block, but it will handle size 0 as file based
+	Type = api.DriverType_DRIVER_TYPE_BLOCK
```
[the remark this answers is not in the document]
lpabon: in CSI

volume/drivers/nfs/nfs.go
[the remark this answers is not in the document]
lpabon: Ok, now I remember (I wrote this code in 2018). pkg.mount.Exists() only checks memory, not the system. And it was failing when I restarted the NFS driver because it thought it was not mounted. By running the mount command, I was able to parse the output and determine if the system thought it was mounted or not. Maybe this code can go into pkg mount?

sangleganesh:
@lpabon its unclear to me why we are changing nfs driver to support block devices, especially just for testing, why not just buse driver ?

volume/drivers/nfs/nfs.go
lpabon: For Raw, yes. For FS block volumes, we need to mount to format.

volume/drivers/nfs/nfs.go
lpabon: correct, it is returned at line 509.

lpabon:
>@lpabon its unclear to me why we are changing nfs driver to support block devices, especially just for testing, why not just buse driver ?
@sangleganesh Good question. For CSI we need a driver that can do Shared, Block, and Raw.

volume/drivers/nfs/nfs.go
```diff
@@ -234,6 +243,11 @@ func (d *driver) Create(
 		return "", fmt.Errorf("volume name cannot contain space characters")
 	}
 
+	// Fall back to older shared if Sharedv4 is requested
+	if spec.GetSharedv4() {
```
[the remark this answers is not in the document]
lpabon: fixed by adding a `isShared()` function

volume/drivers/nfs/nfs.go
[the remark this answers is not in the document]
lpabon: @adityadani and I talked and he showed how to do this using the pkg/mount library. I have updated the code to use that.

PR #1401, opened 2020-01-31
WIP - Ephemeral APIs review

ggriffiths:

* Need to add support for creating using volumes with CSI

**Notes**:
* This is just an API review - implementation will happen later


api/api.proto
```diff
@@ -2729,6 +2744,53 @@ message SdkVolumeDetachRequest {
 message SdkVolumeDetachResponse {
 }
 
+// Defines a request to setup an ephemeral volume
+message SdkVolumeSetupEphemeralRequest {
```
ggriffiths: This is essentially all parameters required to perform Create, Attach, and Mount.

I modelled TeardownEphemeralRequest the same way.

PR #1413, opened 2020-02-04
Add script to start OSD CSI Driver in kind

ggriffiths:

Closes #

### DIFF
--- a/cmd/osd/main.go
+++ b/cmd/osd/main.go
@@ -177,6 +177,11 @@ func main() {
 			Usage: "Cluster Domain Name",
 			Value: "",
 		},
+		cli.StringFlag{
+			Name:  "csidrivername",
+			Usage: "CSI Driver name",
+			Value: "",
+		},
 	}
 	app.Action = wrapAction(start)
 	app.Commands = []cli.Command{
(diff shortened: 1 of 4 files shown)


cmd/osd/main.go
```diff
@@ -177,6 +177,11 @@ func main() {
 			Usage: "Cluster Domain Name",
 			Value: "",
 		},
+		cli.StringFlag{
+			Name:  "csidrivername",
+			Usage: "CSI Driver name",
+			Value: "osd.openstorage.org",
```
lpabon: This value should be empty so that the csi server uses the name of the driver if CSIDriverName is empty

hack/setup-kind-csi-cluster.sh
```diff
@@ -0,0 +1,18 @@
...
+cd $GOPATH/src/github.com/libopenstorage/openstorage
+if [ $DELETE_CLUSTER ]; then
+	kind delete cluster --name kind-csi 
+fi
+kind create cluster --name kind-csi --config hack/kind.yaml --image kindest/node:v1.17.0
+kind_worker_id=$(docker inspect --format="{{.Id}}" kind-csi-control-plane)
+docker cp $GOPATH/bin/osd ${kind_worker_id}:/osd
+docker exec -ti ${kind_worker_id} bash -c "mkdir -p /var/lib/kubelet/plugins/osd.openstorage.org"
+docker exec -tid ${kind_worker_id} bash -c "pid=$(ps -ef | grep osd | grep -v grep | awk '{print $2}'); kill -9 $pid"
+docker exec -tid ${kind_worker_id} bash -c "CSI_ENDPOINT=/var/lib/kubelet/plugins/osd.openstorage.org/csi.sock /osd -d --driver=name=fake --sdkport 9106 --sdkrestport 9116"
```
lpabon: here pass in the csi name as `osd.openstorage.org`

cmd/osd/main.go
ggriffiths: Fixed

hack/setup-kind-csi-cluster.sh
ggriffiths: Fixed

PR #1424, opened 2020-02-11
Fix the retry logic of our api client.

pault84:
Currently it's possible for the Do() call to receive a 503 status
without a retry-after header.

This results in calls looping every second for 5 minutes.

Fixed this with a check for the header and a boolean check.

### DIFF
--- a/api/client/request.go
+++ b/api/client/request.go
@@ -284,7 +284,11 @@ func (r *Request) Do() *Response {
 			break
 		}
 		attemptNum++
-		handleServiceUnavailable(resp, attemptNum)
+
+		// If we don't receive a retry header we should exit.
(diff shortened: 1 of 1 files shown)


api/client/request.go
```diff
@@ -302,17 +306,25 @@ func (r *Request) Do() *Response {
 	}
 }
 
-func handleServiceUnavailable(resp *http.Response, attemptNum int) {
+func handleServiceUnavailable(resp *http.Response, attemptNum int) bool {
```
lpabon: If it is returning a `bool` then the name of the function should change to a question, something like `retryHeaderReceived()` that way the `if` statement is easier to read.

pault84: sure will address that.

PR #1437, opened 2020-02-20
Add the prometheus GRPC interceptor

pault84:

It adds the prometheus GRPC interceptor.
It exposes our GRPC metrics on port 9021 and /metrics

We can use this to expose dashboards on our performance / error codes etcetera.


api/server/sdk/rest_gateway.go
```diff
@@ -118,6 +119,8 @@ func (s *sdkRestGateway) restServerSetupHandlers() (http.Handler, error) {
 	mux.Handle(prefix,
 		http.StripPrefix(prefix, http.FileServer(swaggerUIBox)))
 
+	mux.Handle("/metrics", promhttp.Handler())
```
ggriffiths: Maybe change to `/prometheus/metrics` to be clear? Also, is it ok to expose this endpoint without auth? I saw prometheus has an option to hit a metrics endpoint with basic auth, but that might be a pain for configuration.

ggriffiths: i.e. to distinguish between any other metrics we provide in the future

pault84: prometheus defaults to /metrics if we move these to another endpoint would require extra configuration

PR #1446, opened 2020-02-29
Add VolumeCatalog call to the SDK

pault84:

This adds SDK support for the Catalog Feature.
Basically just link this to the protodriver implementation.


api/api.proto
```diff
@@ -1793,6 +1793,14 @@ service OpenStorageVolume {
         body: "*"
       };
     }
+
+    rpc VolumeCatalog(SdkVolumeCatalogRequest)
```
lpabon: This needs a comment as if it was explained to a non-portworx developer. Maybe add an example also.

api/api.proto
```diff
@@ -4518,3 +4526,15 @@ message RestoreVolumeSpec {
   // fastpath extensions
   RestoreParamBoolType fp_preference = 21;
 }
+
+// Request message to get the volume catalog
+message SdkVolumeCatalogRequest {
+  string volume_id =1;
```
lpabon: These three values need comments.

PR #1464, opened 2020-03-16
Credential option to use IAM creds for s3

talakad:

Allows users to specify if IAM credential should be used for aws s3 client
Closes #

Missed updating sdk part of creds. Requesting review again.


api/api.go
```diff
@@ -129,6 +129,8 @@ const (
 	OptCredOwnership = "CredOwnership"
 	// OptCredProxy proxy key in params
 	OptCredProxy = "CredProxy"
+	// OptCredIAMPolicy if set, indicates IAM credentials for s3
```
disrani-px: It would be to use IAM for any of the 3 providers

talakad: Done.

PR #1472, opened 2020-03-23
Update NotInQuorum self node state to listeners.

adityadani:
- Whenever a node's status changes to NotInQuorum, Update all the listeners.

Closes #

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -625,7 +625,7 @@ func (c *ClusterManager) initNodeInCluster(
 			if self.Status != api.Status_STATUS_MAINTENANCE {
 				self.Status = api.Status_STATUS_ERROR
 			}
-			logrus.Warnf("Failed to initialize Init %s: %v",
+			logrus.Warnf("Failed to initialize %s: %v",
 				e.Value.(cluster.ClusterListener).String(), err)
 			c.cleanupInit(clusterInfo, self)
 			return nil, err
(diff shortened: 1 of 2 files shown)


cluster/manager/manager.go
```diff
@@ -1186,7 +1232,7 @@ func (c *ClusterManager) initListeners(
 		// already in cluster, check with listeners if it is OK to join
 		// this cluster.
 		for e := c.listeners.Front(); e != nil; e = e.Next() {
-			err := e.Value.(cluster.ClusterListener).CanNodeJoin(&c.selfNode, clusterInfo, nodeInitialized)
+			err := e.Value.(cluster.ClusterListener).CanNodeJoin(&c.selfNode, db, nodeInitialized)
```
adityadani: Send the latest ClusterInfo object (db) to listeners instead of an older clusterInfo object

cluster/manager/manager.go
```diff
@@ -622,13 +654,19 @@ func (c *ClusterManager) initNodeInCluster(
 	for e := c.listeners.Front(); e != nil; e = e.Next() {
 		finalizeCb, err := e.Value.(cluster.ClusterListener).Init(self, clusterInfo)
 		if err != nil {
-			if self.Status != api.Status_STATUS_MAINTENANCE {
-				self.Status = api.Status_STATUS_ERROR
+			initErr := err
```
sangleganesh: this does not look correct - if init fails for one listener, with no quorum member error, and we wait and quorum is established, we would continue initing other listeners - so some listeners init failed and for others it was successful ? 
we can simply return the error and cleanup. let listener do the waiting.

adityadani: Agreed. The assumption here was init has succeeded but the listener asked to wait. 

The problem with doing a cleanup is we will be cleaning up the Init done by this listener as well. Also currently if an error is returned from this function then  for the next retry the cluster manager has to go through its whole state machine again if it wants to proceed.

Instead can we change the listener's Init API itself to look something like this
func Init(api.Node, api.ClusterInfo) (waitForQuorum bool, waitForQuorumMessage string, error)

In this way the API itself is clear and the listener can return a nil error but still ask the cluster manager to wait for quorum.

PR #1482, opened 2020-03-30
do not store kvdb instance

sangleganesh:
when kvdb becomes unavailable, a new wrapper is added to it.
query kvdb instance instead of saving it.

Signed-off-by: ganesh <ganesh@portworx.com>

Closes #


adityadani:
> LGTM.
> I'm assuming since we're not multie cluster the clusterid was useless also?

This was probably done for older version of lighthouse that could get alerts from multiple clusters?

PR #1485, opened 2020-03-31
PWX-11466

stgleb:
Unsecure nodehealth handler for oci monitor


zoxpx:
Perhaps I missed something -- the cf14701 just removes `/nodehealth` URI, and does not replace it with anything.  This immediately breaks the Readiness probe we have to report to Kubernetes.

stgleb:
@zoxpx Nope, there is a code https://github.com/libopenstorage/openstorage/blob/master/api/server/cluster.go#L46 that grabs all secured and unsecured handlers. First it applies secured to the router and than applies those unsecured handlers that were not applied on previous step. So that, by deleting nodehealth from the list of secured handlers I simple make it unsecured, but stil existing.

PR #1494, opened 2020-04-07
Implementation specific volumes attributes support

prabirpaul:

This adds support for specifying extended attributes for volume. An extended attribute is something which is specific to the underlying datastore and may or may not be supported by all.

Closes #


api/api.proto
```diff
@@ -302,6 +302,13 @@ message IoStrategy {
   bool early_ack = 2;
 }
 
+// Xattr defines implementation specific volume attribute
+enum Xattr {
+	XATTR_NONE = 0;
```
lpabon, line 307: A couple of things (i need to add this to the sdk.md if not already there).

1. Enums need to be CamelCase.
2. Enums need to be in a message so that you do not need to prefix them like C. Like this:

```
// Doc needed
message VolumeXattr {
   // Do not use
   Unknown = 0;
   // Doc needed
   None = 1;
   // Doc needed
   CowOnDemand = 2;
}
```

Also docs needed for each so that it automatically comes on to the doc site on https://libopenstorage.github.io/. The docs need to be very clear. Try to add more than just a simple line for a Portworx dev. Try to add enough information for those who are not Portworx developers.

PR #1535, opened 2020-04-27
Add trace level logs for NFS mount load

ggriffiths:

Closes #


pkg/mount/nfs.go
```diff
@@ -88,29 +93,34 @@ func (m *nfsMounter) normalizeSource(info *mount.Info, host string) {
...
+	logrus.Trace("Entered Load()")
 	info, err := mount.GetMounts()
 	if err != nil {
 		return err
 	}
 	re := regexp.MustCompile(`,addr=(.*)`)
 MountLoop:
-	for _, v := range info {
+	for i, v := range info {
+		logrus.Trace("Info[%d] = %v", i, *v)
```
harsh-px, line 104: does this need to be `Tracef` instead? Same at other places

ggriffiths: Ah yeah, will add as Tracef

PR #1543, opened 2020-04-29
fix type of replication set ID in rebalance audit

harsh-px:

Closes #


api/api.proto
```diff
@@ -3176,14 +3176,16 @@ message StorageRebalanceJob {
 
 // StorageRebalanceJobState is an enum for state of the current rebalance operation
 enum StorageRebalanceJobState {
+    // Pending indicates job is still pending and has not started work
+  PENDING = 0;
```
harsh-px, line 3180: changing the enum numbers right now as we don't have to worry about upgrages

PR #1544, opened 2020-04-30
TLS configuration for Px REST MGMT and SDK/iSDK endpoints

taaraora:
To have functions which can produce dial options and endpoints.
Closes #

### DIFF
--- a/volume/drivers/pwx/connection.go
+++ b/volume/drivers/pwx/connection.go
@@ -1,0 +1,178 @@
+package pwx
+
+import (
+	"fmt"
+	"github.com/libopenstorage/openstorage/pkg/grpcserver"
+	"github.com/portworx/sched-ops/k8s/core"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"os"
+	"strconv"
+	"strings"
+)
+
+// ConnectionParamsBuilder contains dependencies needed for building Dial options and endpoints
+// to SDK/iSDK and legacy REST API connections
+type ConnectionParamsBuilder struct {
+	Config  *ConnectionParamsBuilderConfig
+	kubeOps core.Ops
+}
+
+// ConnectionParamsBuilderConfig contains default values for all PX legacy rest API and SDk API connections
+// also can be used for redefining names of Env variables which are used for external configuration
+type ConnectionParamsBuilderConfig struct {
+	// DefaultServiceName is the default name of the Portworx service
+	DefaultServiceName string
+	// DefaultServiceNamespaceName is the kubernetes namespace in which Portworx daemon set runs
+	DefaultServiceNamespaceName string
+	// DefaultRestPortName is name of Porx legacy REST API port in service
+	DefaultRestPortName string
+	// DefaultSDKPortName is name of Porx SDK/iSDK API port in service
+	DefaultSDKPortName string
+
+	// Environment variables names to get  values config properties
+	EnableTLSEnv       string
+	NamespaceNameEnv   string
+	ServiceNameEnv     string
+	CaCertSecretEnv    string
+	CaCertSecretKeyEnv string
+}
+
+// NewConnectionParamsBuilderDefaultConfig returns ConnectionParamsBuilderConfig with default values set
+func NewConnectionParamsBuilderDefaultConfig() *ConnectionParamsBuilderConfig {
+	return &ConnectionParamsBuilderConfig{
+		DefaultServiceName:          "portworx-service",
+		DefaultServiceNamespaceName: "kube-system",
+		DefaultRestPortName:         "px-api",
+		DefaultSDKPortName:          "px-sdk",
+		EnableTLSEnv:                "PX_ENABLE_TLS",
+		NamespaceNameEnv:            "PX_NAMESPACE",
+		ServiceNameEnv:              "PX_SERVICE_NAME",
+		CaCertSecretEnv:             "PX_CA_CERT_SECRET",
+		CaCertSecretKeyEnv:          "PX_CA_CERT_SECRET_KEY",
+	}
+}
+
+// NewConnectionParamsBuilder constructor function to create ConnectionParamsBuilder with needed dependencies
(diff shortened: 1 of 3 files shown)


volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,210 @@
...
+	DefaultSDKPortName:          "px-sdk",
+	EnableTLSEnv:                "PX_ENABLE_TLS",
+	NamespaceNameEnv:            "PX_NAMESPACE",
+	ServiceNameEnv:              "PX_SERVICE_NAME",
+	CaCertSecretEnv:             "PX_CA_CERT_SECRET",
+	CaCertSecretKeyEnv:          "PX_CA_CERT_SECRET_KEY",
+}
+
+func NewConnectionParamsBuilder(ops core.Ops, params *ConnectionParamsBuilderConfig) *ConnectionParamsBuilder {
+	if params == nil {
```
lpabon, line 50: nit: This works, but when reading the code that uses this, the reader does not know what `nil` will be interpreted as. Instead, there could be a `NewConnectionParamsBuilderConfig()` and a `NewConnectionParamsBuilderDefaultConfig()`. When the caller calls `NewConnectionParamsBuilderDefaultConfig()` they can use that to pass into this function. 

That way you do not have to silently set the values if it is not passed, and instead return an error.

volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+// BuildClientsEndpoints returns two endpoints for PX MGMT API and gRPC SDK/iSDL API
+func (cpb *ConnectionParamsBuilder) BuildClientsEndpoints() (pxMgmtEndpoint string, sdkEndpoint string, err error) {
+	var endpoint string
+
+	// Check if service name and namespace is provided
+	// as environment variables
+	serviceName := getPxService(cpb.Config.ServiceNameEnv, cpb.Config.DefaultServiceName)
```
lpabon, line 102: nit: These can be part of the object. So that way it would look like this: `cpb.getPxService()` and it always returns the same data for the same object. Same with getPxNamespace()

volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,210 @@
...
+	ns := getPxNamespace(cpb.Config.NamespaceNameEnv, cpb.Config.DefaultServiceNamespaceName)
+
+	svc, err := cpb.kubeOps.GetService(serviceName, ns)
+	if err == nil {
+		endpoint = svc.Spec.ClusterIP
+	} else {
+		return "", "", fmt.Errorf("failed to get k8s service spec: %v", err)
```
lpabon, line 109: nit: may want to say the full information. Instead of `spec` use `specification`, since this will be read by the admin in the logs

volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	if len(endpoint) == 0 {
```
lpabon, line 112: This probably should with line 107. Since that is when it is assigned. Or it can first check `len(svc.Spec.ClusterIP)`

volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,210 @@
...
+			svcPort.Port != 0 {
+			restPort = int(svcPort.Port)
+		}
+	}
+
+	// check if the ports were parsed
+	if sdkPort == 0 || restPort == 0 {
+		err := fmt.Errorf("%s in %s namespace has been not updated to the latest spec. "+
+			"%s and/or %s ports are missing. Follow "+
+			"https://docs.portworx.com/portworx-install-with-kubernetes/operate-and-maintain-on-kubernetes/upgrade/"+
```
lpabon, line 134: We should not be adding links to the docs, since the docs can move after this code is released. The link should be removed.

volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,210 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to load CA cert secret: %v", err)
+		}
+
+		exist := false
+		rootCA, exist = secret.Data[caCertSecretKey]
+		if !exist {
+			return nil, fmt.Errorf("failed to load CA cert from secret: %s using key: %s", caCertSecretName, caCertSecretKey)
+		}
+		if len(rootCA) == 0 {
```
lpabon, line 176: Does this mean if the caller calls this function there _must_ be a rootCA? What if the CA is already part of the container/host?

volume/drivers/pwx/connection.go
```diff
@@ -0,0 +1,142 @@
...
+	defaultServiceName = "portworx-service"
+
+	// defaultNamespace is the kubernetes namespace in which portworx daemon set runs
+	defaultNamespace = "kube-system"
+
+	pxRestPort = "px-api"
+	pxSdkPort  = "px-sdk"
+)
+
+func NewClientEndPointBuilder(pxServiceNameEnv, pxNamespaceNameEnv string) func(kubeOps core.Ops) (pxMgmtEndpoint string, sdkEndpoint string, err error) {
```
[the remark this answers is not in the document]
taaraora: Got it. I've added comments to all exportable entities.

volume/drivers/pwx/connection.go
taaraora: That is a good idea.

volume/drivers/pwx/connection.go
taaraora: Thank you, fixed.

volume/drivers/pwx/connection.go
taaraora: That is a good point. Checks were reorganised.

volume/drivers/pwx/connection.go
taaraora: Got it. That error was copied from Porx repo. 
Fixed.

volume/drivers/pwx/connection.go
taaraora: I add that check intentionally. My reasoning is: if a user has configured service to read CA cert from a specific secret/secret key we should expect a certificate.
If CA cert is part of container already a user should just enable TLS and do not configure an application to read CA cert from k8s secret.

PR #1547, opened 2020-05-01
PWX-12646 Fix error message for empty token error

stgleb:


api/server/middleware_auth.go
```diff
@@ -64,26 +64,26 @@ func SecurityHandler(authenticators map[string]auth.Authenticator, next http.Han
...
-
 		tokenHeader := r.Header.Get("Authorization")
 		tokens := strings.Split(tokenHeader, " ")
 
 		if len(tokens) < 2 {
 			w.WriteHeader(http.StatusUnauthorized)
-			json.NewEncoder(w).Encode(&api.ClusterResponse{
-				Error: fmt.Sprintf("Access denied, token is malformed"),
-			})
+			fmt.Fprintf(w, "Access denied token is empty")
```
lpabon, line 72: I think the logic should be to check for token len of 0 then check for token len of less than 2. The way the logic is now, it is hard to determine if it is empty or malformed.

stgleb: Yep, I agree with you. If len of tokens slice is 2 that means that token itself can not be empty.

PR #1568, opened 2020-05-12
envs to set porx endpoint and ports with no dynamic configuration

taaraora:
Signed-off-by: Dmytro Tsapko <dmytro.tsapko@gmail.com>

Updates to let a user manually set Portworx endpoint, SDK and rest APIs ports
Closes #

### DIFF
--- a/volume/drivers/pwx/connection.go
+++ b/volume/drivers/pwx/connection.go
@@ -31,11 +31,22 @@ type ConnectionParamsBuilderConfig struct {
 	DefaultSDKPortName string
 
 	// Environment variables names to get  values config properties
-	EnableTLSEnv       string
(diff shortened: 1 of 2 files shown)


volume/drivers/pwx/connection.go
```diff
@@ -154,6 +176,38 @@ func (cpb *ConnectionParamsBuilder) BuildDialOps() ([]grpc.DialOption, error) {
 	return dialOptions, nil
 }
 
+func (cpb *ConnectionParamsBuilder) checkStaticEndpoints() (string, string, error) {
```
lpabon, line 179: Everything looks good. Could you write a unit test just for this function.

PR #1578, opened 2020-05-22
Add a new volume xattr for direct-io support

prabirpaul:

Closes #


api/server/sdk/volume_ops.go
```diff
@@ -811,6 +811,12 @@ func (s *VolumeServer) mergeVolumeSpecs(vol *api.VolumeSpec, req *api.VolumeSpec
 		spec.ExportSpec = vol.GetExportSpec()
 	}
 
+	// Xattr
+	if req.GetXattr() >= 0 {
```
lpabon, line 815: here, you want to use req.GetXattrOpt() referring to the in line 514 of api.proto. This will be nil if not provided. See the example above on line 801 on this file

api/spec/spec_handler.go
```diff
@@ -398,6 +399,14 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
 					spec.Xattr = api.Xattr_COW_ON_DEMAND
 				}
 			}
+		case api.SpecDirectIo:
+			if directIo, err := strconv.ParseBool(v); err != nil {
+				return nil, nil, nil, err
+			} else {
+				if directIo {
+					spec.Xattr |= api.Xattr_DIRECT_IO
```
lpabon, line 407: Are xattr flags? It is not commented in that manner. You may want to use hex value and comment the volumespec accordingly. Also, it is too bad it is already committed, but it would be good to instead have xattr be something like `xttar_options` where it was an array of options instead of C-style flags. Makes it easier to do in  gRPC from any language to use an array of options instead of flags.

PR #1583, opened 2020-05-28
Only log trace cache when source is not in nfs mount table

ggriffiths:

Closes #

### DIFF
--- a/pkg/mount/nfs.go
+++ b/pkg/mount/nfs.go
@@ -42,7 +42,6 @@ func NewNFSMounter(servers []string,
 	}
 	err := m.Load([]string{""})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/mount/nfs.go
```diff
@@ -67,6 +67,11 @@ func (m *nfsMounter) Reload(source string) error {
 
 	err = m.reload(source, newNFSmounter.mounts[source])
 	m.LogDevices()
+
+	if m.mounts[source] == nil {
```
adityadani, line 71: This will end up logging every time Reload() is called when source is not mounted which might be a genuine case.
Instead before "reload()" is called check if m.mounts[source] exists and if it exists and now it is nil then log them.

PR #1592, opened 2020-06-03
Add fix for consistent mount table read

ggriffiths:

Closes #

### DIFF
--- a/pkg/mount/bind_mount.go
+++ b/pkg/mount/bind_mount.go
@@ -3,6 +3,7 @@
 package mount
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/docker/docker/pkg/mount"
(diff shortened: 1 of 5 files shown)


ggriffiths:
Travis hook was not synced up. Build is passing:
https://travis-ci.org/github/libopenstorage/openstorage/builds/694418153?utm_source=github_status&utm_medium=notification

kolyshkin:
I am aware of this problem (see https://github.com/kolyshkin/procfs-test). The kernel serves mountinfo contents line by line, iterating through the mounts. If the next mount to be printed is getting unmounted, we have this bug. So, it happens when there are lots of mounts/unmounts happening while we read the file.

The fix used here does not look to be correct. Reading mountinfo twice in these circumstances most probably will produce different results, as mounts are added and removed. This also creates a heavy load on the system, which is already quite loaded with doing all these mounts/unmounts.

From what I see, it looks like you are looking for a particular mount. In this case, I guess, the best option to mitigate the issue is to re-read mountinfo again _only_ when you haven't found the mount you're looking for.

While at it, may I suggest you to use `github.com/moby/sys/mountinfo` package rather that forking some old, slow, and not entirely correct code from Docker?

kolyshkin:
@ggriffiths PTAL ^^^

adityadani:
@kolyshkin Thank you for providing your insight on this issue. We are aware of the procfs behavior which is why we added this workaround. Our workaround is based of what is done in kubernetes, which also tries to read the mount table before proceeding with Mount and Unmount operations - https://github.com/kubernetes/utils/blob/master/io/read.go#L33-L35
Having said that I am open to other cleaner and optimized approaches. 

We will definitely look into moving to `github.com/moby/sys/mountinfo`, but as of now we are only using the MountInfo structure from the docker package to parse the mountinfo file. I did not completely follow why would it be slow with using the old package just for referencing a golang object. May be I am missing something here?

PR #1596, opened 2020-06-05
Rename system.public to system.guest

ggriffiths:

Renames from system.public to system.guest, and corresponding logic renaming.

Closes #


pkg/auth/auth.go
```diff
@@ -20,6 +20,10 @@ import (
 	"context"
 )
 
+const (
```
ggriffiths, line 23: keeping this as an unexported constant, as importing the role pkg creates a cyclic dependency.

pkg/auth/userinfo.go
```diff
@@ -60,14 +57,14 @@ func NewUserInfoFromContext(ctx context.Context) (*UserInfo, bool) {
 func NewPublicUser() *UserInfo {
```
lpabon, line 57: Do you want to change this API name also?

ggriffiths: Yep, nice catch. Fixed this

PR #1607, opened 2020-06-12
osd changes for fsck end-to-end support

vjs-px:
This change is required to provide additional API to support fsck end-to-end use cases mentioned in this defect
https://portworx.atlassian.net/browse/PWX-13557

Closes #


api/api.proto
```diff
@@ -710,8 +710,10 @@ message Volume {
   google.protobuf.Timestamp attach_time = 24;
   // DetachTime time this device was detached.
   google.protobuf.Timestamp detach_time= 25;
+  // LastFsckScanFix is the time when an integrity check fixed errors in filesystem
+  google.protobuf.Timestamp last_fsck_scan_fix = 26;
```
talakad, line 714: you may want to move this to 27,  and retain fpConfig at 26 to retain backward compatibility with clients

vjs-px: moved the new field to 27

PR #1622, opened 2020-06-16
fsck changes based on feekback from demo

vjs-px:
This PR has changes to fsck service based on feedback from demo
1. Remove FS_HEALTH_STATUS_NOT_HEALTHY and replace with less FS_HEALTH_STATUS_SAFE_TO_FIX and FS_HEALTH_STATUS_NEEDS_INSPECTION

2.Instead of using volume labels to turn on/off fsck on a volume use a specific field in volume spec. See ScanPolicy structure in api.proto

3. Changes to API naming from SdkFilesystemCheckGetStatus to SdkFilesystemCheckStatus

4. Changes to API naming from SdkFilesystemTrimGetStatus to SdkFilesystemTrimStatus

5. Add new field to volume struct - LastScanStatus

6. Rename fields LastFsckScan and LastFsckScanFix to LastScan and LastScanFix

Closes #
https://portworx.atlassian.net/browse/PWX-13557


api/api.proto
```diff
@@ -391,6 +403,31 @@ message FastpathConfig {
...
+    }
+
+    ScanTrigger Trigger = 1;
+    ScanAction Action = 2;
+}
+
+message ScanPolicies {
+  // ScanPolicy specifies the filesystem check policy to apply for this volume
+  // There can be more than one ScanPolicy associated with a volume
+  repeated ScanPolicy scan_policy = 37;
```
talakad, line 428: this should be 1? i mean repeated ScanPolicy scan_policy = 1; or did you want to reserve some fields?

vjs-px: Had a discussion with Veda and Harsh. We will only have one ScanPolicy per volume. ScanPolicies structure would not be required, hence removing it

PR #1630, opened 2020-06-23
rebalance data structure changes

sangleganesh:
- explicity add source and target pool structures
- rename error to status

Closes #


api/api.proto
```diff
@@ -3155,11 +3155,13 @@ message SdkStorageRebalanceRequest {
   repeated  StorageRebalanceTriggerThreshold trigger_thresholds = 1;
   // TrialRun if true the job only produces steps that would be taken without making any changes
   bool trial_run = 2;
-  // PoolSelectorLabels allows selecting pools to which trigger thresholds will apply
-  map<string, string> pool_selector_labels = 3;
+  // SourcePoolSelectorLabels allows selecting pools to which trigger thresholds will apply as source
+  map<string, string> source_pool_selector_labels = 3;
+  // TargetPoolSelectorLabels allows selecting pools to which trigger thresholds will apply as target
+  map<string, string> target_pool_selector_labels = 4;
```
harsh-px, line 3161: @lpabon heads up.

We are changing fields in the proto without updating SDK version as this is the rebalance API and this has not been released yet. These changes incorporate feedback during 2.6.0 testing.

api/api.proto
```diff
@@ -3155,11 +3155,13 @@ message SdkStorageRebalanceRequest {
   repeated  StorageRebalanceTriggerThreshold trigger_thresholds = 1;
   // TrialRun if true the job only produces steps that would be taken without making any changes
   bool trial_run = 2;
-  // PoolSelectorLabels allows selecting pools to which trigger thresholds will apply
-  map<string, string> pool_selector_labels = 3;
+  // SourcePoolSelectorLabels allows selecting pools to which trigger thresholds will apply as source
+  map<string, string> source_pool_selector_labels = 3;
```
harsh-px, line 3159: Instead of map<string, string> let's use the existing `LabelSelectorRequirement` type already in this proto that we use for volume placement strategies.

That will enable the API to support more primitives like NotEquals, In, Gt .. ? With a map, we are stuck with a strict equals label match.

sangleganesh: updated, though cli support is not available, we can special case.

api/api.proto
```diff
@@ -3156,9 +3156,9 @@ message SdkStorageRebalanceRequest {
   // TrialRun if true the job only produces steps that would be taken without making any changes
   bool trial_run = 2;
   // SourcePoolSelectorLabels allows selecting pools to which trigger thresholds will apply as source
-  map<string, string> source_pool_selector_labels = 3;
+  LabelSelectorRequirement source_pool_selector = 3;
```
harsh-px, line 3159: nit: Update comment to `SourcePoolSelector` and `TargetPoolSelector`

SDK_CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Releases
 
+### v0.84.0 - (6/24/2020)
+
+* Updated rebalance data structures
```
harsh-px, line 7: ah, is this needed to make CI pass?

I was hoping to skip that since the rebalance structures were never released so we didn't need to bump the version.

sangleganesh: unfortunately it won't allow me to merge. It is not clear why we need to update this text file - we can just generate it always from git commits.

PR #1665, opened 2020-07-02
Export default role RBAC rules

ggriffiths:

To be used with operator - so we can import the system.guest default rules.

Closes #


pkg/role/sdkserviceapi.go
```diff
@@ -47,8 +47,9 @@ type defaultRole struct {
 }
 
 var (
-	// Default roles. Should be prefixed by `system.` to avoid collisions
-	defaultRoles = map[string]*defaultRole{
```
lpabon, line 51: You may need access to `defaultRole` struct, right?

PR #1667, opened 2020-07-06
Add MountOptions to Volume

adityadani:
Add MountOptions to Volume objects.
- Add MountOptions to Volume, VolumeSpec, VolumeSpecUpdate and VolumeRestoreSpec.
- Make changes to the spec handler to handle this new option.

Handle MountOptions in SDK volume update API.

Closes #


SDK_CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Releases
 
+### v0.88.0 - (7/06/2020)
+
+* Add a new mount_options field to Volume and VolumeSpec object.
```
jvinod, line 7: ~Add a new mount_options field~ Add mount_options field

api/api.proto
```diff
@@ -348,6 +348,16 @@ message ExportSpec {
    string export_options = 2;
 }
 
+// MountOptionsSpec defines the mount options with which a volume is mounted.
+message MountOptionsSpec {
```
jvinod, line 352: Should we just call it MountOptions

api/spec/spec_handler.go
```diff
@@ -576,6 +588,12 @@ func (d *specHandler) SpecOptsFromString(
 	if ok, exportOptions := d.getVal(exportOptionsRegex, str); ok {
 		opts[api.SpecExportOptions] = exportOptions
 	}
+	if ok, mountOptions := d.getVal(mountOptionsRegex, str); ok {
+		// mount options will be provided as a string in the following format
+		// mount_options=k1;k2:v2;k3:v3
+		// convert it to k1,k2=v2,k3=v3
```
jvinod, line 594: Why not have the input as how you want it. Is there a problem with the ','

adityadani: Yes the `,` is the issue. This is used by the docker driver and w/o using different symbols, each mount option would end up as a new volume argument.

api/api.proto
adityadani: done

PR #1700, opened 2020-08-03
Remove IO_PROFILE_BKUPSRC from supported list

talakad:
Moved IO_PROFILE_BKUPSRC to its own struct as its applicable to
RestoreVolSpec only.

Closes #


api/api.proto
```diff
@@ -4780,6 +4778,12 @@ message RestoreVolSnashotSchedule {
 message RestoreVolStoragePolicy {
 	string policy = 1;
 }
+message RestoreVolIoProfile {
+	oneof io_profile {
+		bool io_profile_bkup_src = 1;
+        IoProfile restore_io_profile = 2;
```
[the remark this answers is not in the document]
talakad: IO_PROFILE_BKUPSRC was applicable only to RestoreVolumeSpec and not for regular volume. Hence removed it. The bool is needed if someone wishes to simply inherit the Ioprofile from the cloudbackup without specifying it.

api/api.proto
```diff
@@ -112,8 +112,6 @@ enum IoProfile {
   IO_PROFILE_CMS = 4;
   IO_PROFILE_SYNC_SHARED = 5;
   IO_PROFILE_AUTO = 6;
-  // IO_PROFILE_BKUPSRC inherits cloudbackup's IOprofile
```
sangleganesh, line 115: protobuf recommends marking as obsolete instead of removing fields/constants.

talakad: Since this is introduced recently, taking it out completely rather than marking obsolete.

PR #1725, opened 2020-08-12
Added NodeDrain SDK API.

adityadani:
**What this PR does / why we need it**

- The NodeDrain API provides a way to drain a node from storage perspective. Since
  storage node drain involves a lot of sub operations the NodeDrainRequest takes
  different options to tune the overall drain operation.
- The API also provides a way to perform certain specific sub operations of a drain request.

The SDKNodeDrainRequest comprises of different NodeDrainOptions:
- SDKNodeDrainAttachmentOptions:
  - These define how volume attachments need to be removed from a node.
- SDKNodeDrainVolumeReplicaOptions:
  -  These define how volume replicas can be moved out of a node.
- SDKNodeDrainKvdbOptions
  - Currently even though this options is accepted as a protobuf message it acts as a
    boolean if kvdb needs to be moved out of this node.
- SDKNodeDrainCloudDriveOptions
  - These define options for draining/detaching cloud drives from a node.
- There exists a global switch to perform all the above node drain sub operations. But by default
  each option needs to be set in order for it to be performed by the SDK service.

NodeDrainSelector:
- The selector is used for selecting specific resources for draining from a node.
  For ex. a selector can be used for draining volume replicas for a subset of volumes
  which satisfy a given set of labels.

Asynchronous Model:
The NodeDrain API is asynchronous. The Request API starts a new job for NodeDrain with the provided options.
The Status API can be used to query the status of a node drain job. It will return an in-depth summary for
each of the sub operation.


api/api.proto
```diff
@@ -3229,6 +3238,73 @@ message SdkNodeInspectRequest {
   string node_id = 1;
 }
 
+// Options for draining volume attachment from a node
+message SdkNodeDrainAttachmentOptions {
+    // Selector is used for selecting volumes whose
+    // attachment needs to be moved from this node.
+    // The selector could be a list of volume label
+    // key value pairs to select a subset of volumes.
+    map<string, string> selector = 1;
```
harsh-px, line 3247: for all selectors, use `LabelSelectorRequirement` (see examples in other places in the proto file).

Recently, we have been using this for all new selector use cases as it's more extensible.

Same for all selectors in the PR

api/api.proto
```diff
@@ -3229,6 +3238,73 @@ message SdkNodeInspectRequest {
...
+    // replicas need to be moved from this node to another.
+    // The selector could be a list of volume label
+    // key value pairs to select a subset of volumes.
+    map<string, string> selector = 1;
+}
+
+// Options for draining/detaching cloud drives from a node
+message SdkNodeDrainCloudDriveOptions {
+    // When set the cloud drives will also get deleted.
+    bool delete_drives = 1;
```
harsh-px, line 3265: the delete drive part should be part of the node decommission API. So we can keep it out of here.

api/api.proto
```diff
@@ -3229,6 +3238,73 @@ message SdkNodeInspectRequest {
...
+message SdkNodeDrainKvdbOptions {
+}
+
+
+// Options for draining a node in the cluster
+message SdkNodeDrainOptions {
+    // Perform all the node drain related operations.
+    // This is a single switch to perform all the supported
+    // node drain sub operations.
+    bool all = 1;
```
harsh-px, line 3278: if we want to use this as the default, we can remove it.

Because since it's a bool, if not supplied, it wills all=false, but our default implementation will perform all sub operations.

api/api.proto
```diff
@@ -3229,6 +3238,73 @@ message SdkNodeInspectRequest {
...
+    // Selector is a global selector used while performing
+    // a node drain sub operation. Per option selector take
+    // precedence over global selector
+    map<string, string> selector = 2;
+    // When set volume attachments will be drained from this node.
+    SdkNodeDrainAttachmentOptions drain_attachments = 3;
```
harsh-px, line 3284: to actually drain the attachments, we will have to move the application pods from the node as for the non-shared use case, that's the primary way volumes will be attached on a node.

So should we call this `SdkNodeDrainAppOptions` instead?

For sharedv4, we will drain app pods AND also move the volume attachment.

api/api.proto
```diff
@@ -3229,6 +3238,73 @@ message SdkNodeInspectRequest {
...
+    // When set volume replicas from this node will be moved to
+    // another node
+    SdkNodeDrainVolumeReplicaOptions drain_volume_replicas =  4;
```
harsh-px, line 3287: Not sure if we want to do this be default.

If we consider the PKS case, during drain, we will detach the cloud drives itself. So doing repl moves by default might be an overkill since it's upgrades one node at a time.

api/api.proto
adityadani: I was thinking by default we will **not** do all the drain sub operations, but instead only do the operations which are set in the NodeDrainRequest. So this boolean is a single watch for callers of this API to perform all drain sub-operations with default settings.

api/api.proto
adityadani: done

api/api.proto
adityadani: makes sense. done

api/api.proto
adityadani: done

api/api.proto
adityadani: A node drain operation could also involve just draining the replicas of a volume or just a subset of volumes from a node. Thats the reason why if an option is not set the Storage implementation will not perform that specific node drain operation. That also allows performing only certain sub drain operations.

sangleganesh:
- this api should not be synchronous, instead it should follow the rebalance api model for request and status check.
- why does it need to be prefixed with SDK ?

api/api.proto
adityadani: How we drain the attachments is an implementation detail. Shouldn't the API reflect what it should do and in this case move away all the volume attachments from this node?

api/api.proto
adityadani: I have removed the boolean flag "all". The API is now more explicit, and the caller has to set what all sub operations need to be done as a part of this node drain request. The callers of this API whether its CLI or Operator should provide wrappers on performing sub operations.

adityadani:
I have updated the PR to use the asynchronous model similar to the Rebalance service. Also handled other comments.

api/api.proto
```diff
@@ -1786,6 +1786,45 @@ service OpenStorageNode {
...
+  // NodeDrainJobState_CANCELLED - acceptable only from running/pause state
+  // NodeDrainJobState_RUNNING - acceptable only from pause state
+  rpc UpdateNodeDrainJobState(SdkUpdateNodeDrainJobRequest)
+  returns (SdkUpdateNodeDrainJobResponse) {
+    option(google.api.http) = {
+        put: "/v1/nodes/drain/job/{id}"
+      };
+  }
+
+  // GetDrainStatus drains the provided node in the cluster. The options provided
```
piyush-nimbalkar, line 1811: nit: Update the comment

api/api.proto
```diff
@@ -3229,6 +3268,174 @@ message SdkNodeInspectRequest {
...
+}
+
+// Provides a summary of  a node drain job
+message NodeDrainSummary {
+    // Attachment summary
+    NodeDrainVolumeSummary attachment_summary = 1;
+    // Volume replica summary
+    NodeDrainVolumeSummary volume_replica_summary = 2;
+    // Kvdb drain summary
+    NodeDrainSubOperationState kvdb_summary = 3;
```
piyush-nimbalkar, line 3434: Should we have a summary object for kvdb and cloud drive as well for future? Initially we can only have a state in it similar to NodeDrainVolumeSummary.

api/api.proto
adityadani: done

api/api.proto
adityadani: makes sense. I have made those changes.

PR #1728, opened 2020-08-18
Add CredentialRemoveRefrences API

talakad:

Allows users to delete any pending references to a credential 
Closes #


api/server/sdk/credentials_test.go
```diff
@@ -1099,3 +1099,80 @@ func TestSdkCredentialGetOwnershipFromCred(t *testing.T) {
...
+	}
+
+	s.MockDriver().
+		EXPECT().
+		CredsEnumerate().
+		Return(enumerateData, nil)
+	s.MockDriver().
+		EXPECT().
+		CredsRemoveReferences(uuid).
+		Return(fmt.Errorf("Failed to remove refs"))
```
vjs-px, line 1167: I did not clearly get the difference between the TestSdkCredentialRemovereferencesSuccess() and TestSdkCredentialRemoveReferencesFailed()

Only the uuid is different and the mock driver return code is different. If the mock driver is not going to consult the enumerateData, then why even populate it?

talakad: The goal is to test error response for the bad case.  We dont need to populate the data as such. but does not hurt.

api/server/sdk/credentials.go
```diff
@@ -231,6 +231,41 @@ func (s *CredentialServer) Validate(
...
+func (s *CredentialServer) RemoveReferences(
+	ctx context.Context,
+	req *api.SdkCredentialRemoveReferencesRequest,
+) (*api.SdkCredentialRemoveReferencesResponse, error) {
+	if s.driver(ctx) == nil {
+		return nil, status.Error(codes.Unavailable, "Resource has not been initialized")
+	}
+
+	if len(req.GetCredentialId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must provide credentials uuid")
```
disrani-px, line 244: This can be name or uuid

api/api.proto
```diff
@@ -2217,6 +2217,14 @@ service OpenStorageCredentials {
         get: "/v1/credentials/validate/{credential_id}"
       };
     }
+  // RemoveReferences is used to remove references to credentials
+  rpc RemoveReferences(SdkCredentialRemoveReferencesRequest)
```
disrani-px, line 2221: I would structure this as DeleteReferences and add an API for GetReferences

The REST API paths could be DELETE /v1/credentials/references/{cred_id} and GET /v1/credentials/references/{cred_id}

api/server/sdk/credentials.go
talakad: Yes. fixed the error.

api/api.proto
talakad: Done

PR #1731, opened 2020-08-22
Add stream interceptors for auth, logging, and rwLock

ggriffiths:

Adds stream interceptors

Closes #

### DIFF
--- a/api/server/sdk/server_interceptors.go
+++ b/api/server/sdk/server_interceptors.go
@@ -42,8 +42,8 @@ const (
 	ContextMetadataTokenKey = "bearer"
 )
 
-// This interceptor provides a way to lock out any calls while we adjust the server
-func (s *sdkGrpcServer) rwlockIntercepter(
+// This interceptor provides a way to lock out any unary calls while we adjust the server
+func (s *sdkGrpcServer) rwlockUnaryIntercepter(
 	ctx context.Context,
 	req interface{},
 	info *grpc.UnaryServerInfo,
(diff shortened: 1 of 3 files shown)


api/server/sdk/server_interceptors.go
```diff
@@ -136,12 +149,34 @@ func (s *sdkGrpcServer) loggerServerInterceptor(
...
+	log := logrus.New()
+	log.Out = s.accessLogOutput
+	logger := log.WithFields(logrus.Fields{
+		"method": info.FullMethod,
+		"reqid":  reqid,
+	})
+
+	logger.Info("Start")
+	ts := time.Now()
+	err := handler(srv, stream)
```
lpabon, line 168:
```go
        var i ..
 	err := accessLogger(func() {
	  i , err := handler(srv, stream)
	  return err
	})
	if err != nil || handler_err != nil ..
	
	func (s *sdkGrpcServer) accessLoggerInter(handler func() error) {
	
```

api/server/sdk/server_interceptors.go
```diff
@@ -163,39 +205,69 @@ func (s *sdkGrpcServer) authorizationServerInterceptor(
...
+) (interface{}, error) {
+	var i interface{}
+	var err error
+
+	err = s.authorizationInterceptor(ctx, func() error {
+		i, err = handler(ctx, req)
+		return err
+	}, info.FullMethod)
+	if err != nil {
+		return i, err
```
lpabon, line 257: This is the same as line 261 -- seems like it can be removed.

PR #1747, opened 2020-09-10
Add new API VolumeUsageByNode

talakad:

Allows to query volume capacity usage by node.
Closes #


api/server/volume.go
```diff
@@ -1305,6 +1305,30 @@ func (vd *volAPI) usedsize(w http.ResponseWriter, r *http.Request) {
...
+
+	d, err := vd.getVolDriver(r)
+	if err != nil {
+		notFound(w, r)
+		return
+	}
+
+	used, err := d.VolumeUsageByNode(nodeID)
+	if err != nil {
+		e := fmt.Errorf("Failed to get VolumeUsageByNode: %s", err.Error())
```
sangleganesh, line 1325: instead of wrapping the error, suggest passing it as it is.

api/server/sdk/volume_ops.go
```diff
@@ -709,6 +709,25 @@ func (s *VolumeServer) CapacityUsage(
 	return resp, nil
 }
 
+/*
+func (s *VolumeServer) VolumeUsageByNode(
+	ctx context.Context,
+	req *api.SdkVolumeUsageByNodeRequest,
+) (*api.SdkVolumeUsageByNodeResponse, error) {
+
+	resp, err := s.driver(ctx).VolumeUsageByNode(req.GetNodeId())
```
sangleganesh, line 718: should we check permission for each volume returned by the API ? 
I think simply adding the following should be enough once you have the response:

@ggriffiths  can you confirm if this is all we need ?

```
// Get spec. This also checks if the parend id exists.
	// This will also check for Ownership_Read access.
	parentVol, err := s.Inspect(ctx, &api.SdkVolumeInspectRequest{
		VolumeId: volumeID,
	})
	if err != nil {
	    // user does not have permission ? 	
	}

```

api/api.proto
```diff
@@ -890,6 +890,26 @@ message CapacityUsageInfo {
...
+  // VolumeName is name of the volume
+  string volume_name = 2;
+  // PoolUUID is the UUID of the pool that this volume belongs to
+  string pool_uuid = 3;
+  // ExclusiveBytes size exclusively used by the volume
+  uint64 exclusive_bytes = 4;
+  // TotalBytes used by the volume
+  uint64 total_bytes = 5;
+  // LocalCloudSnapshot is true if this volume is snapshot created by cloudbackups
+  bool local_cloud_snapshot = 6;
```
sangleganesh, line 905: is this required ? can it be inferred from the volume name ?

sangleganesh (approved):
minor comments looks good otherwise

api/server/sdk/node.go
```diff
@@ -182,3 +182,22 @@ func (s *NodeServer) EnumerateNodeDrainJobs(
...
+func (s *NodeServer) VolumeUsageByNode(
+	ctx context.Context,
+	req *api.SdkVolumeUsageByNodeRequest,
+) (*api.SdkVolumeUsageByNodeResponse, error) {
+
+	resp, err := s.server.driver(ctx).VolumeUsageByNode(req.GetNodeId())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal,
+			"Failed obtain usage stats for volumes in node:%v",
+			req.GetNodeId(),
```
sangleganesh, line 195: suggest no wrapping the error, also has two params

api/server/volume.go
talakad: Done

api/server/sdk/volume_ops.go
talakad: This was commented code that i meant to remove. Sorry missed removing.  The security context comes from volume annotations? and since there are multiple volumes and it's really for the node, moved the api to node service.

api/api.proto
talakad: yes. It can be.

api/server/sdk/node.go
talakad: Done.

api/api.proto
```diff
@@ -890,6 +890,26 @@ message CapacityUsageInfo {
   int64 total_bytes = 3;
 }
 
+message VolumeUsage {
```
lpabon, line 893: There should be a comment for VolumeUsage message so that it shows up in the documentation on https://libopenstorage.github.io/w/index.html

But do not add the name of the message, just describe the message.

Example: https://libopenstorage.github.io/w/release-6.4.generated-api.html#sdkvolumecreaterequest

api/api.proto
```diff
@@ -890,6 +890,26 @@ message CapacityUsageInfo {
...
+   // VolumeID is the id for the volume
```
lpabon, line 894: fyi, there is no need to add the "golang" name in the comment because on gRPC the clients are for different languages. A lot of the users use python.

Instead you can make it look like this:

```proto
  // Id for the volume
  string volume_id = 1;
```

Example: https://libopenstorage.github.io/w/release-6.4.generated-api.html#sdkvolumeattachoptions

api/api.proto
```diff
@@ -890,6 +890,26 @@ message CapacityUsageInfo {
...
+  string pool_uuid = 3;
+  // ExclusiveBytes size exclusively used by the volume
+  uint64 exclusive_bytes = 4;
+  // TotalBytes used by the volume
+  uint64 total_bytes = 5;
+  // LocalCloudSnapshot is true if this volume is snapshot created by cloudbackups
+  bool local_cloud_snapshot = 6;
+}
+
+message VolumeUsageByNode {
```
lpabon, line 908: comment

api/api.proto
```diff
@@ -1824,6 +1844,13 @@ service OpenStorageNode {
       };
   }
 
+	// Returns capacity usage of all volumes/snaps for a give node
+	rpc VolumeUsageByNode(SdkVolumeUsageByNodeRequest)
```
lpabon, line 1848: The messages for the respose and request should be in the following format (I know that some don't do this, but they should :) ):

```
Sdk<Service><Rpc Name><Request/Response>
```

so this would should be `SdkNodeVolumeUsageByNodeRequest`, etc.

if this is not in `docs/dev-sdk.md` please feel free to add it there.

api/server/sdk/node.go
```diff
@@ -182,3 +182,19 @@ func (s *NodeServer) EnumerateNodeDrainJobs(
 
 	return s.cluster().EnumerateNodeDrainJobs(ctx, req)
 }
+
+func (s *NodeServer) VolumeUsageByNode(
+	ctx context.Context,
+	req *api.SdkVolumeUsageByNodeRequest,
+) (*api.SdkVolumeUsageByNodeResponse, error) {
+
```
lpabon, line 190: There needs to be checks here for `.driver()` could return `nil` if the SDK objects are not ready. Check other code for examples.

api/server/sdk/node_test.go
```diff
@@ -379,3 +379,42 @@ func TestSdkNodeInspectCurrent(t *testing.T) {
 	assert.Len(t, rn.GetNodeLabels(), 1)
 	assert.Equal(t, rn.GetNodeLabels(), node.NodeLabels)
 }
+
+func TestSdkVolumeUsageByNode(t *testing.T) {
```
lpabon, line 383: 👍🏼 awesome

lpabon (changes requested):
I have a possible suggestion for the security of the call. We can discuss offline

api/api.proto
talakad: Done

api/api.proto
talakad: Done

api/api.proto
talakad: Done

api/server/sdk/node.go
talakad: thanks!.  Fixed.

api/server/sdk/node.go
```diff
@@ -182,3 +184,52 @@ func (s *NodeServer) EnumerateNodeDrainJobs(
...
+	}
+	// Get the info locally
+	if s.server.driver(ctx) == nil {
+		return nil, status.Error(codes.Unavailable, "Resource has not been initialized")
+	}
+	resp, err := s.server.driver(ctx).VolumeUsageByNode(req.GetNodeId())
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, " Failed to get olumeUsageByNode :%v", err.Error())
+	}
+	//TODO check secruity ( only admin must be allowed to do this
```
lpabon, line 213: nit: I think this comment is no longer necessary. We can fix it in a future PR

api/server/sdk/server.go
```diff
@@ -84,6 +84,9 @@ type ServerConfig struct {
 	// Address is the port number or the unix domain socket path.
 	// For the gRPC Server. This value goes together with `Net`.
 	Address string
+	// Port is the port number at which remote SdkGrpcServer is running.Same
+	// across cluster. Exampl: 9020
+	port string
```
lpabon, line 89: Where is this set?

lpabon (approved):

PR #1767, opened 2020-10-29
Make unit tests compatible with podman

lpabon:
Docker is starting to be replaced by podman, rc, and others. This
change makes it possible to run the unit tests using podman on
the development system. There seems to be an issue with mouting
inside podman and using chattr. By default the unit tests will
not be changed.

A developer using podman or another container runtime can use:

    make TAGS=daemon docker-test


ggriffiths (approved):

lpabon:
Kind doesn't work with podman at the moment. When it does, we will revisit this.

PR #1768, opened 2020-10-29
Add Cloud Drive Transfer job type

harsh-px:

Closes #


api/api.proto
```diff
@@ -3374,6 +3369,10 @@ message Job {
         // NodeDrainAttachmentsJob if selected this job desribes
         // the task for removing volume attachments from a node
         NodeDrainAttachmentsJob drain_attachments = 4;
+        // CloudDriveTransferJob is selected describes the task to transfer a cloud driveset
```
adityadani, line 3372: nit: if selected

api/api.proto
```diff
@@ -3374,6 +3369,10 @@ message Job {
...
+        // from one node to another
+        CloudDriveTransferJob clouddrive_transfer = 401;
```
adityadani, line 3374: Why 401? Also you can reorder the numbers since Job message has not been released yet.

api/api.proto
```diff
@@ -3447,6 +3446,15 @@ message NodeDrainAttachmentsJob {
     google.protobuf.Timestamp last_update_time = 6;
 }
 
+message CloudDriveTransferJob {
+  // SourceNodeID is the ID of the node whose cloud driveset needs to be transferred
+  string source_node_id = 1;
```
adityadani, line 3451: Should this be instance id or scheduler node name? 
For a user this will be a bit convoluted since the source node id is the ID that the destination instance will latch onto.

adityadani (commented):
There is no openstorage API to trigger the transfer?

api/api.proto
harsh-px:
```
  string s3_storage_class = 7;
  // Start at field number 200 to allow for expansion
  oneof credential_type {
    // Credentials for AWS/S3
    SdkAwsCredentialRequest aws_credential = 200;
    // Credentials for Azure
    SdkAzureCredentialRequest azure_credential = 201;
    // Credentials for Google
    SdkGoogleCredentialRequest google_credential = 202;
  }
```

wanted the oneof to have it's own range. e.g credential_type above.

Updated it to 400 and 401.

api/api.proto
harsh-px: renamed to source_drive_set and destination_drive_set and updated the variable doc.

harsh-px (commented):
Yeah no SDK API to initiate cloud drive transfer. Like all cloud drive APIs, this is still driver from the porx HTTP handlers.

adityadani (approved):

PR #1777, opened 2020-11-06
Automated cherry pick of 1775 release 9.0

adityadani:


SDK_CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 # Changelog
 
 ## Releases
+### v0.103.0 - (11/05/2020)
+
+* Change the API definitions for OpenstorageJobServer RPCs
+
+### v0.102.0 - (11/05/2020)
```
lpabon, line 8: These are incorrect. I don't think that this patch has both CloudDriveTransfer and Job apis. Also, it should be v0.101.1

lpabon (changes requested):

SDK_CHANGELOG.md
adityadani: Done

PR #1789, opened 2021-01-08
[CSI] Create mountpath if it does not exist

ggriffiths:
Due to a change in k8s 1.20:
```
ACTION REQUIRED: For CSI drivers, kubelet no longer creates the target_path for NodePublishVolume in accordance with the CSI spec. Kubelet also no longer checks if staging and target paths are mounts or corrupted. CSI drivers need to be idempotent and do any necessary mount verification.
```

Closes #


harsh-px (approved):

adityadani (approved):
lgtm

PR #1793, opened 2021-01-11
cluster/manager: replace logrus.Printf with dbg.LogErrorAndPanicf

saheienko:
Signed-off-by: Serhii Aheienko <serhii.aheienko@gmail.com>

Dump output goroutines to a file instead of OS default logger.

Closes #

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -17,10 +17,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/libopenstorage/openstorage/pkg/job"
-	"github.com/libopenstorage/openstorage/pkg/nodedrain"
-	"github.com/libopenstorage/openstorage/pkg/storagepool"
-
 	"github.com/libopenstorage/gossip"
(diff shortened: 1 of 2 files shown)


cluster/manager/manager.go
```diff
@@ -680,7 +680,7 @@ func (c *ClusterManager) joinCluster(
 	initState, err := snapAndReadClusterInfo(c.snapshotPrefixes)
 	kvdb.Unlock(kvlock)
 	if err != nil {
-		logrus.Panicf("Fatal, Unable to create snapshot: %v", err)
+		dbg.LogErrorAndPanicf(err, "fatal: unable to create snapshot")
```
adityadani, line 683: there is another one on line 709

adityadani (approved):
lgtm after all log.Panic* are removed from cluster/manager package.

cluster/manager/manager.go
saheienko: updated

PR #1798, opened 2021-01-27
Handle spec update for fastpath

prabirpaul:

Need it for setting fastpath volume option

Closes #


vjs-px (approved):

api/api.proto
```diff
@@ -750,6 +750,8 @@ message VolumeSpecPolicy {
   oneof proxy_spec_opt { ProxySpec proxy_spec = 61; }
   // sharedv4_service_spec_opt provides the spec for sharedv4 volume service
   oneof sharedv4_service_spec_opt { Sharedv4ServiceSpec sharedv4_service_spec = 62; }
+  // fastpath preference
+  oneof fastpath_opt { bool fastpath = 63; }
```
sulakshm, line 754: how is volumeSpecPolicy different from VolumeSpecUpdate?
So far, I have been using pxctl volume update --fastpath={true|false} <volname> and they seem to work fine.
Can you help me understand whats broken? based on your response, please see if this needs to go into 2.7.0?

talakad: @prabirpaul , correct me if understood this incorrectly. Someone could create a storagepolicy with fastPath preference.  and then use the policy to create volumes, so that they automatically inherit whas specified in the volumeSpec policy. VolumeSpec update is for updating the volume, which is an action after volume is created.

talakad (approved):

PR #1799, opened 2021-01-28
Add diags SDK

harsh-px:
Signed-off-by: Harsh Desai <hadesai@purestorage.com>

This PR adds the Diags SDK service which has a single Collect method.

The SDK uses the job framework and returns a Job in response. Inside the job, there is a new diags job which has the original request and a list of statuses. The list of statuses will be used when a particular request results in diags collection on multiple nodes.

Closes #


SDK_CHANGELOG.md
```diff
@@ -26,7 +30,7 @@
 
 ### v0.103.0 - (11/05/2020)
 
-* Change the API definitions for OpenstorageJobServer RPCs
+* Change the API definitions for OpenstorageJobServer RPCpx/sched/filter/k8s/k8s.gos
```
lpabon, line 33: Looks like a find and replace issue?

harsh-px: Fixed now

api/api.proto
```diff
@@ -3522,6 +3540,66 @@ message CloudDriveTransferJob {
...
+  string node = 1;
+  // State is the current state of diags collection on the node
+  State state = 2;
+  // Message is a user friendly message for current status of diags collection
+  string message = 3;
+
+  // State is an enum for state of diags collection on a given node
+  enum State {
+    // Pending indicates the diags collection is pending and hasn't started
+    PENDING = 0;
```
lpabon, line 3561: FYI, enum = 0 should be set to `UNSPECIFIED`.

api/api.proto
```diff
@@ -3522,6 +3540,66 @@ message CloudDriveTransferJob {
...
+    DONE = 2;
+    // Failed indicates diags collection has failed
+    FAILED = 3;
+  }
+}
+
+
+message SdkDiagsCollectRequest {
+  // Node selects the node(s) for diags collection
+  DiagsNodeSelector node = 100;
```
lpabon, line 3574: just curious, why out of order numbering?

api/api.proto
```diff
@@ -1827,6 +1827,20 @@ service OpenStoragePool {
   }
 }
 
+// OpenStorageDiags service provides methods to manage diagnostic bundles
+service OpenStorageDiags {
+
+  // Collect diags based on the provided request
```
lpabon, line 1833: I think this needs more documentation on how to use. Also, may want to return a `SdkDiagsCollectResponse` which has a `SdkJobResponse` or `SdkJob` object.

harsh-px: Add documentation and also added a dedicated SdkDiagsCollectResponse.

api/api.proto
```diff
@@ -3522,6 +3542,77 @@ message CloudDriveTransferJob {
...
+  // Volume selects the volume(s) for diags collection
+  DiagsVolumeSelector volume = 2;
+  // ProfileOnly is an optional flag if true will only collect the stack and heap profile of the driver and will skip
+  // other diag components
+  bool profile_only = 3;
+  // Issuer is an optional  user friendly name for the caller invoking the API
+  string issuer = 4;
+  // TimeoutMins is the timeout in minutes for the job. This is an optional field and if not provided, the
+  // implementation of the SDK will use a sane default
+  uint64 timeout_mins = 5;
```
lpabon, line 3588: Use int64. Java has an issue with uint64

api/api.proto
```diff
@@ -3522,6 +3542,77 @@ message CloudDriveTransferJob {
...
+
+// SdkDiagsCollectResponse defines a response for an SDK request to collect diags
+message SdkDiagsCollectResponse {
+  // Job that was created for the SDK request
+  Job job = 1;
+}
+
+// DiagsNodeSelector allows selecting nodes for diags collection
+message DiagsNodeSelector {
+  // NodeLabelSelector is a label selector used to select the nodes for which diags will be collected
```
lpabon, line 3599: Should these three be in a `oneof` ? Sounds like not all three should be chosen. If `all` is true, then there is no need to set the others, right? Or if `node_ids` are set, then labels do not need to be used?

pkg/diags/diags.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/libopenstorage/openstorage/api"
+	"github.com/libopenstorage/openstorage/api/errors"
+	"golang.org/x/net/context"
+)
+
+// Provider is collection of APIs that implement the SDK OpenStorageDiags service
+type Provider interface {
+	// Collect collects diagnostics based on the provided request
+	Collect(ctx context.Context, in *api.SdkDiagsCollectRequest) (*api.SdkDiagsCollectResponse, error)
```
lpabon, line 12: Instead of using a copy of the interface, you can use the interface created in `api.pb.gb` here. See `pkg/role` as an example.

lpabon: https://github.com/libopenstorage/openstorage/blob/0c2f335282d9ed2bb26b5ac9f5f2f2d744271274/pkg/role/role.go#L27

harsh-px: Good idea. Done.

api/api.proto
harsh-px: Done

api/api.proto
harsh-px: Added more documentation.

api/server/sdk/diags.go
```diff
@@ -0,0 +1,28 @@
+package sdk
```
lpabon, line 1: New files need apache lic. That's the last comment

harsh-px: added

harsh-px:
CI seems to be failing due to docker issue but I'm unable to restart it. Luis, do you see a restart option?

PR #1802, opened 2021-02-02
Additions to cloudbackup enumerate api

talakad:

Allows users to identify cloudbackups that do not belong to this cluster or whose source volume may already be deleted.
Closes #


api/api.go
```diff
@@ -398,8 +398,16 @@ type CloudBackupGenericRequest struct {
 	// MetadataFilter indicates backups whose metadata has these kv pairs
 	MetadataFilter map[string]string
 	// CloudBackupID must be specified if one needs to enumerate known single
-	// backup( format is clusteruuidORBucketName/srcVolId-SnapId(-incr)
+	// backup( format is clusteruuidORBucketName/srcVolId-SnapId(-incr). If t\
```
sangleganesh, line 401: nit: space after backup and no space after '('.

// backup (format is clusteruuidORBucketName/SrcVolumeID-SnapId(-incr).
// If specified everything else in the request is ignored

api/api.go
```diff
@@ -398,8 +398,16 @@ type CloudBackupGenericRequest struct {
...
+	// this is specified, everything else n the command is ignored
 	CloudBackupID string
+	// MissingSrcVol set to true enumerates cloudbackups for which srcVol is not
```
sangleganesh, line 404: nit: prefer to use full names when possible.
s/srcVol/SrcVolumeID/

Redo the comment ? The '(' does not end ?

api/api.go
```diff
@@ -416,6 +424,11 @@ type CloudBackupInfo struct {
 	Metadata map[string]string
 	// Status indicates the status of the backup
 	Status string
+	// Cluster indicates if the cloudbackup was uploaded by this
+	// cluster. Could be unknown with older version cloudbackups
+	Cluster SdkCloudBackupClusterType
+	// Namespace to which this cloudbackup belongs to
```
sangleganesh, line 430: is the namespace of the source volume ? 
Also, I would name it ClusterType instead of Cluster for clarity.

sangleganesh (approved):
minor nits.

api/api.go
talakad: yes. k8s namespace that source volume belongs to.

api/api.go
talakad: Done

api/api.go
talakad: Since the field does not represent ID changed it to MissingSrcVolumes

PR #1819, opened 2021-03-10
Auth-middleware delete is not idempotent

lpabon:
* Also fixed an error in auth middleware Enumerate function where
  it returns the incorrect object in an error.

Signed-off-by: Luis Pabón <lpabon@purestorage.com>

Auth middleware has the following errors:

- Deletions of volumes that do not exist where returning error. If a volume does not exist, return success according to idempotency rules
- the client.Inspect() call was receiving the incorrect object from the auth middleware. The auth middleware encountered an error condition it returned the error object instead of an empty list as expected by the client.

Closes #

- Unit tests were updated to completely reset the KV stores used by the `fake` driver and the `cluster` objects.

### DIFF
--- a/api/server/middleware_auth.go
+++ b/api/server/middleware_auth.go
@@ -262,6 +262,7 @@ func (a *authMiddleware) setWithAuth(w http.ResponseWriter, r *http.Request, nex
 
 func (a *authMiddleware) deleteWithAuth(w http.ResponseWriter, r *http.Request, next http.HandlerFunc) {
 	fn := "delete"
+	volumeResponse := &api.VolumeResponse{}
 	d, authRequired := a.isTokenProcessingRequired(r)
 	if !authRequired {
 		next(w, r)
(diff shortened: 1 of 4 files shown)


ggriffiths (approved):
Looks good

api/server/middleware_auth.go
```diff
@@ -358,23 +360,22 @@ func (a *authMiddleware) enumerateWithAuth(w http.ResponseWriter, r *http.Reques
...
 		return
 	}
 
-	volumeResponse := &api.VolumeResponse{}
 	tokenSecretContext, err := a.parseSecret(vols[0].Spec.VolumeLabels, vols[0].Locator.VolumeLabels, false)
 	if err != nil {
 		a.log(volumeID, fn).WithError(err).Error("failed to parse secret")
-		volumeResponse.Error = "failed to parse secret: " + err.Error()
-		json.NewEncoder(w).Encode(volumeResponse)
+		json.NewEncoder(w).Encode(emptyVols)
```
adityadani, line 378: Why are we changing the behavior here? Shouldn't we return an error to the user if he has an incorrect token/secret ?

adityadani (approved):

api/server/middleware_auth.go
lpabon: This is the second bug. The client gets a `inspect volume: json: cannot unmarshal object into Go value of type []*api.Volume` because the client for `Inspect()` calls the server enumerate and doesn't expect an error.

adityadani: But this is not Inspect() but it is parsing an already inspected volume and it is unable to parse the secret from the labels of that volume.

PR #1833, opened 2021-03-26
Add github actions for PR metadata 

ggriffiths:

Closes #

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,11 @@
+area/csi:
+- csi/**/*
+area/ci:
+- .travis
+- .github/**/*
+area/osd-cmd:
+- osd/**/*
+area/api:
+- api/**/*
+area/sdk-server:
+- api/server/sdk/**/*
--- a/.github/workflows/cron-stale.yml
+++ b/.github/workflows/cron-stale.yml
@@ -1,0 +1,17 @@
+name: 'Mark stale issues PR that need review'
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+jobs:
+  stale-in-review:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-pr-message: 'This issue is stale because it has been in review for 5 days with no activity.'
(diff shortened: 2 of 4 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,14 @@
...
+area/ci:
+- .travis
+- .github/**/*
+area/osd-cmd:
+- osd/**/*
+area/api-proto:
+- api/api.pb.go
+- api/api.proto
+area/api:
+- api/*
```
harsh-px, line 12: Let's start with a wider area and keep everything under `api/*` as `area/api`.

.github/workflows/cron-stale.yml
```diff
@@ -0,0 +1,17 @@
...
+    - cron: '30 1 * * *'
+
+jobs:
+  stale-in-review:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/stale@v3
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          stale-pr-message: 'This issue is stale because it has been in review for 2 days with no activity.'
```
harsh-px, line 13: 5 days would be a better conservative start :)

.github/labeler.yml
ggriffiths: Ok, I kept api/server/sdk as well since I think it's helpful to know about SDK server changes. The proto regex I've removed.

ggriffiths: I think these labels will be even more useful in porx

ggriffiths:
The actions will fail on this PR until they're added to the origin - they don't want people running actions on PRs from forks.

.github/labeler.yml
```diff
@@ -0,0 +1,13 @@
...
+- .travis
+- .github/**/*
+area/osd-cmd:
+- osd/**/*
+area/api:
+- api/**/*
+area/sdk-server:
+- api/server/sdk/**/*
+area/pkg:
+- pkg/**/*
```
harsh-px, line 13: the pkg area doesn't add much value since it's the kitchen sink. Can be removed unless you see a use case

harsh-px (approved):

.github/labeler.yml
ggriffiths: Removed

PR #1834, opened 2021-03-29
[CSI] Compatability with CSI 1.4 

ggriffiths:
- Updates CSI spec to 1.4
- Runs CSI-Test against our 1.4 CSI Driver
- Implements a few more optional CSI calls (ControllerGetVolume, NodeGetVolumeStats) - useful for volume health monitoring down the line
- Chose to continue ignoring ListSnapshots for our driver. We have synchronous snapshotting, so once a CreateSnapshot returns, it is ready to use. 

Closes #


csi/controller.go
```diff
@@ -123,6 +110,27 @@ func (s *OsdCsiServer) ControllerUnpublishVolume(
 	return nil, status.Error(codes.Unimplemented, "This request is not supported")
 }
 
+// ControllerGetVolume is a CSI API which implements getting a single volume
+func (s *OsdCsiServer) ControllerGetVolume(
```
lpabon, line 114: Add a comment here about security and how this call can only be used through a unix domain socket by the controller, and that it is a read only request.

csi/node.go
```diff
@@ -246,16 +246,101 @@ func (s *OsdCsiServer) NodeGetCapabilities(
...
+	default:
+		condition.Message = "Volume status is unknown"
+	}
+
+	return condition
+}
+
+// NodeGetVolumeStats get volume stats for a given node
+func (s *OsdCsiServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVolumeStatsRequest) (*csi.NodeGetVolumeStatsResponse, error) {
+	logrus.Debugf("NodeGetVolumeStats req[%#v]", req)
```
lpabon, line 300: This needs to be removed to not hit what I did

csi/node.go
```diff
@@ -246,16 +246,101 @@ func (s *OsdCsiServer) NodeGetCapabilities(
...
+		condition.Message = "Volume status is degraded"
+
+	default:
+		condition.Message = "Volume status is unknown"
+	}
+
+	return condition
+}
+
+// NodeGetVolumeStats get volume stats for a given node
```
lpabon, line 298: Add here information about security just like the others

lpabon (changes requested):

csi/node.go
ggriffiths: Ah duh, good point

csi/node.go
ggriffiths: Added

csi/controller.go
ggriffiths: Added

lpabon (approved):

ggriffiths:
Waiting to do a quick coordinated merge with porx master as this will introduce breaking changes until csi/proto is updated in porx

PR #1836, opened 2021-03-31
Make SDK RoleManager generic.

adityadani:
- Added a new interface Datastore for persisting SDK Role objects.
  Currently the RoleManager assumes that kvdb will be used for persistence
  but with this new interface any backend which implements the Datastore
  interface can be used to persist roles.
- The Default roles by Openstorage are moved to a separate package. This allows
  the role manager to be extensible such that any new projects which leverage
  this role manager can define their own Default Roles.
- Implemented the Datastore interface for kvdb. Using kvdb.Instance() instead of
  the passed in kvdb object since PX recommends using this singleton as it
  gracefully handles lost connectivity with kvdb.

Closes #


pkg/role/datastore.go
```diff
@@ -0,0 +1,162 @@
...
+
+var (
+	// ErrRoleNotFound is returned when role is not found in datastore
+	ErrRoleNotFound = errors.New("role not found in datastore")
+	// ErrRoleExists is returned when a role with the same name exists
+	ErrRoleExists = errors.New("role exists with the same name")
+)
+
+const (
+	rolePrefix = "cluster/roles"
```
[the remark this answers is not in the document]
adityadani: This key will only be used in case of kvdb implementation which is defined in this file. I 
I don't think we will be adding the mongo implementation of Datastore here.

pkg/role/datastore.go
```diff
@@ -0,0 +1,162 @@
...
+// Simple function which creates key for Kvdb
+func prefixWithName(name string) string {
+	return rolePrefix + "/" + name
+}
+
+type kvRoleDatastore struct {
+}
+
+// NewKvdbRoleDatastore returns a kvdb implementation of Datastore interface
+func NewKvdbRoleDatastore() (Datastore, error) {
```
lpabon, line 61: This is a style change, but instead of returning an interface, return the object. Let the caller decide if they want to use the interface. It also helps with unit tests.

If you really want to return an interface, then have `New..` call `new...` which returns the object, that way the unit tests can have access to the real object.

pkg/role/datastore.go
```diff
@@ -0,0 +1,162 @@
...
+const (
+	rolePrefix = "cluster/roles"
+)
+
+// Simple function which creates key for Kvdb
+func prefixWithName(name string) string {
+	return rolePrefix + "/" + name
+}
+
+type kvRoleDatastore struct {
```
lpabon, line 57: All these methods are not part of the gRPC interface as they used to be before which included the `ctx`. If possible, we should try to maintain the receivers to maintain the ctx variable which may have a lot of information about the request like the request id (in a future release).

lpabon (commented):
Everything looks great and clean. Just two comments as suggestions.

ggriffiths (approved):
Looks good

pkg/role/datastore.go
adityadani: Do you mean we should pass in the Context to all the APIs in this Datastore interface? 
I thought that the layer which talks to a persistence store should not worry about who is requesting it and rely on their callers to do the proper context checks.

pkg/role/datastore.go
adityadani: Makes sense. Will do the change.

adityadani:
Not needed anymore.
Will re-open if required.

PR #1851, opened 2021-04-14
Add IPv6 addr handling by adding '[]' around IPv6 connect addresses.

jrivera-px:

Inspect IP and add square brackets around the IPv6 address to be used for the tcp:// & udp:// connect calls.

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -419,8 +419,10 @@ func (c *ClusterManager) getCurrentState() *api.Node {
 	return nodeCopy
 }
 
+// Return ip:port string
 func getPeerAddress(ip string, port string) string {
 	if strings.Contains(ip, ":") {
+		// IPv6 addresses and '[]'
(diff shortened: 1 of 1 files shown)


cluster/manager/manager.go
```diff
@@ -427,8 +427,12 @@ func (c *ClusterManager) getNonDecommisionedPeers(
 		if nodeEntry.Status == api.Status_STATUS_DECOMMISSION {
 			continue
 		}
+		dataIp := nodeEntry.DataIp
```
sangleganesh, line 430: suggest writing a function and using it instead of repeating code:

```
func getPeerAddress(ip string, port string) string
{
      if strings.Contains(ip, ":") {
 			ip = fmt.Sprintf("[%s]", ip)
      }
      return fmt.Sprintf("%s:%s", ip, port)
}
```

sangleganesh (approved):

PR #1853, opened 2021-04-15
Allow pkg/role verifyRule method to be accessible from external lib

ram-infrac:
 This PR  make verifyRule method to be public so that lib  with correct `sdkRule` can use VerifyRule method to validate user

Closes #


prashanthpx (approved):
LGTM

siva-portworx (approved):

PR #1854, opened 2021-04-16
[CSI] Raw Block support

ggriffiths:

Closes #

### DIFF
--- a/csi/node.go
+++ b/csi/node.go
@@ -21,11 +21,10 @@ import (
 	"os"
 	"strings"
 
+	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/pkg/grpcutil"
 	"github.com/portworx/kvdb"
-
-	csi "github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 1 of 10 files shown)


volume/drivers/fake/fake.go
```diff
@@ -226,9 +226,6 @@ func (d *driver) Unmount(volumeID string, mountpath string, options map[string]s
 	if err != nil {
 		return err
 	}
-	if len(v.AttachPath) == 0 {
```
ggriffiths, line 229: Needed for csisanity idempotency

csi/node.go
```diff
@@ -389,3 +412,36 @@ func ensureMountPathCreated(targetPath string) error {
...
+		if err != nil {
+			logrus.Warnf("failed to close file: %s", err.Error())
+		}
+	}()
+	if err != nil {
+		if !os.IsExist(err) {
+			return err
+		}
+	}
+	_, err = f.Stat()
```
lpabon, line 429: Just curious on this style. Don't want to check any stat values? May be just add some comments to why this style is being used.

pkg/mount/raw_mount.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	// try to find all bind mounts of raw volumes
+	if len(rawVolumeDevicesPaths) == 0 || rawVolumeDevicesPaths[0] == "" {
+		mountPointsByMajMin := make(map[string]*[]mount.Info)
+		mountPointsByTarget := make(map[string]*mount.Info)
+
+		for _, mp := range mountPoints {
+			// skip proc mount points
+			if strings.HasPrefix(mp.Mountpoint, "/proc") || mp.Root == "/null" {
```
lpabon, line 69: Maybe add these as a const string array of locations that need to be skipped so we can add more in the future if required.

pkg/mount/raw_mount.go
```diff
@@ -0,0 +1,158 @@
...
+		b.trashLocation,
+	)
+	if err != nil {
+		return err
+	}
+	return b.reload(rootSubstring, newRBM.mounts[rootSubstring])
+}
+
+// this mount filtering implementation is done based on logic implemented in findmnt + libmount
+func (b *rawMounter) Load(rawVolumeDevicesPaths []string) error {
```
lpabon, line 56: These require some unit tests

lpabon (changes requested):

pkg/mount/raw_mount.go
ggriffiths: Will add some tests.

csi/node.go
ggriffiths: Removed the stat call, it was leftover from the original implementation.

pkg/mount/raw_mount.go
ggriffiths: Added tests in `mount_test.go` similar to other mounters @lpabon.

ggriffiths: Having trouble with these tests on Travis. Working on this now.

ggriffiths: Got this passing, had to switch to a test directory that was mounted in the container `/mnt/...`

lpabon (approved):

pkg/mount/raw_mount.go
```diff
@@ -0,0 +1,170 @@
...
+
+		mps := *mountPointsForNumber
+		filteredMPs := []mount.Info{}
+		for i := range mps {
+			if mps[i].Mountpoint != "/dev" {
+				filteredMPs = append(filteredMPs, mps[i])
+			}
+		}
+
+		// TODO: when raw volume is mounted a few times on the same node, that case is not handled
```
adityadani, line 115: The driver (pxd) should not allow mounting the same block device on two different mountpaths  if it does not support it.

pkg/mount/raw_mount.go
```diff
@@ -0,0 +1,170 @@
...
+		// skipping dev mount point
+		for _, mp := range filteredMPs {
+			devicePath := "/dev" + mp.Root
```
adityadani, line 118: Wouldn't mp.Root contain the full device path - `/dev/sda` ?

pkg/mount/raw_mount.go
```diff
@@ -0,0 +1,170 @@
...
+				rm.mounts[devicePath] = mount
+			}
+		}
+	}
+
+	// find raw volume bind mounts
+	for _, rawVolumeDevicePath := range rawVolumeDevicesPaths {
+		var mountPointForRoot *mount.Info
+		for _, mp := range mountPoints {
+			if strings.HasSuffix(rawVolumeDevicePath, mp.Root) {
```
adityadani, line 141: The rawVolumeDevicePath should be a suffix like `/dev/pxd/pxd`. So this should change to
```
strings.HasPrefix(mp.Root, rawVolumeDevicePath)
```
Again this depends on what is mp.Root set to. I believe it is set to the whole device path

adityadani (approved):
Have a few comments. Happy to go over them in an offline call. Apart from that everything else looks good.

csi/node.go
```diff
@@ -202,43 +206,44 @@ func (s *OsdCsiServer) NodeUnpublishVolume(
...
-		} else if err != nil {
-			return nil, status.Errorf(codes.NotFound, "Volume id %s not found: %s",
-				req.GetVolumeId(),
-				err.Error())
-		} else {
-			logrus.Infof("Volume %s was deleted or cannot be found", req.GetVolumeId())
-			return &csi.NodeUnpublishVolumeResponse{}, nil
-		}
+	vols, err := s.driver.Inspect([]string{volumeId})
+	if err == kvdb.ErrNotFound || len(vols) < 1 {
```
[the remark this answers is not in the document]
ggriffiths: Good catch, this was a bad merge on my part. The raw block code was very stale.

pkg/mount/raw_mount.go
ggriffiths: Yes good point, I am building a test image with this change.

PR #1858, opened 2021-04-22
[CSI] Add feature flag for enabling inline volumes

ggriffiths:

* Adds feature flag for enabling inline volume support.
* Inline volume has a security flag in that it allows for creation of volumes without respect to any quotas or k8s RBAC. Customers will need to opt-in to this feature as a result.
* Generic ephemeral volumes should be used instead:
https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes

Closes #

### DIFF
--- a/csi/node.go
+++ b/csi/node.go
@@ -32,6 +32,15 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var (
+	ephemeralDenyList = []string{
+		api.SpecPriorityAlias,
+		api.SpecPriority,
+		api.SpecSticky,
+		api.SpecScale,
+	}
+)
+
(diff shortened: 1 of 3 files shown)


ggriffiths:
Added denylist for parameters we want to disable, as well as a unit test for them.

lpabon (approved):

csi/node.go
```diff
@@ -31,6 +31,13 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+var (
+	ephemeralDenyList = []string{
```
adityadani, line 35: I will add the following to the list:
- api.SpecScale
- api.SpecSticky

adityadani (approved):

PR #1860, opened 2021-04-22
PWX-18693 osd changes for iops throttling

vjs-px:

Closes #


api/spec/spec_handler.go
```diff
@@ -531,6 +535,42 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
 			} else {
 				spec.AutoFstrim = autoFstrim
 			}
+		case api.SpecIoThrottleRdIOPS:
+			if spec.IoThrottle == nil {
+				spec.IoThrottle = &api.IoThrottle{}
+			}
+			if throttleIOPS, err := strconv.ParseUint(v, 10, 64); err != nil {
+				return nil, nil, nil, err
+			} else {
```
talakad, line 544: nit: no need for else.  value can be simply set since the if case, takes care of error. Smae for other params added below.

talakad (approved):
small changes suggested. rest looks good

api/spec/spec_handler.go
vjs-px: without the else clause, the "throttleIOPS" would be undeclared and I cannot directly assign the output of strconv.ParseUint() to spec.IoThrottle.ReadIops. 

Unless I declare a variable in the beginning of the function, I cannot avoid the else clause

talakad: waht I meant was: 
`throttleIOPS, err := strconv.ParseUint(v, 10, 64)
if err != nil {
	return nil, nil, nil, err
}
spec.IoThrottle.WriteIops = uint32(throttleIOPS)
`
However its OK to have them in else too.

PR #1863, opened 2021-04-24
Add timeouts to all RPCs

lpabon:
Need timeouts on all RPC calls

* This fix also cleans up the gRPC client connect to use
`grpc.WithBlock()` and `grpc.DialWithContext()` since the APIs we used
to use where experimental.
* Needed to bump the docker-test golang version from 1.10 to something more than 1.13 since we require `http.NewRequestWithContext()`. FYI, when I used 1.15 or 1.16 golang versions, the unit test SDK server did not allow connections. OSD built with 1.15 did allow. Something to investigate later.

### DIFF
--- a/api/client/request.go
+++ b/api/client/request.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -15,6 +16,7 @@ import (
 	"time"
 
 	"github.com/libopenstorage/openstorage/pkg/auth"
+	"github.com/libopenstorage/openstorage/pkg/grpcutil"
 )
 
 const (
@@ -41,6 +43,7 @@ type Request struct {
 	timeout     time.Duration
 	authstring  string
 	accesstoken string
+	ctx         context.Context
 }
 
 // Response is a representation of HTTP response received from the server.
@@ -59,6 +62,19 @@ type Status struct {
 
 // NewRequest instance
 func NewRequest(client *http.Client, base *url.URL, verb string, version string, authstring, userAgent string) *Request {
+	return NewRequestWithContext(nil, client, base, verb, version, authstring, userAgent)
+}
+
+// NewRequestWithContext takes in context which may describe a timeout
+func NewRequestWithContext(
+	ctx context.Context,
+	client *http.Client,
+	base *url.URL,
+	verb string,
+	version string,
+	authstring,
+	userAgent string,
+) *Request {
 	r := &Request{
 		client:     client,
 		verb:       verb,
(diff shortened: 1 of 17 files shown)


api/client/request.go
```diff
@@ -57,15 +59,33 @@ type Status struct {
 	ErrorCode int
 }
 
+var (
+	DefaultContextTimeout = 5 * time.Minute
+)
+
 // NewRequest instance
 func NewRequest(client *http.Client, base *url.URL, verb string, version string, authstring, userAgent string) *Request {
```
ggriffiths, line 67: Would it be better to introduce an API breaking change and add context here? That would ensure usage of context, instead of giving the option to do so.

pault84 (approved):
LGTM

Let's run a SMOKE and BVT to see if we need to alter the timeout length

api/client/request.go
lpabon: I thought about that, I would prefer to avoid a big impact first. We can adjust as we go with this change. Also, since each package has a `DefaultContextTimeout` var, the caller can affect the default just by changing that value.

ggriffiths: Makes sense :+1:

ggriffiths (approved):

pkg/grpcserver/grpcutil.go
```diff
@@ -25,14 +25,16 @@ import (
...
-	"github.com/libopenstorage/openstorage/pkg/util"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/metadata"
 )
 
+var (
+	DefaultContextTimeout = 5 * time.Minute
```
lpabon, line 35: Change to 1 minute

pkg/defaultcontext/default.go
```diff
@@ -0,0 +1,85 @@
+/*
+Package grpcserver is a generic gRPC server manager
```
ggriffiths, line 2: nit: fix this comment

pkg/defaultcontext/default.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// GetDefaultTimeout returns the default timeout duration used by contexts without a timeout
+func (d *defaultContextManager) GetDefaultTimeout() time.Duration {
+	d.lock.RLock()
+	defer d.lock.RUnlock()
+
+	return d.timeout
+}
+
```
ggriffiths, line 81: nit: a new function `WithTimeout(context ctx)` could make things cleaner.

i.e....
```
ctx := context.Background()

ctx = defaultcontext.WithTimeout(ctx)
```

ggriffiths: and when we add a global correlation ID starting in the CSI driver, 
```
ctx = defaultcontext.WithCorrelationID(ctx)
```

ggriffiths (approved):
LGTM just a few nits

PR #1864, opened 2021-04-24
[NSTK-55] API updates to add Pure pass through volume specs into ProxySpec

pure-jliao:
**What this PR does / why we need it**:  API updates to add Pure volume specs into ProxySpec, and pass mountOptions to Mount call
* `endpoint` parameter specified as `pure_file` or `pure_block` to create pure passthrough volumes
* Pure endpoint type is stored as ProxyProtocol internally

[NSTK-55](https://portworx.atlassian.net/browse/NSTK-55?jql=project%20%3D%20NSTK)


api/api.proto
```diff
@@ -359,6 +359,10 @@ enum ProxyProtocol {
     PROXY_PROTOCOL_S3 = 2;
     // PXD indicates that the external data source is a Portworx block volume.
     PROXY_PROTOCOL_PXD = 3;
+    // PURE_BLOCK indicates that the external data source is a pure_block volume.
+    PROXY_PROTOCOL_PURE_BLOCK = 4;
+    // PURE_FILE indicates that the external data source is a pure_file volume.
+    PROXY_PROTOCOL_PURE_FILE = 5;
```
ggriffiths, line 365: For these changes, you will need to bump the API minor version in `api.proto`. Found here
https://github.com/libopenstorage/openstorage/blob/master/api/api.proto#L4863

After bumping the minor version, please add some release notes here (new version will be 116)
https://github.com/libopenstorage/openstorage/blob/master/SDK_CHANGELOG.md

ggriffiths (commented):
I still need to do a full review, but here's some tips to get the CI passing:

1. Make sure all files are `go fmt` with go 1.14 or greater - looks like travis is complaining about this
2. Make sure to bump the API version and add release notes (needed for any api.proto changes)
3. Sign your commit with `git commit -s`

csi/node.go
```diff
@@ -113,6 +114,15 @@ func (s *OsdCsiServer) NodePublishVolume(
 	// Get volume encryption info from req.Secrets
 	driverOpts := s.addEncryptionInfoToLabels(make(map[string]string), req.GetSecrets())
 
+	// Parse storage class 'mountOptions' flags from CSI req
+	// flags from 'mountOptions' will be used as the only source of truth for Pure volumes upon mounting
+	if req.GetVolumeCapability() != nil && req.GetVolumeCapability().GetMount() != nil {
+		mountFlags := strings.Join(req.GetVolumeCapability().GetMount().GetMountFlags(), ",")
+		if mountFlags != "" {
+			driverOpts[api.SpecMountOptions] = mountFlags
```
ggriffiths, line 122: Let's create a new `api.SpecCSIMountOptions = csi_mount_options`, separate from the standard Portworx mount options we support. 

I don't want to conflict with any existing Portworx mount options that can be passed in via storageclass parameters. This might not affect Pure integration, but if we want to support csi_mount_options in Portworx in the future, it will make things easier.

ggriffiths: Also, let's create a separate `mountOptions` map here and pass that to mounts.Mount below. Let's not re-use the driverOpts used for encryption secrets on attach.

csi/node.go
```diff
@@ -113,6 +114,15 @@ func (s *OsdCsiServer) NodePublishVolume(
 	// Get volume encryption info from req.Secrets
 	driverOpts := s.addEncryptionInfoToLabels(make(map[string]string), req.GetSecrets())
 
+	// Parse storage class 'mountOptions' flags from CSI req
+	// flags from 'mountOptions' will be used as the only source of truth for Pure volumes upon mounting
+	if req.GetVolumeCapability() != nil && req.GetVolumeCapability().GetMount() != nil {
```
ggriffiths, line 119: Can you add a unit test for this in `node_test.go`? It can simply validate that the mount flags are being added to the expected mock mount request.

You can copy this test and edit it to this:
https://github.com/libopenstorage/openstorage/blob/master/csi/node_test.go#L410

```
		s.MockDriver().
			EXPECT().
			Mount(name, targetPath, map[string]string{"mount-opts": "here"}).
			Return(nil).
			Times(1),
```

ggriffiths (changes requested):
Mostly looks good, just some cleanup and renaming suggestions.

csi/node.go
ggriffiths: Discussed offline, okay to use existing driverOpts with the new csi mount options flag.

api/api.go
```diff
@@ -91,6 +92,10 @@ const (
 	SpecSharedv4ServiceName  = "sharedv4_svc_name"
 	SpecFastpath             = "fastpath"
 	SpecAutoFstrim           = "auto_fstrim"
+	SpecBackendType          = "backend"
+	SpecBackendPureBlock     = "pure_block"
+	SpecBackendPureFile      = "pure_file"
+	SpecPureFileExportRules  = "export_rules"
```
adityadani, line 98: Can we call this `pure_export_rules` ?

adityadani (approved):
looks good

api/api.go
```diff
@@ -61,6 +61,7 @@ const (
 	SpecExportOptions        = "export_options"
 	SpecExportOptionsEmpty   = "empty_export_options"
 	SpecMountOptions         = "mount_options"
+	SpecCSIMountOptions      = "csi_mount_options"
```
Pure-AdamuKaapan, line 64: Out of curiosity, why didn't we do some sort of merge between CSI and <wherever else mount options can come from> and store them in the same field? I can't think of any reason off the top of my head for why we'd want to treat them differently, but I could definitely have just missed a conversation.

api/api.go
```diff
@@ -91,6 +92,10 @@ const (
 	SpecSharedv4ServiceName  = "sharedv4_svc_name"
 	SpecFastpath             = "fastpath"
 	SpecAutoFstrim           = "auto_fstrim"
+	SpecBackendType          = "backend"
```
Pure-AdamuKaapan, line 95: Is the idea that for other vendors/"backend" types, they'd add their own `SpecBackend<Vendor><Type>` constant?

api/api.go
```diff
@@ -91,6 +92,10 @@ const (
...
+	SpecBackendPureBlock     = "pure_block"
+	SpecBackendPureFile      = "pure_file"
+	SpecPureFileExportRules  = "pure_export_rules"
```
Pure-AdamuKaapan, line 98: We could make this `SpecNFSExportRules` so that other vendors that support NFS can use it?

api/api.proto
```diff
@@ -381,6 +385,14 @@ message PXDProxySpec {
     string id = 1;
 }
 
+// PureBlockSpec is the spec for proxying a volume on pure_block backends
+message PureBlockSpec {
```
Pure-AdamuKaapan, line 389: Do we want to add some sort of "TODO" here for implementing this later, or specifically leave it out? @ggriffiths @pure-jliao 

I can think of reasons for both so I'm generally indifferent, just want to propose the idea

api/api.proto
```diff
@@ -395,6 +407,10 @@ message ProxySpec {
     S3ProxySpec s3_spec = 4;
     // PXDProxySpec is the spec for proxying a Portworx volume
     PXDProxySpec pxd_spec = 5;
+    // PureFileSpec is the spec for proxying a volume on pure_file backends
+    PureBlockSpec pure_block_spec = 6;
```
Pure-AdamuKaapan, line 411: I thought we ended up saying that passthrough wouldn't be implemented using "proxy volumes"? Is this:
a) a naming difference and "proxy" in `libopenstorage` means something different than in internal code, or
b) the way we're implementing it uses proxy functionality in `libopenstorage` and the difference is purely in internal code?

Pure-AdamuKaapan (commented):
Left a couple questions on some technicalities, but overall the code is looking good. Will approve once I have answers to most of them, thanks!

csi/node.go
pure-jliao: @Pure-AdamuKaapan this might help answer your question about csi_mount_options

api/api.go
pure-jliao: @ggriffiths has a comment about this, ping you there.

api/api.go
pure-jliao: yes this could be potentially used for the other type of backend in the future, now we use `backend` key but in the code we translate it to `ProxyProtocol`, and save the specs for new backend into `ProxySpec` as discussed offline

api/api.go
pure-jliao: change requested by @adityadani . I think. it makes sense, right now it's for FB only, and so far we don't have a standard for it yet,

api/api.proto
pure-jliao: I created this just for future usage, now FA pass through is not supported and requests will be rejected if the backend is pure_block

api/api.proto
pure-jliao: it's just store pure backend info into the ProxySpec, and make everything clean and manageable instead of making new fields in VolumeSpec which is already pretty large now , just for internal usage, the sc parameter key is set to `backend` instead of the original proxy volume keys to reduce confusion, and on porx side, we don't go through the proxy volume filter.

Pure-AdamuKaapan (approved):
LGTM!

PR #1866, opened 2021-04-30
[PWX-19922] Don't pass shared, sharedv4, secure parameters to Pure volume specs

pure-jliao:
**What this PR does / why we need it**: Portworx volume spec parameters including `shared`, `sharedv4` and `secure` doesn't apply to pure volume, and shouldn't be set when parsing from storage classes

Closes # [PWX-19922](https://portworx.atlassian.net/browse/PWX-19922)

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -591,6 +591,11 @@ func resolveBlockSpec(spec *api.VolumeSpec, req *csi.CreateVolumeRequest) (*api.
 // 1. When a volume is set to RWX or a similar multi-node access mode, we default to Sharedv4
 // 2. If a user prefers shared over sharedv4, they may still use it by explicity declaring "shared": true
 func resolveSharedSpec(spec *api.VolumeSpec, req *csi.CreateVolumeRequest) (*api.VolumeSpec, error) {
+	// shared or sharedv4 parameter doesn't apply to pure backends so don't set them
+	if spec.IsPureVolume() {
+		return spec, nil
+	}
+
 	var shared bool
 	for _, cap := range req.GetVolumeCapabilities() {
 		mode := cap.GetAccessMode().GetMode()
--- a/api/api.go
+++ b/api/api.go
@@ -1200,6 +1200,10 @@ func (v *VolumeSpec) IsPublic(accessType Ownership_AccessType) bool {
 	return v.GetOwnership() == nil || v.GetOwnership().IsPublic(accessType)
 }
 
+func (v *VolumeSpec) IsPureVolume() bool {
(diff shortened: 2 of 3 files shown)


api/spec/spec_handler.go
```diff
@@ -309,12 +309,14 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
...
 			if secure, err := strconv.ParseBool(v); err != nil {
 				return nil, nil, nil, err
-			} else {
+			} else if !spec.IsPureVolume() {
 				spec.Encrypted = secure
 			}
 		case api.SpecPassphrase:
-			spec.Encrypted = true
-			spec.Passphrase = v
+			if !spec.IsPureVolume() {
```
adityadani, line 316: Why do we need to do this? I would suggest instead let this module set all the VolumeSpec params as passed in opts. 
The PureVolume filter should return a proper error when it sees conflicting params set. With this change the user will be able to create a volume but we won't error out and the user might think they created a Pure volume with encryption set.

api/spec/spec_handler.go
```diff
@@ -542,6 +544,12 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
 				}
 				spec.ProxySpec.ProxyProtocol = backendType
 			}
+			// Those spec parameters should not be set for pure volume
```
adityadani, line 547: Don't overwrite users params. Return an error instead saying `cannot set shared/sharedv4/secure flags when using pure backend`

adityadani (approved):
lgtm after the changes suggested in spec handler are addressed

csi/controller.go
```diff
@@ -602,11 +602,14 @@ func resolveSharedSpec(spec *api.VolumeSpec, req *csi.CreateVolumeRequest) (*api
 	}
 
 	// Handle legacy support for shared
-	if spec.Shared && !spec.Sharedv4 {
-		spec.Shared = shared
-	} else {
-		// assume sharedv4 by default
-		spec.Sharedv4 = shared
+	// shared or sharedv4 parameter doesn't apply to pure backends so don't set them
+	if !spec.IsPureVolume() {
```
ggriffiths, line 606: Let's move this to the beginning of `resolveSharedSpec`:

```
if spec.IsPureVolume() {
  return spec, nil
}
```

Will make this case easier to read.

ggriffiths (changes requested):
Mostly looks good. We can end early and avoid checking for the RWX capability for pure backends.

ggriffiths (approved):
Looks good, thanks!

PR #1870, opened 2021-05-04
[CSI] Round volume size up to nearest GiB

ggriffiths:

Rounds up volume size to nearest GiB to match in-tree behavior:
https://github.com/kubernetes/kubernetes/blob/442a69c3bdf6fe8e525b05887e57d89db1e2f3a5/pkg/volume/portworx/portworx_util.go#L63

Closes #

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -19,11 +19,14 @@ package csi
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 
 	"github.com/portworx/kvdb"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/pkg/grpcutil"
+	"github.com/libopenstorage/openstorage/pkg/units"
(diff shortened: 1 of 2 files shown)


csi/controller.go
```diff
@@ -49,7 +51,8 @@ const (
 	volumeCapabilityMessageNotMultinodeVolume = "Volume is not a multinode volume"
 	volumeCapabilityMessageReadOnlyVolume     = "Volume is read only"
 	volumeCapabilityMessageNotReadOnlyVolume  = "Volume is not read only"
-	defaultCSIVolumeSize                      = uint64(1024 * 1024 * 1024)
+	defaultCSIVolumeSize                      = uint64(GiB * 1)
+	GiB                                       = 1024 * 1024 * 1024
```
harsh-px, line 55: can we use https://github.com/libopenstorage/openstorage/tree/master/pkg/units instead?

trierra (approved):
lgtm

csi/controller.go
ggriffiths: Updated

harsh-px (approved):

PR #1871, opened 2021-05-04
Add API to support credential update

talakad:

Closes #


api/server/sdk/credentials.go
```diff
@@ -589,3 +589,194 @@ func validateAndDeleteIfInvalid(ctx context.Context, s *CredentialServer, uuid s
...
+	if len(req.GetBucket()) != 0 {
+		params[api.OptCredBucket] = req.GetBucket()
+	}
+	if len(aws.GetRegion()) != 0 {
+		params[api.OptCredRegion] = aws.GetRegion()
+	}
+	if len(aws.GetEndpoint()) != 0 {
+		params[api.OptCredEndpoint] = aws.GetEndpoint()
+	}
+	if len(aws.GetAccessKey()) != 0 {
```
ram-infrac, line 638: should we put additional checks before calling driver update for missing access/secrets ?

ram-infrac (approved):
lgtm

api/server/sdk/credentials.go
talakad: It would be good for server to handle those. Hence not included here.

adityadani: Isn't this server itself ? I see the checks on google but not for aws and azure

api/server/sdk/credentials.go
```diff
@@ -589,3 +589,194 @@ func validateAndDeleteIfInvalid(ctx context.Context, s *CredentialServer, uuid s
...
+
+func (s *CredentialServer) googleUpdate(
+	ctx context.Context,
+	req *api.SdkCredentialCreateRequest,
+	google *api.SdkGoogleCredentialRequest,
+	credId string,
+) (*api.SdkCredentialUpdateResponse, error) {
+
+	if len(google.GetJsonKey()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Must supply JSON Key")
```
adityadani, line 710: Why are we not doing the same checks for azure and aws ?

adityadani (approved):
looks good

api/server/sdk/credentials.go
talakad: Thanks!, actually these checks are not needed. Users need not provide these, if they dont want to modify them. I should take these checks out.

api/server/sdk/credentials.go
talakad: What I meant is combination of these fields for an update the implementation can decide to check on in whatever way they want. Here we dont need to check as not all of these fields are mandatory for update.

PR #1880, opened 2021-05-07
Add Cluster Manager options to set quorum timeout and lock try duration.

adityadani:

- Added two new options to ClusterConfig
  - QuorumTimeoutInSeconds configures time after which an out of quorum node
    will restart
  - SnapLockTryDurationInMinutes is the time for which
    the cluster manager will try acquiring a lock for cluster snapshot


cluster/manager/manager.go
```diff
@@ -1062,9 +1070,19 @@ func (c *ClusterManager) GetGossipState() *cluster.ClusterState {
...
-func (c *ClusterManager) waitForQuorum(exist bool) error {
+func (c *ClusterManager) getMaxQuorumRetries() int {
 	// Max quorum retries allowed = 600
 	// 600 * 2 seconds (gossip interval) = 20 minutes before it restarts
+	quorumRetries := defaultQuorumRetries
+	if c.config.QuorumTimeoutInSeconds > 0 {
+		quorumRetries = c.config.QuorumTimeoutInSeconds / 2
+	}
+	return quorumRetries
+
```
pureneelesh, line 1081: nit: remove empty line

cluster/manager/manager.go
```diff
@@ -1084,7 +1102,7 @@ func (c *ClusterManager) waitForQuorum(exist bool) error {
 			break
 		} else {
 			c.status = api.Status_STATUS_NOT_IN_QUORUM
-			if quorumRetries == 600 {
+			if quorumRetries == maxQuorumRetries {
 				err := fmt.Errorf("Unable to achieve Quorum." +
 					" Timeout 20 minutes exceeded.")
```
pureneelesh, line 1107: This log message might be misleading if the timeout has been changed from the default value of 20 min.

cluster/manager/manager.go
```diff
@@ -689,15 +693,19 @@ func (c *ClusterManager) joinCluster(
...
 	// cluster db state.
 
-	kvdb := kvdb.Instance()
-	kvlock, err := kvdb.LockWithID(clusterLockKey, c.config.NodeId)
+	lockKey := path.Join(c.config.NodeId, self.DataIp, fn)
+	lockTryDuration := defaultClusterLockTryDuration
+	if c.config.SnapLockTryDurationInMinutes > 0 {
+		lockTryDuration = time.Duration(c.config.SnapLockTryDurationInMinutes) * time.Minute
+	}
+	kvlock, err := kvdb.Instance().LockWithTimeout(clusterLockKey, lockKey, lockTryDuration, kvdb.DefaultLockTryDuration)
```
pureneelesh, line 701: Last parameter is called "lock hold duration" in kvdb.go. Just making sure that we are passing the right value here.

Names of these parameters are too similar to each other. Wonder if there is some way to clarify what's what using more specific names. Just a nit.

pureneelesh (approved):

cluster/manager/manager.go
adityadani: fixed

nrevanna (approved):

cluster/manager/manager.go
```diff
@@ -671,6 +674,7 @@ func (c *ClusterManager) initNodeInCluster(
 func (c *ClusterManager) joinCluster(
 	self *api.Node,
 ) error {
+	fn := "joinCluster"
```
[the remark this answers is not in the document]
adityadani: Ideally this should be done by the logging package itself or our own wrapper over logrus, instead of each function doing a runtime.Callers to get a `stack` instance pointer and then passing it to runtime.FuncForPC

PR #1881, opened 2021-05-08
[CSI] Fix Volume Snapshot restore size

ggriffiths:

Fixes the volume snapshot restore size to be non-zero


ggriffiths:
Only fixed the "snapshot already exists" case before. For new snapshots, we were adding the size from `v`, which was nil if the snapshot did not exist yet.

PR #1887, opened 2021-05-13
WIP: separate out node cache update

pureneelesh:
updateClusterStatus() function had dual purpose. It would update the node cache with the new info received from gossip and it would also notify the listeners of any changes in the node status. If one of the listeners took long time to process the change notification, node cache updates stopped. This patch refactors node-cache updates and listener notifications into separate goroutines so that one does not block the other.

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -845,28 +845,15 @@ func (c *ClusterManager) startHeartBeat(
 	}
 }
 
-func (c *ClusterManager) updateClusterStatus() {
+func (c *ClusterManager) notifyListeners() {
 	gossipStoreKey := types.StoreKey(heartbeatKey + c.config.ClusterId)
 	for {
-		node := c.getCurrentState()
-		c.putNodeCacheEntry(node.Id, *node)
-
 		// Process heartbeats from other nodes...
 		gossipValues := c.gossip.GetStoreKeyValue(gossipStoreKey)
 
-		numNodes := 0
 		for id, gossipNodeInfo := range gossipValues {
-			numNodes = numNodes + 1
-
-			size := atomic.LoadInt64(c.size)
-			// Check to make sure we are not exceeding the size of the cluster.
-			if size > 0 && int64(numNodes) > size {
-				logrus.Fatalf("Fatal, number of nodes in the cluster has"+
-					"exceeded the cluster size: %d > %d", numNodes, size)
-			}
-
 			// Special handling for self node
-			if id == types.NodeId(node.Id) {
+			if id == types.NodeId(c.selfNode.Id) {
 				// TODO: Implement State Machine for node statuses similar to the one in gossip
 				if c.selfNode.Status == api.Status_STATUS_OK &&
 					gossipNodeInfo.Status == types.NODE_STATUS_SUSPECT_NOT_IN_QUORUM {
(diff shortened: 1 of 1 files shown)


cluster/manager/manager.go
```diff
@@ -992,37 +964,76 @@ func (c *ClusterManager) updateClusterStatus() {
...
+		}
+	}
+	apiNode.Id = string(id)
+	if gossipNodeInfo.Status == types.NODE_STATUS_DOWN {
+		apiNode.Status = api.Status_STATUS_OFFLINE
+	} else {
+		if apiNode.Status == api.Status_STATUS_MAINTENANCE {
+			maintenance = true
+		}
+		apiNode.Status = api.Status_STATUS_OK
```
adityadani, line 1032: Lets set this to OK only if `gossipNodeInfo.Status == api.Status_STATUS_ONLINE`
In this way for any other statuses we will re-use what what is embedded in `gossipNodeInfo.Value.(api.Node)`

Also we don't need to return the maintenance boolean here since apiNode.Status is already set to Maintenance state.

adityadani: FYI: 
`gossipNodeInfo.Status` is the status of a peer node as seen by our gossip module
`gossipNodeInfo.Value.(api.Node)` is the object that is being gossiped by the peer node. It also has a status that the peer node is advertising for itself. But this could be a stale status, which is why we give gossipNodeInfo.Status highest privilege.

cluster/manager/manager.go
```diff
@@ -992,37 +964,76 @@ func (c *ClusterManager) updateClusterStatus() {
...
+				continue
+			}
+
+			// Update cache with gossip data
+			peerNode, maintenance := c.gossipNodeToAPINode(id, &gossipNodeInfo)
+			if maintenance {
+				// restore the maintenance status which was overwritten by gossipNodeToAPINode()
+				peerNode.Status = api.Status_STATUS_MAINTENANCE
+			}
+			c.putNodeCacheEntry(peerNode.Id, *peerNode)
```
adityadani, line 1006: I am also thinking if we should skip updating our cache if we are in the else condition on line 1022. Why update the cache with an invalid/empty api.Node object. Instead we should read that node's entry in the cache and simply just update the Status field.  Wdyt ?

cluster/manager/manager.go
pureneelesh: > Lets set this to OK only if `gossipNodeInfo.Status == api.Status_STATUS_ONLINE`
> In this way for any other statuses we will re-use what what is embedded in `gossipNodeInfo.Value.(api.Node)`
> 
I was trying match exactly what the current code is doing. Your suggestion will introduce a change in behavior for the listeners. Currently, listeners don't see any status other than api.Status_STATUS_OK or api.Status_STATUS_OFFLINE. I don't know if the listeners will be able to handle other statuses. Do you still want me to make this change?

cluster/manager/manager.go
pureneelesh: Is there a possibility that the gossibNodeInfo.Value() would go from <populated> to <empty>? If not, it does not matter. I would have expected the <empty> value only intially, but I don't know how it works.

pureneelesh:
Merged via https://github.com/libopenstorage/openstorage/pull/1891

cluster/manager/manager.go
pureneelesh: Created https://portworx.atlassian.net/browse/PWX-20197 to address this.

cluster/manager/manager.go
pureneelesh: Per slack conversation, leave it as it is.

PR #1898, opened 2021-06-04
PWX-20172: Make sure the volume is detached before we return

pault84:
Signed-off-by: Paul Theunis <ptheunis@purestorage.com>

Makes it so PVC's don't get bound until PX is done detaching.
If PVC's get stuck in detaching we'd mark volumes/pvc's as ready which is incorrect.
If a detach fails we retry the create. the volume already exists and has UP state at which point we bind the PVC.

PWX-20172


pault84:
Do i need to update the API version?

ggriffiths (approved):

PR #1905, opened 2021-07-07
[PWX-20437] Add a flag in Sharedv4ServiceSpec to indicate whether the service needs to be accessed outside of the cluster

JimOuPure:
Signed-off-by: Jim Ou <jqou@purestorage.com>

**Which issue(s) this PR fixes** [PWX-20437]


api/spec/spec_handler.go
```diff
@@ -452,6 +453,15 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
 			} else if v == "LoadBalancer" || v == "loadBalancer" {
 				spec.Sharedv4ServiceSpec.Type = api.Sharedv4ServiceSpec_LOADBALANCER
 			}
+		case api.SpecSharedv4ExternalAccess:
+			if spec.Sharedv4ServiceSpec == nil {
+				spec.Sharedv4ServiceSpec = &api.Sharedv4ServiceSpec{}
+			}
+			if v == "true" || v == "True" {
```
ggriffiths, line 460: It is a little more reliable to use `strconv.ParseBool(v)` (see line 494 below for an example - https://github.com/libopenstorage/openstorage/pull/1905/files#diff-2e2db81cab87c799e2f24955e8905fc6fdbb5a5fcdfc837dbe8ad4f00c4f4b39R494)

adityadani (approved):

PR #1915, opened 2021-07-31
Add Correlation tracing via logrus hooks with CSI

ggriffiths:
* adds correlation tracing package
* adds CSI driver as first consumer of correlation tracing
* Correlation tracing uses logrus hooks to add info from context.Context
* Correlation is generated at the CSI driver layer. 
* CSI driver is being used to test this initially. We'll add this elsewhere later on.

Closes #

There are two ways to use the correlation package:
1. Create logger per-package (more refactor required, but allows per-component log fields to be added. I.e. for CSI, we will have `component=csi-driver` on every log.
2. Global logger - this will be set on a global level. Whenever `logrus.WithContext(ctx).Infof...` or similar is called, we will always log the correlation ID and origin. This will be very helpful at first, before we do per-component loggers.


ggriffiths:
WIP. Not really ready for review yet. need to test on a live cluster next.

ggriffiths:
Tested w/ PX
```
grant@colo-70-0-0-129:~/workspace/kubeup$ kn logs px-cluster-72d387a1-6e69-46e8-8799-7e9b08ee8fe1-4lkf6 -c portworx | grep 21e3392e-d5d9-488d-b392-abc0e681d085
@grant-k8s-1-node1 portworx[25664]: time="2021-08-02T19:35:50Z" level=info msg="csi.NodePublishVolume request received. VolumeID: 702582855451624944, TargetPath: /var/lib/kubelet/pods/c26e0d29-60f8-4e61-a0e8-c3b8d69bd82d/volumes/kubernetes.io~csi/pvc-706969a5-57a0-4ade-9eb9-0b4cecdbb183/mount" component=csi-driver correlation-id=21e3392e-d5d9-488d-b392-abc0e681d085 origin=csi-driver
@grant-k8s-1-node1 portworx[25664]: time="2021-08-02T19:35:53Z" level=info msg="CSI Volume 702582855451624944 mounted on /var/lib/kubelet/pods/c26e0d29-60f8-4e61-a0e8-c3b8d69bd82d/volumes/kubernetes.io~csi/pvc-706969a5-57a0-4ade-9eb9-0b4cecdbb183/mount" component=csi-driver correlation-id=21e3392e-d5d9-488d-b392-abc0e681d085 origin=csi-driver
```

ggriffiths:
offline feedback from @lpabon :
* Add choice to have a per-function logger instance based on the context
* Provide some examples in README.md for the package in openstorage

api/server/sdk/server.go
```diff
@@ -467,11 +468,17 @@ func (s *sdkGrpcServer) Start() error {
...
+	correlationInterceptor := correlation.ContextInterceptor{
+		Origin: correlation.ComponentSDK,
+	}
+
 	// Setup authentication and authorization using interceptors if auth is enabled
 	if len(s.config.Security.Authenticators) != 0 {
 		opts = append(opts, grpc.UnaryInterceptor(
 			grpc_middleware.ChainUnaryServer(
 				s.rwlockUnaryIntercepter,
+				correlationInterceptor.ContextUnaryInterceptor,
```
lpabon, line 481: nice

csi/csi_test.go
```diff
@@ -391,3 +393,45 @@ func TestAddEncryptionInfoToLabels(t *testing.T) {
...
+			Times(1),
+		s.MockDriver().
+			EXPECT().
+			Enumerate(nil, nil).
+			Return([]*api.Volume{}, nil).
+			Times(1),
+	)
+
+	var buf bytes.Buffer
+	clogger.SetOutput(&buf)
```
lpabon, line 416: nice trick

pkg/auth/auth.go
```diff
@@ -18,8 +18,15 @@ package auth
 
 import (
 	"context"
+
+	"github.com/libopenstorage/openstorage/pkg/correlation"
 )
 
+func init() {
+	correlation.RegisterComponent("auth")
```
lpabon, line 26: do you think this should be `osd/auth` as a pkg to make it more unique and void conflicts with other `auth` packages?

pkg/correlation/README.md
```diff
@@ -0,0 +1,103 @@
...
+- The CSI Driver (/csi)
+- The SDK Server (/api/server/sdk)
+
+### Method 1: Per-package correlation logger
+This method is nice if a package maintainer wants to register a global 
+logging hook at a per-package level. This is nice as you only need to
+create the logger once per package. However, each log line can be quite
+verbose as you must provide WithContext(ctx)
+
+File 1:
```
lpabon, line 17: Suggestion: Always add an empty line before markdown tags

Also you can add the word ```go``` just after the ticks so that github adds Go colors

pkg/correlation/README.md
```diff
@@ -0,0 +1,103 @@
...
+```
+package example
+
+var (
+    clogger := correlation.NewPackageLogger("test")
+)
+	
+func testFunc() {
+    ctx := correlation.NewContext(context.Background(), "source-component")
```
lpabon, line 26: Instead of showing users to use context from `context.Background()` which has no correlation id, have it passed in to `testFunc()`

pkg/correlation/README.md
```diff
@@ -0,0 +1,103 @@
...
+### Method 2: Per-function correlation logger
+This method is great for reducing the amount of count needed per-line, 
+as you do not need to pass in the context at every log line. However,
+you must remember to instantiate the logger at every function.
+
+```
+package example
+
+func testFunc() {
+    ctx := correlation.NewContext(context.Background(), "test_origin")
```
lpabon, line 51: Suggestion

```go
package example

func testFunc(ctx context.Context) {
    ctx = correlation.NewContext(ctx, "example")
    clogger := correlation.NewFunctionLogger(ctx)
    clogger.Info("test info log 1")
    testFuncTwo(ctx)
}
```

Also, what happens if we just pass `ctx` without adding the `NewContext()` call right before it?

pkg/correlation/README.md
```diff
@@ -0,0 +1,103 @@
...
+    clogger := correlation.NewFunctionLogger(ctx)
+
+    clogger.Info("test info log")
+}
+```
+
+### Method 3: Global correlation logging hook
+This is a nice catch-all for covering packages where a
+package-level or function-level logger is not created. However,
+it does not support component metadata. To mark certain packages as
```
lpabon, line 68: ` However, it does not support component metadata` What does this mean? Could an example be provided?

pkg/correlation/README.md
```diff
@@ -0,0 +1,103 @@
...
+it does not support component metadata. To mark certain packages as 
+a component, call `RegisterComponent` from a given package. Each 
+log line in this package will have the registered component added.
+
+In main.go, first we should register the global hook:
+```
+package main
+
+init() {
+    correlation.RegisterGlobalHook()
```
lpabon, line 77: Could some comments be added to explain what the functions are doing?

pkg/correlation/README.md
```diff
@@ -0,0 +1,103 @@
...
+This will help add extra context in our log lines for those who are unfamiliar with our codebase. 
+For example, if a support engineer sees many log line failures with "csi-driver" component,
+they'll know who to contact regarding that failure.
+
+The following example shows how a helper package can be registered as a component:
+```
+package example
+
+func init() {
+    correlation.RegisterComponent("example-package")
```
lpabon, line 97: Can this method work with `clogger := correlation.NewFunctionLogger(ctx)` in method 2?

pkg/correlation/context.go
```diff
@@ -0,0 +1,48 @@
...
+	ID string
+
+	// Origin is the starting point for this request.
+	// Examples may include any of the following:
+	// pxctl, pxc, kubernetes, CSI, SDK, etc
+	Origin Component
+}
+
+// NewContext returns a new correlation context object
+func NewContext(ctx context.Context, origin Component) context.Context {
```
lpabon, line 38: Suggestion: Instead of calling it `NewContext()` should it be called `WithCorrelation()` or something like that to match context APIs?

pkg/correlation/hook.go
```diff
@@ -0,0 +1,157 @@
+package correlation
```
lpabon, line 1: missing APACHE 2 LIC header in all new files. Need to be added.

pkg/correlation/interceptor.go
```diff
@@ -0,0 +1,25 @@
...
+
+// UnaryInterceptor creates a gRPC interceptor for adding
+// correlation ID to each request
+func (ci *ContextInterceptor) ContextUnaryInterceptor(
+	ctx context.Context,
+	req interface{},
+	info *grpc.UnaryServerInfo,
+	handler grpc.UnaryHandler,
+) (interface{}, error) {
+	ctx = NewContext(ctx, ci.Origin)
```
lpabon, line 22: Should there should be a bug-check here if Origin is not set or something.

lpabon (commented):
I see that it is now merged because GitHub had issues during my review, but these changes will require a new PR

pkg/auth/auth.go
ggriffiths: Yep makes sense

pkg/correlation/README.md
ggriffiths: I think the existing example is sufficient. `testFunc` creates the context from no context at all. 

testFuncTwo logs will include this context.

pkg/correlation/README.md
ggriffiths: this is old, will update

pkg/correlation/README.md
ggriffiths: yeah, will add this

pkg/correlation/README.md
ggriffiths: will add

pkg/correlation/context.go
ggriffiths: I considered this, maybe I'll make it `WithCorrelationContext`

pkg/correlation/hook.go
ggriffiths: adding

pkg/correlation/interceptor.go
ggriffiths: I think we're fine without that for now to keep the API clean. I don't want to fail a request because our logging is having issues.

pkg/correlation/hook.go
```diff
@@ -0,0 +1,157 @@
...
+	// LogFieldComponent represents a logging field for control plane component.
+	// This is typically set per-package and allows you see which component
+	// the log originated from.
+	LogFieldComponent = "component"
+)
+
+// Levels describes which levels this logrus hook
+// should run with.
+func (lh *LogHook) Levels() []logrus.Level {
+	return []logrus.Level{logrus.InfoLevel}
```
adityadani, line 43: Why not add warn and error levels as well ?

pkg/correlation/hook.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// Fire is used to add correlation context info in each log line
+func (lh *LogHook) Fire(entry *logrus.Entry) error {
+	// Default to tne entry context. This is populated
```
adityadani, line 48: nit: the

pkg/correlation/hook.go
```diff
@@ -0,0 +1,157 @@
...
+// i.e. /go/src/github.com/libopenstorage/openstorage/pkg/correlation
+// will return openstorage/pkg/correlation
+func getLocalPackage(dir string) string {
+	parts := strings.Split(dir, "/")
+	var githubIndex int
+	for i, d := range parts {
+		// This will work for both imported versions of this
+		// as well as in openstorage itself, since we'll always
+		// keep track of the last index for "github" in the path
+		if d == "github.com" {
```
adityadani, line 145: a path could contain something other than github.com like k8s.io
In that case we will end up logging a huge package path starting from the 2nd index.

adityadani: Having said that this won't be a problem for openstorage or porx, but could be a problem for someone using this library.

PR #1918, opened 2021-08-11
[context] Add context to mount/unmount APIs

ggriffiths:

* Need to pass correlation down to lower layers
* Updates Volume interface to include a context.Context for Mount/Unmount

Closes #

* Mostly API changes and a few fixes to the correlation pkg for logging.


ggriffiths:
Ready for review, but do not merge - needs to be timed w/ porx branch.

ggriffiths:
Need to fix UTs, WIP still.

pault84 (approved):
LGTM

pkg/correlation/context.go
```diff
@@ -61,3 +65,10 @@ func WithCorrelationContext(ctx context.Context, origin Component) context.Conte
 
 	return ctx
 }
+
+// TODO is an alias for context.TODO(), specifically
+// for keeping track of areas where we might want to add
+// the correlation context.
+func TODO() context.Context {
```
nrevanna, line 72: nit: Why have a TODO function in correlation package. Why not use context.TODO() where you have correlation.TODO()?

nrevanna (approved):
LGTM

pkg/correlation/context.go
ggriffiths: Good question, we have this to mark spots where we MAY want to add the correlation context in the future. context.TODO() is nice for a general TODO, where correlation.TODO() is for us to easily search the whole codebase for spots to add correlation context.

ggriffiths: Others writing code in porx may use context.TODO() which would make it harder to remember where we may want to add the correlation context.

nrevanna: In that case would it be better with something like a correlation.GetContext()? Then you need to change only the implementation?

ggriffiths: Not necessarily, context.Context objects are typically generated once at the beginning of a call stack and passed down. So it's more likely that it'll be added as a function parameter and we'll need to replace `context.TODO()` with the context object parameter.

I.e.
```go
func someFunc() {
  logrus.WithContext(correlation.TODO()).Infof("...")
}
```

refactored to 
```go
func someFunc(ctx context.Context) {
  logrus.WithContext(ctx).Infof("...")
}
```

nrevanna: Makes sense. Thanks.

PR #1929, opened 2021-09-10
PWX-19388 Migrate go vendor to go mod

dahuang-purestorage:
Signed off by Daniel Huang (dahuang@purestorage.com)

Migrate go vendor to go mod. Some packages requires go mod to be upgraded.

Closes #

We shouldn't merge this until porx is upgraded to kube api version 1.2 as well.


ggriffiths (approved):

Makefile
```diff
@@ -408,8 +409,9 @@ sdk-check-version:
 	go run tools/sdkver/sdkver.go --check-major=0 --check-patch=0
 
 mockgen:
-	go get github.com/golang/mock/gomock
-	go get github.com/golang/mock/mockgen
+	GO111MODULE=off go get github.com/golang/mock/gomock
+	GO111MODULE=off go get github.com/golang/mock/mockgen
+	cp -r /go/src/github.com/golang/mock ./vendor/github.com/golang/
```
piyush-nimbalkar, line 414: Instead of assuming a GOROOT directory, we can do something like https://github.com/libopenstorage/operator/blob/master/tools.go#L8 to make sure go mod pulls the mockgen correctly?

Makefile
```diff
@@ -93,57 +93,57 @@ all: build $(OSDSANITY)
...
-	go get -d -v -t $(PKGS)
+	GO111MODULE=off go get -d -v -t $(PKGS)
 
 update-test-deps:
-	go get -tags "$(TAGS)" -d -v -t -u -f $(PKGS)
+	GO111MODULE=off go get -tags "$(TAGS)" -d -v -t -u -f $(PKGS)
 
 vendor-update:
-	GOOS=linux GOARCH=amd64 go get -tags "daemon btrfs_noversion have_btrfs have_chainfs" -d -v -t -u -f $(PKGS)
+	GOOS=linux GOARCH=amd64 GO111MODULE=off go get -tags "daemon btrfs_noversion have_btrfs have_chainfs" -d -v -t -u -f $(PKGS)
```
piyush-nimbalkar, line 146: Won't we be replacing the `vendor` make targets to use go mod instead of govendor in this PR? It would be easier for people to just run `make vendor`.

dahuang-purestorage: yeah that's a good point. adding the vendor command.

PR #1936, opened 2021-09-20
export pool maintenance status

sulakshm:
Pool Maintenance status is not exported in openstorage api.
This PR extends it so it can be used as node status as is.

Closes #


api/api.proto
```diff
@@ -34,8 +34,9 @@ enum Status {
   STATUS_STORAGE_REBALANCE = 11;
   STATUS_STORAGE_DRIVE_REPLACE = 12;
   STATUS_NOT_IN_QUORUM_NO_STORAGE = 13;
+  STATUS_POOLMAINTENANCE = 14;
   // Add statuses before MAX and update the number for MAX
-  STATUS_MAX = 14;
+  STATUS_MAX = 15;
```
lpabon, line 39: Hmm, this is not part of this PR, but now I see that `STATUS_MAX` does not work as a protocol buffer value since it keeps changing. That means that if another version of the gRPC/REST command sends `STATUS_MAX at version 14` it will be interpreted as a `STATUS_POOLMAINTENANCE`. We have to create a doc change to inform callers to NOT use `STATUS_MAX`

lpabon (approved):

api/api.proto
sulakshm: Although your assumption is correct, STATUS_MAX is defined as a sentinel to sanity check status values and carry no significance otherwise. Secondly, this is a SDK library that is statically linked to the client binary, so if there are clients outside that uses old library then it does not know POOLMAINTENANCE and should correctly report that based on STATUS_MAX.

PR #1939, opened 2021-09-28
PWX-21341: add failover delay option for sharedv4 volumes

pureneelesh:
Added a new spec for options common to both sharedv4 and
sharedv4 service volumes. Added "failover delay" enum to that spec.

PWX-21341

We can discuss any suggestions for the naming changes.


lpabon (approved):
👍🏼 nice job

api/spec/spec_handler.go
```diff
@@ -472,7 +473,15 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
...
-
+		case api.SpecSharedv4FailoverDelay:
+			if spec.Sharedv4Spec == nil {
+				spec.Sharedv4Spec = &api.Sharedv4Spec{}
+			}
+			if v == "short" {
+				spec.Sharedv4Spec.FailoverDelay = api.Sharedv4FailoverDelay_SHORT
+			} else if v == "normal" {
+				spec.Sharedv4Spec.FailoverDelay = api.Sharedv4FailoverDelay_NORMAL
+			}
```
adityadani, line 484: Add an else here to error out for all unknown strings.

Also add the short and normal strings as constants

api/api.proto
```diff
@@ -944,6 +973,7 @@ message Volume {
   bool in_trashcan = 31;
   // DerivedIoProfile the IO profile determined from the pattern
   IoProfile derived_io_profile = 32;
+  // TODO should we add sharedv4_spec here?
```
adityadani, line 976: It is better to add it here. If an input spec is not provided we can set the default value in the Volume object.

adityadani (approved):
lgtm

api/api.proto
pureneelesh: In that case, it makes sense to not add it because the default depends on the type of volume (sharedv4 vs sharedv4 service). If the value is not specified, we use the appropriate default.

api/api.go
```diff
@@ -80,27 +80,31 @@ const (
...
+	StoragePolicy                   = "storagepolicy"
+	SpecCowOnDemand                 = "cow_ondemand"
+	SpecDirectIo                    = "direct_io"
+	SpecScanPolicyTrigger           = "scan_policy_trigger"
+	SpecScanPolicyAction            = "scan_policy_action"
+	SpecProxyWrite                  = "proxy_write"
+	SpecSharedv4ServiceType         = "sharedv4_svc_type"
+	SpecSharedv4ServiceName         = "sharedv4_svc_name"
+	SpecSharedv4FailoverDelay       = "sharedv4_failover_delay"
+	SpecSharedv4FailoverDelayNormal = "normal"
```
harsh-px, line 94: Instead of `normal` and `short`, we can use `conservative` and `aggressive` instead.

When selecting the two, the end user is picking a strategy instead of the actual duration.

@adityadani wdyk?

api/api.proto
adityadani: So shouldn't we be setting that value here? Essentially when someone inspects the volume they will know what strategy is being used by looking under this Volume struct.

api/api.go
adityadani: In that case, we have the following options:
Option 1: A bit verbose but self explanatory
```
sharedv4_failover_delay_strategy: conservative | aggressive
```
Option 2: With the shorter version of key it is not explicit since they key `sharedv4_failover_delay` implies that you are setting some kind of time period
```
sharedv4_failover_delay: conservative | aggressive
```
Option 3: The original one
```
sharedv4_failover_delay: short | normal
```

I would vote for Option 1.

pureneelesh: One advantage of using "strategy" is that we could hide more settings behind it in the future. If that is the intention behind using the strategy, we should keep the name of the option generic and not include "delay" in the name. Also, we should keep "normal" as one of the two values so that we can add "conservative" as the third option in the future.

So maybe something like:
```
sharedv4_failover_strategy: aggressive | normal
```

I would still prefer the "sharedv4_failover_delay" option since it is crisp and clear and the values are shorter. We can still add other options as separate parameters in the future. But I am fine with going with the strategy option if you guys prefer that.

harsh-px: Let's go ahead with `sharedv4_failover_strategy: aggressive | normal`

api/api.proto
pureneelesh: I was thinking that we want to leave it empty to indicate that the user does not have a preference and wants to use the appropriate system default. If the volume type is changed from sharedv4 -> sharedv4 service, a different default value is used automatically.

PR #1940, opened 2021-09-29
PWX-21438 Added api for auto fs trim status

nikhil-bhupale:
Added api and messages for auto fs trim status cli


api/api.proto
```diff
@@ -4772,6 +4779,19 @@ message SdkFilesystemTrimStatusResponse{
   string message = 2;
 }
 
+// SdkAutoFSTrimStatusRequest defines a request to start a background autofs trim operation
```
vjs-px, line 4782: update comment

nikhil-bhupale: Fixed

vjs-px (approved):

PR #1942, opened 2021-10-01
Spec handler: Add an API to generate RestoreVolumeSpec from opaque options.

adityadani:

- RestoreVolumeSpec is used in CloudBackupRestore API and defines how a volume
  should be created on restore from a cloud backup.
- The new RestoreSpecFromOpts follows the same pattern as the other APIs in spec
  handler package, wherein it parses opaque options and converts them into a
  well defined spec.
- These opaque options could be the parameters provided in Storage Class.

Closes #


api/spec/spec_handler.go
```diff
@@ -847,3 +859,353 @@ func (d *specHandler) SpecFromString(
...
+	*api.RestoreVolumeSpec,
+	*api.VolumeLocator,
+	error,
+) {
+	spec := &api.RestoreVolumeSpec{}
+	locator := &api.VolumeLocator{}
+	locator.VolumeLabels = make(map[string]string)
+	for k, v := range opts {
+		switch k {
+		case api.SpecNodes:
```
adityadani, line 875: This function is mostly a copy paste of UpdateSpecFromOpts. I had to remove a few opt handling since those params are not supported by RestoreVolumeSpec

ram-infrac (approved):

ram-infrac:
UT's failing for git validation, please fix it before merging

api/spec/spec_handler.go
```diff
@@ -847,3 +859,353 @@ func (d *specHandler) SpecFromString(
...
+			}
+			spec.ReplicaSet = &api.ReplicaSet{Nodes: nodeList}
+		case api.SpecQueueDepth:
+			if queueDepth, err := strconv.ParseInt(v, 10, 64); err != nil {
+				return nil, nil, err
+			} else {
+				spec.QueueDepth = uint32(queueDepth)
+			}
+		case api.SpecHaLevel:
+			haLevel, _ := strconv.ParseInt(v, 10, 64)
```
talakad, line 891: no need to handle error?

api/spec/spec_handler.go
```diff
@@ -847,3 +859,353 @@ func (d *specHandler) SpecFromString(
...
+			if proxyWrite, err := strconv.ParseBool(v); err != nil {
+				return nil, nil, err
+			} else {
+				spec.ProxyWrite = convertToRestoreParamBoolType(proxyWrite)
+			}
+		case api.SpecFastpath:
+			if fastpath, err := strconv.ParseBool(v); err != nil {
+				return nil, nil, err
+			} else {
+				spec.FpPreference = convertToRestoreParamBoolType(fastpath)
```
talakad, line 1135: If an option is not specified in storage-class, (specifically the boolean values), will it make sense to have the  default value be to set to RestoreParamType_BKUPSRC?

PR #1957, opened 2021-11-12
PWX-21966: Increase retry interval and timeout

nrevanna:
250 ms of retry interval is too frequent for k8s api calls. This is
      resulting in us hitting the k8s rate limiting and a bunch of
      issues following that. This patch decreases the frequency and also
      the timeout.

  250 ms of retry interval is too frequent for k8s api calls. This is
      resulting in us hitting the k8s rate limiting and a bunch of
      issues following that. This patch decreases the frequency and also
      the timeout.

Closes # PWX-21966


harsh-px (approved):

nrevanna:
Will wait for Travis to complete and merge this.

adityadani (approved):

PR #1969, opened 2022-01-03
NEWSTACK-5: FA DirectAccess and PSO feature parity initial commit

david-enli:

Closes #


api/server/sdk/volume_ops.go
```diff
@@ -155,6 +155,16 @@ func (s *VolumeServer) create(
 				err.Error())
 		}
 
+		if spec.IsPureVolume() {
```
ggriffiths, line 158: was it not possible to do this at the filter or porx driver layer? We typically try to avoid driver-specific things in the SDK layer

ggriffiths:
You'll also need to bump the minor version in api.proto:
https://github.com/libopenstorage/openstorage/blob/master/api/api.proto#L5028

and add release notes for these new API fields for your new version (v0.130.0)
https://github.com/libopenstorage/openstorage/blob/master/SDK_CHANGELOG.md

api/api.go
```diff
@@ -63,6 +63,8 @@ const (
 	SpecExportOptionsEmpty   = "empty_export_options"
 	SpecMountOptions         = "mount_options"
 	SpecCSIMountOptions      = "csi_mount_options"
+	SpecCSIRawBlock			 = "fadirectRawBlock"
```
ggriffiths, line 66: most spec options appear to be snake case - `fa_direct_raw_block`, let's stick to that convention

api/api.go
```diff
@@ -63,6 +63,8 @@ const (
...
+	SpecCSIFsType			 = "fsType"
```
ggriffiths, line 67: we already have a `fs` spec - see above `	SpecFilesystem           = "fs"`

Can we reuse that one?

api/api.go
ggriffiths: this variable is named `SpecCSIRawBlock`, but the value seems to be specific to FA direct integration. 

Is the goal here to enable FA direct csi raw block integration via a SC parameter? If so, let's just name the spec name & value something like `SpecFADirectRawBlock`.

david-enli: will do, yea this is just to pass the raw_block flag through driver options down to the filter level so NSM knows its a raw block

api/server/sdk/volume_ops.go
david-enli: I'll think about that a little more, I think we went through the code together and thats how we concluded, may need to revisit.

api/api.go
adityadani: Yes we should reuse the `fs` parameter. Also the other parameter `CSIRawBlock` is not really required. The CSI driver in openstorage already handles Raw Block Devices. When RawBlock PVCs are used the spec.Format is set to `api.FSType_FS_TYPE_NONE`

api/server/sdk/volume_ops.go
ggriffiths: Sorry yeah I think we agreed that too. Lets try in the filter layer first as it's more ideal. We try to handle things like this in the driver layer.

david-enli: so I think the reason I added this check is specifically for restoring. for restoring, we are creating a new PVC from a snapshot, without this check, we go directly to create a new snapshot. I could add a check in creating snapshot and direct that to restore, but I dont know if that makes more sense. what are your thoughts?

david-enli:
closing this PR, seems lots of changes aren't needed after all.

PR #1971, opened 2022-01-04
Added a filename to our diags collection request

pault84:
Signed-off-by: Paul Theunis <ptheunis@purestorage.com>

With the async diags functionality it was impossible to test file creation without a filename.
This PR added one.

I have also changed the makefile to make sure dockerbuild uses network=host because on KIF machines the network was not propagated


pault84:
not sure why the build is failing I see 2 fatal not a git repo but the build continues fine.

api/api.proto
```diff
@@ -3755,6 +3755,8 @@ message SdkDiagsCollectRequest {
   int64 timeout_mins = 5;
   // Live is an optional flag if true will collect live cores from running processes of the driver
   bool live = 6;
+  // Filename is an optional flag only to be used for testing purposes.
+  string filename = 7;
```
ggriffiths, line 3759: Need to bump API minor version, that's likely why this is failing

ggriffiths (approved):
lgtm

PR #1972, opened 2022-01-06
WIP - CSI: Decouple CSI Driver from Volume Driver

ggriffiths:

Decouple CSI Driver from Volume Driver

Closes #


ggriffiths:
Does not work well, hard to tell why the job failed. It just exits without a message.

ggriffiths:
PR Feedback from Luis: Use app issuer instead of system issuer. Will make this change.

ggriffiths:
PR on hold, will re-pickup in a few weeks

PR #1973, opened 2022-01-06
Terminate travis job when command fails

ggriffiths:

Closes #


ggriffiths:
Does not work well, hard to see why the job is failing. Just exits w/o a message.

PR #1977, opened 2022-01-25
NEWSTACK-5: Add a flag for FA direct access volume in node server

david-enli:
We need to pass in a CSI file system flag to the filter chain so NSM filter can use it for FA directAccess volume
Closes #
NEWSTACK-5


ggriffiths (approved):
lgtm, thanks!

PR #2000, opened 2022-02-23
[CSI] Add extra idempotency check for resizing

ggriffiths:

There appear to be some cases where resizer things a resize is needed, so a request comes into PX to resize. However, since no change is needed, we return an error in the proto driver. This fixes these cases at the CSI level.

Closes #

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -607,6 +607,24 @@ func (s *OsdCsiServer) ControllerExpandVolume(
 	// should be issued. Otherwise, no operation should occur.
 	volumes := api.NewOpenStorageVolumeClient(conn)
 
+	// Check for idempotency. There are cases where we return an error in the proto
+	// driver if the volume update request has no changes, so we return early here instead.
+	inspectResp, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{
+		VolumeId: req.GetVolumeId(),
+	})
+	if err != nil {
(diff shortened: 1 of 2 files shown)


csi/controller.go
```diff
@@ -607,6 +607,24 @@ func (s *OsdCsiServer) ControllerExpandVolume(
...
+	// Check for idempotency. There are cases where we return an error in the proto
+	// driver if the volume update request has no changes, so we return early here instead.
+	inspectResp, err := volumes.Inspect(ctx, &api.SdkVolumeInspectRequest{
+		VolumeId: req.GetVolumeId(),
+	})
+	if err != nil {
+		if err == kvdb.ErrNotFound {
+			return nil, status.Errorf(codes.NotFound, "Volume id %s not found", req.GetVolumeId())
+		}
+		return nil, status.Errorf(codes.Internal, "Failed to get volume: %v", err)
```
harsh-px, line 619: nit: include volume ID in the error message.

harsh-px (commented):
were you able to repro the issue?

ggriffiths:
No, but it's clear what was happening - PX volume had correct size (32gb) but PVC status was 8gb. Somehow the request went through but it didn't reach resizer/k8s. The logs did not help for this case.

In a future occurrence of this, CSI will see the size is correct and return early.

harsh-px (approved):

PR #2002, opened 2022-03-03
PWX-23061: Add new funcs to allow for dumping profiles to specific fi…

jrivera-px:
…lenames.

We had an issue where the stats profile dumps where being cleaned up and lost.  These dumps help with debugging issues.   This change is to allow stats thread in PX to use its own filename when producing profile dumps.    Doing this will allow us to provide separate cleanup method for these files and we can retain them separately and help prevent losing them during cleanup.

PWX-23061

This is part of the fix and another PR will be submitted for the portworx/porx side.

### DIFF
--- a/pkg/dbg/profile_dump.go
+++ b/pkg/dbg/profile_dump.go
@@ -12,7 +12,8 @@ import (
 )
 
 const (
-	path     = "/var/cores/"
+	// Path defailt location for debug/profile output
+	Path     = "/var/cores/"
 	fnameFmt = "2006-01-02T15:04:05.999999-0700MST"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/dbg/profile_dump.go
```diff
@@ -41,19 +62,10 @@ func DumpGoMemoryTrace() {
...
-	f, err := os.Create(path + GetHostNamePrefix() + "-" + GetTimeStamp() + ".heap")
-	if err != nil {
-		logrus.Errorf("could not create memory profile: %v", err)
-		return
-	}
-	defer f.Close()
-	if err := pprof.WriteHeapProfile(f); err != nil {
-		logrus.Errorf("could not write memory profile: %v", err)
-	}
+	HeapDump(Path + GetHostNamePrefix() + "-" + GetTimeStamp() + ".heap")
```
zoxpx, line 70: These function names are a bit confusing now -- e.g. _"what's the difference between DumpGoProfile and GoProfileDump?"_, and _"how about HeapDump and DumpHeap?"_

I'd suggest to either
a. name them the same but change capitalization (e.g. `func DumpHeap()` vs. unexported `func dumpHeap(dest)`), or
b. if we want to keep them exported, perhaps have `func DumpGoProfile()` vs. `func DumpGoProfileTo()`

zoxpx (approved):
I think this looks OK - it's just the function names are bit confusing (suggest renaming them before the commit)

pkg/dbg/profile_dump.go
jrivera-px: There is no difference don't want to change DumpGoProfile its used in other places in the code.  The change was done to allow changes only to the profiles generated by stats thread in PX.  So wanted to add an exported layer which can be used with a specified filename.  That is all.     I will change the names to `GoProfileDump` to `DumpGoProfileTo` and `HeapDump` to `HeapDumpTo`.

PR #2003, opened 2022-03-04
[correlation] Update volume Create/Delete to use context.Context

ggriffiths:

Adds context.Context to openstorage volume driver create/delete

### DIFF
--- a/api/client/volume/client.go
+++ b/api/client/volume/client.go
@@ -131,7 +131,7 @@ func (v *volumeClient) GraphDriverDiffSize(id string, parent string) (int, error
 
 // Create a new Vol for the specific volume spev.c.
 // It returns a system generated VolumeID that uniquely identifies the volume
-func (v *volumeClient) Create(locator *api.VolumeLocator, source *api.Source,
+func (v *volumeClient) Create(ctx context.Context, locator *api.VolumeLocator, source *api.Source,
 	spec *api.VolumeSpec) (string, error) {
 	response := &api.VolumeCreateResponse{}
 	request := &api.VolumeCreateRequest{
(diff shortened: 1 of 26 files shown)


api/client/volume/client.go
```diff
@@ -131,7 +131,7 @@ func (v *volumeClient) GraphDriverDiffSize(id string, parent string) (int, error
 
 // Create a new Vol for the specific volume spev.c.
 // It returns a system generated VolumeID that uniquely identifies the volume
-func (v *volumeClient) Create(locator *api.VolumeLocator, source *api.Source,
+func (v *volumeClient) Create(ctx context.Context, locator *api.VolumeLocator, source *api.Source,
```
nrevanna, line 134: Just a thought - In terms of number of files that gets changed. Will it be a better idea to to make Context as part of volumeClient instead of changing the interface to make it part of the function args?

nrevanna: Taking volumeClient as an example asking generally if it's a better idea to make it part of the interface object.

ggriffiths: volumeClient is actually an implementation of the volume interface, so that results in the following error:
```
        *volumeClient does not implement volume.VolumeDriver (wrong type for Create method)
                have Create(*api.VolumeLocator, *api.Source, *api.VolumeSpec) (string, error)
                want Create(context.Context, *api.VolumeLocator, *api.Source, *api.VolumeSpec) (string, error)
```

nrevanna (approved):
Had a chat with Grant.
LGTM

api/client/volume/client.go
ggriffiths: Discussed offline, this would help for volume client, but passing the correlation tracing info between proto driver, filters, etc and different packages would be tough.

PR #2005, opened 2022-03-09
PWX-23088: expose gossip intervals

pureneelesh:
Added a method to allow consumers to retrieve the configured values
of the gossip settings.

PWX-23088

will remove the replace directive once the gossip PR is merged

### DIFF
--- a/api/client/cluster/client.go
+++ b/api/client/cluster/client.go
@@ -318,6 +318,11 @@ func (c *clusterClient) GetGossipState() *cluster.ClusterState {
 	return status
 }
 
+func (c *clusterClient) GetGossipIntervals() types.GossipIntervals {
+	// not implemented
+	return types.GossipIntervals{}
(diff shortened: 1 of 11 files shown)


api/client/cluster/client.go
```diff
@@ -318,6 +318,11 @@ func (c *clusterClient) GetGossipState() *cluster.ClusterState {
 	return status
 }
 
+func (c *clusterClient) GetGossipIntervals() types.GossipIntervals {
+	// TODO: do we need to implement this?
```
adityadani, line 322: No we don't need to implement it since there is no need to expose this info over the REST client API.

adityadani (approved):

ggriffiths (approved):
lgtm, just missing a commit signature on your vendor commit:
```
 FAIL - does not have a valid DCO
```

PR #2008, opened 2022-03-21
PWX-23028 & PWX-23107: Changes Pure DA checks for RWX block

Pure-AdamuKaapan:
Signed-off-by: Adam Krpan <akrpan@purestorage.com>

Currently we cannot create Pure FA RWX volumes (which should be supported) due to this check. This will check if the proxy spec is a pure type: if it is, it will go through a different set of checks for which volumes are or are not allowed.

PWX-23028 & PWX-23107

This has now been tested with all the various permutations of RWO/RWX, PX or Pure DA volumes, and block/file access modes. All performed as expected (e.g. PX raw RWX is blocked, FA filesystem RWX is blocked, everything else is allowed)

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -379,12 +379,56 @@ func validateCreateVolumeCapabilities(caps []*csi.VolumeCapability) error {
 		if cap.GetBlock() != nil {
 			block = true
 		}
+	}
 
-		if block && shared {
-			return status.Errorf(
-				codes.InvalidArgument,
-				"Shared raw block volumes are not supported")
+	if block && shared {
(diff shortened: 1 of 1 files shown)


ggriffiths (changes requested):
Looks like you have a compile issue in controller.go:
https://app.travis-ci.com/github/libopenstorage/openstorage/jobs/564145345#L253

Also, we'll need to remove this test in controller_test.go:
https://github.com/libopenstorage/openstorage/blob/master/csi/controller_test.go#L2122

ggriffiths (approved):
Looks good!

PR #2019, opened 2022-03-28
IPv6 support - Use net.JoinHostPort to concat ip & port. Add '[]' to …

jrivera-px:
…host in nfs address if necessary.

Use Go pkg to handle concat IP & port which handles IPv6 addresses and NFS mounts with IPv6 addresses need to have IP enclosed in '[]'
 
Closes #

### DIFF
--- a/pkg/mount/nfs.go
+++ b/pkg/mount/nfs.go
@@ -81,8 +81,12 @@ func (m *nfsMounter) normalizeSource(info *mount.Info, host string) {
 	if info.Fstype != "nfs" {
 		return
 	}
+
 	s := strings.Split(info.Source, ":")
 	if len(s) == 2 && len(s[0]) == 0 {
+		if strings.Contains(host, ":") && !strings.Contains(host, "[") {
+			host = fmt.Sprintf("[%s]", host)
+		}
 		info.Source = host + info.Source
 	}
 }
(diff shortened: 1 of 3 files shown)


pkg/mount/nfs.go
```diff
@@ -81,8 +81,12 @@ func (m *nfsMounter) normalizeSource(info *mount.Info, host string) {
 	if info.Fstype != "nfs" {
 		return
 	}
+
 	s := strings.Split(info.Source, ":")
 	if len(s) == 2 && len(s[0]) == 0 {
+		if strings.Contains(host, ":") && !strings.Contains(host, "[") {
+			host = fmt.Sprintf("[%s]", host)
```
ggriffiths, line 88: will adding the brackets here cause issues for existing NFS mounts? Thinking of the case where we upgrade PX and this change is added. Will PX adjust these mountpaths during a PX upgrade?

jrivera-px: No if the host has contains an ':' then it will be an IPv6 IP address.   If not then it will be dns entry and will not need it.

ggriffiths: Ah, makes sense :+1:

ggriffiths (approved):

PR #2021, opened 2022-04-01
PWX-23408: Adds full_vol_name to Pure DA proxy specs

Pure-AdamuKaapan:
Signed-off-by: Adam Krpan <akrpan@purestorage.com>

To support PSO to Portworx migration, we need a way to pass in a full volume name. This will both denote that the volume being "created" actually already exists, and shouldn't be created and just added to the backend database instead.

PWX-23408


ggriffiths (approved):
Lgtm!

PR #2022, opened 2022-04-05
PWX-23290: handle spec updates with stale config

sulakshm:

When volume spec updates are requested, there is a spec reconciliation based on the current state of the volume.
This is done to validate policy settings.
For some volume update operations the spec gets updated only upon action from px-storage and may take time to complete action. Ex. HA Update on a volume needs action from px-storage before the HA level change gets reflected in  spec.
So, if there is a subsequent volume update, it is possible for the reconciliation action part of update to pick stale config,
and the final action may be inconsistent after.

To avoid this, whenever volume update is applied, only pass the relevant section of the spec that needs update, mask 
rest of the specs that have not been updated. 

This window of stale config is open only for defered spec updated and long running actions to complete the update operations. Boolean based state are mostly instantaneous updates to the spec and does not seem to have this problem.

Closes # PWX-23290
https://portworx.atlassian.net/browse/PWX-23290

The changes have been tested in 2.10.0-ea branch where the issue was reported.

### DIFF
--- a/api/server/sdk/volume_ops.go
+++ b/api/server/sdk/volume_ops.go
@@ -19,6 +19,7 @@ package sdk
 import (
 	"context"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


api/server/sdk/volume_ops.go
```diff
@@ -609,6 +677,11 @@ func (s *VolumeServer) Update(
 	if err != nil {
 		return nil, err
 	}
+
+	// avoid side effect while applying with stale config by masking
+	// other parts of the spec.
+	maskUnModified(updatedSpec, req.GetSpec())
+
 	// Send to driver
 	if err := s.driver(ctx).Set(req.GetVolumeId(), locator, updatedSpec); err != nil {
```
sulakshm, line 686: This ends up calling porx repo: storage/driver/volume/proto_driver.go
https://github.com/portworx/porx/blob/e0bd8dab2f2f684c4b06f7e55ef862ea63d02f7a/storage/driver/volume/proto_driver.go#L5710

sharma-tapas (approved):

prabirpaul:
This is definitely better than what we have now.
What about cases like back to back ha updates where the 2nd update is issued right before the 1st one completes. We would still have a problem there rt?

sulakshm:
> This is definitely better than what we have now. What about cases like back to back ha updates where the 2nd update is issued right before the 1st one completes. We would still have a problem there rt?

Ha Operations do update the spec to hand off request from px to px-storage, but it is a different part of the spec. 
There is a Recovery section that gets exchanged between px and px-storage. 
So any new Ha Operation always refer to below part of the spec. This portion will always get reflected for the ongoing Ha.
So either the new Ha operation is serialized behind or rejected. 

```
  "Recovery": {
    "ID": "437652889672355328",
    "State": 0,
    "AttachedOn": -1,
    "AggregationType": 1,
    "ReplicationSets": [
      {
        "ID": 0,
        "CreateSet": [
          0,
          1
        ],
        "CreateSetPoolIds": [
          0,
          0
        ],
        "ReplicaSetCurr": [
          0,
          1
        ],
        "ReplicaSetNew": null,
        "ReplicaSetNewPoolIds": null,
        "RemoveSet": [],
        "ReAddSet": [],
        "ReAddSetPoolIds": [],
        "ReplicaSetNewSourceNodes": null
      }
    ],
    "AbortOnError": false,
    "SnapRetry": false
  },
```

ggriffiths (changes requested):
You'll need to update the CSI tests based on these changes:

You can sometimes `gomock.Any()` as an expected argument to speed this up.
```
Unexpected call to *mock.MockVolumeDriver.Set([myid <nil> size:49392123904  snapshot_interval:4294967295]) at /go/src/github.com/libopenstorage/openstorage/api/server/sdk/volume_ops.go:701 because: 
expected call at /go/src/github.com/libopenstorage/openstorage/csi/controller_test.go:2678 doesn't match the argument at index 2.
Got: size:49392123904  snapshot_interval:4294967295 (*api.VolumeSpec)
Want: is equal to size:49392123904  snapshot_interval:4294967295 (*api.VolumeSpec)
panic: MOCK TEST FATAL FAILURE
goroutine 1571 [running]:
github.com/kubernetes-csi/csi-test/utils.(*SafeGoroutineTester).Fatalf(0x2d146c0, 0x1d72665, 0x2e, 0xc0003ac1e0, 0x5, 0x5)
	/go/src/github.com/libopenstorage/openstorage/vendor/github.com/kubernetes-csi/csi-test/utils/safegoroutinetester.go:39 +0xd1
```

PR #2026, opened 2022-04-07
[Auth] Remove failed token and role from audit logs

ggriffiths:

Removes invalid token from the audit logs

n/a

n/a

### DIFF
--- a/api/server/sdk/server_interceptors.go
+++ b/api/server/sdk/server_interceptors.go
@@ -99,7 +99,7 @@ func (s *sdkGrpcServer) auth(ctx context.Context) (context.Context, error) {
 	// Determine issuer
 	issuer, err := auth.TokenIssuer(token)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


api/server/sdk/server_interceptors.go
```diff
@@ -99,7 +99,7 @@ func (s *sdkGrpcServer) auth(ctx context.Context) (context.Context, error) {
 	// Determine issuer
 	issuer, err := auth.TokenIssuer(token)
 	if err != nil {
-		return nil, auditLogWarningf(codes.Unauthenticated, "Unable to obtain token issuer from authorization token: %v", err)
```
trierra, line 102: why are you removing the error value here?

trierra (approved):

api/server/sdk/server_interceptors.go
ggriffiths: This was just throwing the error into the logs, which could have a token in it.

PR #2034, opened 2022-04-13
IPv6 support - Use net.JoinHostPort to concat ip & port. Add '[]' to …

jrivera-px:
…host in nfs address if necessary.

Use Go pkg to handle concat IP & port which handles IPv6 addresses and NFS mounts with IPv6 addresses need to have IP enclosed in '[]'
Closes #

These changes are already in master.

### DIFF
--- a/cluster/manager/manager.go
+++ b/cluster/manager/manager.go
@@ -438,7 +438,7 @@ func (c *ClusterManager) getNonDecommisionedPeers(
 			continue
 		}
 		peers[types.NodeId(nodeEntry.Id)] = types.NodeUpdate{
-			Addr:          nodeEntry.DataIp + ":" + c.gossipPort,
+			Addr:          net.JoinHostPort(nodeEntry.DataIp, c.gossipPort),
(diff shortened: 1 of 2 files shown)


ggriffiths (approved):

cluster/manager/manager.go
```diff
@@ -1482,8 +1482,12 @@ func (c *ClusterManager) StartWithConfiguration(
 		QuorumTimeout:    quorumTimeout,
 		SuspicionMult:    types.DEFAULT_SUSPICION_MULTIPLIER,
 	}
+
+	nodeIp := net.JoinHostPort(c.selfNode.DataIp, c.gossipPort)
+	logrus.Infoln("Init gossip: ", nodeIp)
+
 	c.gossip = gossip.New(
-		c.selfNode.DataIp+":"+c.gossipPort,
+		nodeIp,
```
zoxpx, line 1490: nit -- why are they called "node IPs"?   (they are "host+port" combo's -- they should be called "endpoints", or maybe "hostports"?)

PR #2045, opened 2022-04-26
PWX-23530: Disables Pure shared RWX volumes temporarily

Pure-AdamuKaapan:
Temporarily blocks Pure shared raw block volumes.


david-enli (approved):
LGTM, thanks

PR #2057, opened 2022-05-24
Update cloudfoundry/gosigar to remove panic when mult cgroup entries …

jrivera-px:
…are found.

Signed-off-by: Jose Rivera <jrivera@purestorage.com>

Update pkgs to pull in latest changes which does not panic when multiple v1 cgroups exist.  

Closes # 
 PWX-23988

Updated from gosigar to v1.3.3 which contains a change to func 'determineControllerMounts' in vendor/github.com/cloudfoundry/gosigar/sigar_linux.go.   This update will bring it inline with what is used in portworx.


harsh-px (approved):
Make sure BVT and Smoke passes.

PR #2068, opened 2022-06-02
PWX-22823: Adding S3 bucket driver backend for Create and Delete API.

pp511:

To implement Create and Delete S3 bucket driver interfaces.

Closes #PWX-22823:

Driver configs will be provided by Cosi once it's mature. We also need to make cosi configurable to be able to set a specific handler on the SDK server.
For now just for API end to end testing purpose, I have added some code (commented) in "main.go". It can be uncommented and credentials can be updated with the AWS account.
Withthe above code uncommented the below gRPC client calls were able to Create and Delete bucket when hitting Port:9106 on the OSD.

`	// Create a bucket
	bucket := api.NewOpenStorageBucketClient(conn)
	b, err := bucket.Create(
		context.Background(),
		&api.BucketCreateRequest{
			Name: "priyanshu511-test21",
		})
	if err != nil {
		gerr, _ := status.FromError(err)
		fmt.Printf("Error Code[%d] Message[%s]\n",
			gerr.Code(), gerr.Message())
		os.Exit(1)
	}
	fmt.Printf("Bucket created with id %s\n", b.GetBucketId())

	// Delete a bucket
	d, err := bucket.Delete(
		context.Background(),
		&api.BucketDeleteRequest{
			BucketId: "priyanshu511-test21",
		})
	if err != nil {
		gerr, _ := status.FromError(err)
		fmt.Printf("Error Code[%d] Message[%s]\n",
			gerr.Code(), gerr.Message())
		os.Exit(1)
	}

	fmt.Printf("Bucket deleted" + d.String())`


bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type S3Driver struct {
+	//svc is IAM S3 service client
+	svc    *s3.S3
+	config *aws.Config
+}
+
+func New(config *aws.Config) *S3Driver {
+	sess, _ := session.NewSession(config)
```
ggriffiths, line 19: Lets check the `err` here and return `nil, err` if this fails.

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+		svc:    svc,
+		config: config,
+	}
+}
+
+func (d *S3Driver) CreateBucket(name string) (string, error) {
+	_, err := d.svc.CreateBucket(&s3.CreateBucketInput{
+		Bucket: aws.String(name),
+		CreateBucketConfiguration: &s3.CreateBucketConfiguration{
+			LocationConstraint: aws.String(*d.config.Region),
```
ggriffiths, line 32: Region will actually need to a parameter for CreateBucket, as a customer could want a BucketClass for one region and another BucketClass for a different region.

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+		},
+	})
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case s3.ErrCodeBucketAlreadyExists:
+				logrus.Error("Bucket name already in use!")
+				panic(err)
```
ggriffiths, line 40: If BucketAlreadyExists, let's make this call Idempotent and return the existing bucket ID/name without an error.

This makes it easier for the controller to know when a bucket already exists or has been created.

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+			case s3.ErrCodeBucketAlreadyOwnedByYou:
+				logrus.Warn("Bucket exists and is owned by the requester")
```
ggriffiths, line 42: Same as above

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+			default:
+				logrus.Error("Unknown error")
```
ggriffiths, line 44: Lets print the error here with a little more verbosity. Something like
`logrus.Errorf("An unknown error has occurred while creating bucket %s: %v", name, err)`

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+	// DeleteBucket is possible only if the Bucket is empty.
+	// Setup BatchDeleteIterator to iterate through a list of objects.
+	iter := s3manager.NewDeleteListIterator(d.svc, &s3.ListObjectsInput{
+		Bucket: aws.String(id),
+	})
+
+	// Traverse iterator deleting each object
+	err := s3manager.NewBatchDeleteWithClient(d.svc).Delete(aws.BackgroundContext(), iter)
+	if err != nil {
+		logrus.Infof("Unable to delete objects from bucket %q  %v", id, err)
```
ggriffiths, line 64: This should be `Errorf`

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,113 @@
...
+		}
+		return "", err
+	}
+	logrus.Infof("Created Bucket %s in S3 ", name)
+	// Bucket name is same as bucket_id in S3
+	return name, nil
+}
+
+func (d *S3Driver) DeleteBucket(id string) error {
+	// DeleteBucket is possible only if the Bucket is empty.
```
ggriffiths, line 55: I think we should make this an option for the end-user. Let's add a bool parameter `clearBucket` and only clear the bucket if it's true. We can expose this as a BucketAccessClass option.

The end-user make not want to delete the bucket if it hasn't be cleared beforehand.

bucket/drivers/s3/s3.go
ggriffiths: As a result, we'll want to create the session/svc for every Create/Delete call that comes in for now. We can do this in a helper function.

pp511: This will mean adding 3 more parameters to “BucketCreateRequest” param in the OpenStorageBucket service.
Region,
AccessKeyId,
SecretAccessKey
The same 3 parameters will need to be passed to the Bucket and hence BcketDriver interface would need to be updated with the same 3 params.

Two of the above parameters excluding region will need to be passed to the BucketDeleteRequest too.

Can you confirm it is ok to update the Bucket Interface and Request params for Create and Delete  APIs like this?

bucket/drivers/s3/s3.go
pp511: Did’t intend to “panic” here. It is a residue of some testing code that I had added and forgot to remove. Will remove it.
1. BucketAlreadyExists =>  Bucket name needs to be updated since it is not globally unique.  return “”, err
2. BucketAlreadyOwnedByYou => Bucket was created by the current requester. =>   return id, nil
3. Any Other Error => return “”, err
4. Success => return id, nil
So we can combine (1, 3 )   and (2, 4). Does that make sense?
But note that if the controller needs to know if the bucket was created with the current call vs if it already existed, then combining 2, 4 will not work.

bucket/drivers/s3/s3.go
ggriffiths: Let's only add the Region parameter to Create/Delete. We can store the AccessKeyID/SecretAccessKey in the s3 driver struct itself.

On bucket create/delete, we can reference the AccessKeyID/SecretAccessKey from the s3 struct.

bucket/drivers/s3/s3.go
ggriffiths: Yes exactly, that makes sense.

The controller will not care about whether the bucket already exists. In CSI, use the concept of idempotency a lot to ensure that if multiple requests reach the driver, the driver will respond the same way. 

Here is an example in CSI where we return early if the desired expansion size equals the current size already:
https://github.com/libopenstorage/openstorage/blob/master/csi/controller.go#L711-L727

Another example is in sdk.DeleteVolume, we return success if the volume cannot be found:
https://github.com/libopenstorage/openstorage/blob/master/api/server/sdk/volume_ops.go#L336-L347

This is effectively saying: you requested a volume to be deleted. It has already been deleted, so the call is successful.

bucket/drivers/s3/s3.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+func (d *S3Driver) updateRegion(region string) {
+	d.config.Region = aws.String(region)
+}
+
+func (d *S3Driver) CreateBucket(name string, region string) (string, error) {
+	// Update driver region config
+	d.updateRegion(region)
+	svc, err := d.NewSvc()
```
ggriffiths, line 48: Let's just pass in the region to d.NewSvc() instead of updating the global config object. If multiple bucket requests come in at the same time, this might cause a concurrency issue. It is also less reliable if multiple requests come in - regions could be overwritten.

ggriffiths (approved):
Just one thing to fix. Good to merge after that.

PR #2070, opened 2022-06-06
PWX-24205: Extend for thin pool metadata disk in storage resource

sulakshm:

Extend storage resource spec to carry information as to whether the disk resource is to be used for metadata while provisioning thin pools.

Closes # https://portworx.atlassian.net/browse/PWX-24205


ram-infrac (approved):
lgtm

PR #2073, opened 2022-06-08
PWX-24205: extend for thin pool metadata disk

sulakshm:

Closes # https://portworx.atlassian.net/browse/PWX-24205


api/api.proto
```diff
@@ -231,7 +231,7 @@ message StorageResource {
   // True if dedicated as cache
   bool cache = 14;
   // True if the resource is used as thin pool metadata disk
-  bool thin_pool_md = 15;
+  bool pool_metadata_dev = 15;
 }
```
ggriffiths, line 235: You'll need to update the Minor version since we're updating an api.proto field here

sulakshm: resolved

PR #2074, opened 2022-06-09
PWX-23408: Added missing Pure backend volume name parsing

Pure-AdamuKaapan:
It looks like this change was brought in to vendor manually but somehow didn't actually make it into openstorage? This PR catches it up with all that should be there. Genuinely not sure how this built before...


ggriffiths (approved):

Pure-AdamuKaapan:
It's not there either, I think it genuinely just slipped by as a vendor-only change :( that's on me

ggriffiths:
Ah ok, no problem! good that we caught this.

PR #2086, opened 2022-07-07
[BucketDriver] Add admin account as AccountID

ggriffiths:

Closes #


ggriffiths:
The controller was silently erroring out without this change. Even though there is no account ID, we were failing to RevokeAccess because an empty string was passed in. No functional difference here, but to avoid confusing errors this is needed.

pp511 (approved):

PR #2088, opened 2022-07-20
PWX-24720: Changed mounter to use regex instead of prefix

Pure-AdamuKaapan:
Signed-Off-By: Adam Krpan <akrpan@purestorage.com>

As part of enabling NVMe FlashArray drives, we now have to handle a different format of path as well which is not just a prefix. We specifically need to search for the pattern `/dev/mapper/*24a937` (note the asterisk in the middle). The easiest way I could think of doing this is to change from a string prefix to just passing in regexes instead. That way we can still specify prefixes by using the regex "start" character, but also get more complex cases.

Closes #

Tests to run:
- [x] Validate that this works with Pure DA volumes
- [x] Validate that this works with normal PX volumes
- [x] Validate that this works with NFS proxy volumes
- [x] Validate that this works with sharedv4 volumes

### DIFF
--- a/pkg/mount/custom_mount.go
+++ b/pkg/mount/custom_mount.go
@@ -1,9 +1,13 @@
 package mount
 
-import "github.com/libopenstorage/openstorage/pkg/keylock"
+import (
+	"regexp"
+
+	"github.com/libopenstorage/openstorage/pkg/keylock"
+)
 
 // CustomLoad defines the mounter.Load callback function for customer mounters
-type CustomLoad func([]string, DeviceMap, PathMap) error
+type CustomLoad func([]*regexp.Regexp, DeviceMap, PathMap) error
 
 // CustomReload defines the mounter.Reload callback function for customer mounters
 type CustomReload func(string, DeviceMap, PathMap) error
@@ -20,7 +24,7 @@ type CustomMounterHandler struct {
 
 // NewCustomMounter returns a new CustomMounter
 func NewCustomMounter(
-	devPrefixes []string,
+	devRegexes []*regexp.Regexp,
 	mountImpl MountImpl,
 	customMounter CustomMounter,
 	allowedDirs []string,
@@ -38,7 +42,7 @@ func NewCustomMounter(
 	cl, cr := customMounter()
 	m.cl = cl
 	m.cr = cr
-	err := m.Load(devPrefixes)
+	err := m.Load(devRegexes)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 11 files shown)


pkg/mount/custom_mount.go
```diff
@@ -46,7 +50,7 @@ func NewCustomMounter(
 }
 
 // Load mount table
-func (c *CustomMounterHandler) Load(devPrefixes []string) error {
+func (c *CustomMounterHandler) Load(devPrefixes []*regexp.Regexp) error {
```
pure-jliao, line 53: nit: if using regex, do we still call this devPrefixes?

pkg/mount/deleted_mount.go
```diff
@@ -25,7 +26,7 @@ func NewDeletedMounter(
 	mountImpl MountImpl,
 ) (*deletedMounter, error) {
 
-	devMounter, err := NewDeviceMounter([]string{"/dev/"}, mountImpl, nil, "")
+	devMounter, err := NewDeviceMounter([]*regexp.Regexp{regexp.MustCompile("/dev/")}, mountImpl, nil, "")
```
pure-jliao, line 29: when using as prefix, do we need to use this "^/dev/" instead?

pure-jliao (approved):
LGTM, only concern is we changed string to regex, is it needed to add ^ or $ in some cases?

pkg/mount/custom_mount.go
Pure-AdamuKaapan: Good point, I don't think so.

pkg/mount/deleted_mount.go
Pure-AdamuKaapan: I did some testing in the go playground, if I just give it `"/dev/"` it will match `/dev/` anywhere in the string (prefix or not). I can't think of any other possibility where `/dev/` would show up so I think it's fine, but we can definitely add the caret if need be

pkg/mount/nfs.go
```diff
@@ -49,7 +49,7 @@ func NewNFSMounter(servers []string,
 
 // Reload reloads the mount table for the specified source/
 func (m *nfsMounter) Reload(source string) error {
-	newNFSm, err := NewNFSMounter([]string{NFSAllServers},
+	newNFSm, err := NewNFSMounter([]*regexp.Regexp{},
```
adityadani, line 52: Use NFSAllServers here in the regex

adityadani (approved):
looks good. Merge it once tested for all volume types. Make sure you do the following testing:
For the following volume types
- RWO volumes - FA & PX
- RWX volumes - FB and sharedv4
Run the following:
- Setup / Teardown of pods: ensure devices are mounted on the right path
- Ensure mountpaths are getting cleaned up from the node after Teardown
Running torpedo suite might be the easiest to cover all of these.

dsupure (approved):
LGTM!

Pure-AdamuKaapan:
@adityadani I've manually confirmed all the given test cases for all the combinations. Thanks!

ggriffiths (approved):

PR #2095, opened 2022-08-05
NFS credential support

talakad:

Closes #


api/api.proto
```diff
@@ -3258,6 +3272,18 @@ message SdkGoogleCredentialResponse {
   string project_id = 2;
 }
 
+// Defines the response for NFS credential
+message SdkNfsCredentialResponse {
```
adityadani, line 3276: Why do we need to return all the input parameters back?

adityadani (approved):
lgtm

api/api.proto
talakad: We dont have to. Its sort of data that helps confirmation/enumerate for callers

PR #2096, opened 2022-08-05
PWX-26155: Setting default region for purefb in case of missing region.

pp511:

Closes #PWX-26155

Pure FB doesn't require setting a region. We are using S3 compatible APIs which expect a region to be passed so passing a default region in create and delete bucket api calls to satisfy the region check.

### DIFF
--- a/api/server/sdk/bucket_ops.go
+++ b/api/server/sdk/bucket_ops.go
@@ -33,6 +33,13 @@ type BucketServer struct {
 	server serverAccessor
 }
 
+const (
+	// S3 Drive name
+	S3Driver = "S3Driver"
+	// PureDBDrive name
+	PureFBDriver = "PureFBDriver"
+)
+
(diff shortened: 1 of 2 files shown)


ggriffiths (approved):

ggriffiths:
Looks good to me

api/server/sdk/bucket_ops.go
```diff
@@ -52,9 +53,13 @@ func (s *BucketServer) Create(
 	}
 	region := req.GetRegion()
 	if len(region) == 0 {
-		return nil, status.Error(
-			codes.InvalidArgument,
-			"Must supply the region")
+		if strings.Compare(s.driver(ctx).String(), "PureFBDriver") == 0 {
```
lpitstickpstg, line 56: nit: is there a constant for this?

pp511: Didn't have. Created one.

PR #2099, opened 2022-08-17
PWX-26324: extend fastpath config for disabling auto fallback

sulakshm:
Extend spec to keep forced failover request from user and disable automatic fallback to fastpath.

Closes #
part of https://portworx.atlassian.net/browse/PWX-26324


ram-infrac (approved):
lgtm

PR #2102, opened 2022-08-25
PWX-26406-pt1: Optimizing NodeUnpublishVolume

zoxpx:
* replacing "deep volume inspect" with cheaper equivalent "volume enumerate" call

The `NodeUnpublishVolume()` was using the "expensive" volume-inspect call, just to determine if the volume already existed, so that a different error can be displayed.

As a fix, we're using a "close equivalent" volume-enumerate call, which returns the volume-data from the cache, and avoids "deep volume inspect".  The returned data-structures are identical, also the lower-level function that gets called, is the same between the "volume-inspect" and "volume-enumerate" (except with "deep=true/false" parameter difference).

Closes # PWX-26406

### DIFF
--- a/csi/node.go
+++ b/csi/node.go
@@ -233,7 +233,7 @@ func (s *OsdCsiServer) NodeUnpublishVolume(
 	}
 
 	// Get volume information
-	vols, err := s.driver.Inspect([]string{req.GetVolumeId()})
+	vols, err := s.driver.Enumerate(&api.VolumeLocator{VolumeIds: []string{req.GetVolumeId()}}, nil)
 	if err != nil || len(vols) < 1 {
 		if err == kvdb.ErrNotFound {
 			clogger.WithContext(ctx).Infof("Volume %s was deleted or cannot be found: %s", req.GetVolumeId(), err.Error())
--- a/csi/node_test.go
+++ b/csi/node_test.go
@@ -839,7 +839,7 @@ func TestNodeUnpublishVolumeVolumeNotFound(t *testing.T) {
 		// Getting volume information
 		s.MockDriver().
 			EXPECT().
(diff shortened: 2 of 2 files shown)


harsh-px:
Need to fix mockgen

```
Unexpected call to *mock.MockVolumeDriver.Enumerate([volume_ids:"myvol" map[]]) at /go/src/github.com/libopenstorage/openstorage/csi/node.go:236 because: there are no expected calls of the method "Enumerate" for that receiver
```

harsh-px (approved):

zoxpx:
> Need to fix mockgen
> 
> ```
> Unexpected call to *mock.MockVolumeDriver.Enumerate([volume_ids:"myvol" map[]]) at /go/src/github.com/libopenstorage/openstorage/csi/node.go:236 because: there are no expected calls of the method "Enumerate" for that receiver
> ```

Should be fixed with [8c996ff](https://github.com/libopenstorage/openstorage/pull/2102/commits/8c996fff194731079e9e8c3690ba9526a9b8cc17)

ggriffiths (approved):
Lgtm, thanks @zoxpx ! Did you notice any other areas in the CSI driver where we could replace `Inspect` with `Enumerate`?

zoxpx:
> Lgtm, thanks @zoxpx ! Did you notice any other areas in the CSI driver where we could replace `Inspect` with `Enumerate`?

@ggriffiths -- I think this one also gets triggered with every vulumeSet-operation @ https://github.com/libopenstorage/openstorage/blob/master/api/server/volume.go#L468, but I am not sure if it would be safe to replace it with `Enumerate()` call.

zoxpx:
Thanks for the review Harsh and Grant -- merging this PR

PR #2103, opened 2022-08-25
PWX-26578: fix the slow task scheduling when runOnce tasks are present

pureneelesh:

    In the scheduleTasks function, we iterate over all the tasks to find the ready tasks and
    feed them to the workers. When we encounter a runOnce task, we feed it to the workers and
    then remove it from the task list. But the removal resets the current element's next pointer
    to nil. This terminates the iteration prematurely causing other ready tasks to not get added
    to the work queue.

    Now, we remove the runOnce tasks outside of the iteration.

    Also, added dynamic creation of new worker goroutines.

    - start a new batch (10) of worker goroutines if the number of enqueued tasks
      is high enough

    - excess workers exit after being idle for 30 seconds

    - total number of workers is capped at 100

    Also,

    - fixed staticcheck warnings

    - added some basic stats collection to the scheduler

PWX-26578

Testing:
Modified one of the unit tests. It now fails with the original code.
```
--- FAIL: TestMulti (3.00s)
    /root/git/go/src/github.com/libopenstorage/openstorage/pkg/sched/sched_test.go:106: 
        	Error Trace:	sched_test.go:106
        	Error:      	Should be true
        	Test:       	TestMulti
        	Messages:   	count for runOnce task 3 was not updated
FAIL
FAIL	github.com/libopenstorage/openstorage/pkg/sched	3.004s
FAIL
```

Also, repro'ed the slow scheduling problem locally by creating a lot of mountPath removal tasks. The tasks were running late by as much as 6 minutes. The issue was not seen after the fix.

### DIFF
--- a/pkg/sched/sched.go
+++ b/pkg/sched/sched.go
@@ -18,6 +18,9 @@ const (
 	TaskNone         = TaskID(0)
 	workerBatchSize  = 10
 	maxWorkers       = 100
+	// Need to see how this value of idleTimeout works out. We may need to increase it if
+	// we see too much churn in the workers. But on the other hand, increasing the value by a lot could
+	// cause the excess workers to never exit since they might pick up tasks in round-robin fashion.
 	idleTimeout      = 30 * time.Second
 	statsLogDuration = 10 * time.Minute
 )
@@ -32,13 +35,21 @@ type histogram interface {
 }
 
 type histImpl struct {
-	lock       sync.Mutex
+	// lock for this struct
+	sync.Mutex
+	// name of the histogram
 	name       string
-	min        time.Duration
-	max        time.Duration
-	multiplier int
+	// each bucket holds the count of values that fall between the limits of the previous bucket and this bucket
 	buckets    []uint64
+	// limits for each bucket (except for the last one which is a catchall bucket);
+	// these are calcuated using min, max and multiplier
 	limits     []time.Duration
+	// limit for the first bucket; values less than min are counted in the first bucket
+	min        time.Duration
+	// limit for the second-to-last bucket; values greater than max are counted in the last bucket
+	max        time.Duration
+	// each bucket's limit is calculated by multiplying the previous bucket's limit by this value
+	multiplier int
 }
 
 func newHistogram(name string, min, max time.Duration) histogram {
@@ -58,8 +69,8 @@ func newHistogram(name string, min, max time.Duration) histogram {
 }
 
 func (h *histImpl) record(val time.Duration) {
-	h.lock.Lock()
-	defer h.lock.Unlock()
+	h.Lock()
+	defer h.Unlock()
 	for i := range h.limits {
 		if val < h.limits[i] {
 			h.buckets[i]++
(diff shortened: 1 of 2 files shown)


sangleganesh (commented):
thanks for the fix. Should we consider dynamically adjusting the number of go routines instead of adding the log ? The number 10 was just arbitrary when we came up with it. 

Also, logs may become too frequent since this is per second. It may be better to log once every 10 minutes.

pureneelesh:
> thanks for the fix. Should we consider dynamically adjusting the number of go routines instead of adding the log ? The number 10 was just arbitrary when we came up with it.
> 
Added dynamic increase/decrease of worker threads.

> Also, logs may become too frequent since this is per second. It may be better to log once every 10 minutes.
Removed the log lines and added some basic stats collection and periodic logging instead.

ggriffiths (approved):
Looks good! Nice change

pkg/sched/sched.go
```diff
@@ -3,23 +3,96 @@ package sched
...
 func ValidTaskID(t TaskID) bool { return t != TaskNone }
 
 type ScheduleTask func(Interval)
 
+type histogram interface {
+	record(time.Duration)
+	getHistogram() string
+}
+
+type histImpl struct {
```
sangleganesh, line 34: nit: add comments for members

pkg/sched/sched.go
```diff
@@ -3,23 +3,96 @@ package sched
...
+	lock       sync.Mutex
```
sangleganesh, line 35: can use composition for lock ?

pkg/sched/sched.go
```diff
@@ -149,38 +236,90 @@ func (s *manager) Start() {
...
+			logrus.Infof("sched stats: workers: current=%v, started=%v, exited=%v",
+				workersStarted-workersExited, workersStarted, workersExited)
+			for _, hist := range []histogram{
+				s.taskDurationHist, s.taskStartHist, s.taskScheduleHist, s.tickIntervalHist, s.tickProcessHist} {
+
+				logrus.Infof("sched stats: %s", hist.getHistogram())
 			}
-			s.cv.Broadcast()
-			s.enqueuedTasksLock.Unlock()
+			histLastPrinted = time.Now()
```
sangleganesh, line 284: can use now from line 242 /

pkg/sched/sched.go
```diff
@@ -149,38 +236,90 @@ func (s *manager) Start() {
...
+		prevTick = time.Now()
 	}
 }
 
-func (s *manager) runTasks() {
+// for testing
+func (s *manager) getWorkerCount() int {
+	var ret int
+	s.Lock()
+	ret = int(s.workersStarted - s.workersExited)
```
sangleganesh, line 295: nit: can simply be `return int(s.workersStarted - s.workersExited)` with lock and defer ?

sangleganesh (commented):
still reviewing ...

sangleganesh (approved):
Looks great, thanks for fixing.
only minor suggestion would be to reduce number of works slowly since we already paid the cost of creating them, but that can be based on your observations on how this is working out.

pkg/sched/sched.go
pureneelesh: added

pureneelesh:
> only minor suggestion would be to reduce number of works slowly since we already paid the cost of creating them, but that can be based on your observations on how this is working out.

Added a comment about this. But left the idleTimeout at 30 seconds for now. In my testing, 10 workers seem to be enough after another fix to avoid the duplicate mountPath removal tasks.

PR #2104, opened 2022-08-29
PWX-26615: avoid scheduling duplicate mountpath removal tasks

pureneelesh:

We delay the removal of mountPath by 30 seconds. CSI NodeUnpublishVolume()
call fails if the mountPath still exists and kubelet retries the same call
call in quick succession (before the back-off slows it down). This
causes us to add duplicate tasks for removing the same mountPath.

This patch detects that we already scheduled the mountPath reomoval and
skips adding a task again.

PWX-26615

Tested by adding a temporary debug log line.


ggriffiths (approved):
lgtm

PR #2108, opened 2022-09-01
PWX-26709: don't print 0 valued buckets in sched stats

pureneelesh:

Also, don't print the histogram when there are no samples.

PWX-26709


ggriffiths (approved):

sangleganesh (approved):
👍 thanks for fixing

PR #2112, opened 2022-09-06
Adds a missing method that slipped into Porx directly

Pure-AdamuKaapan:
Cherry pick into release-9.5

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1362,3 +1362,19 @@ func (s *ProxySpec) IsPureImport() bool {
 
 	return (s.PureBlockSpec != nil && s.PureBlockSpec.FullVolName != "") || (s.PureFileSpec != nil && s.PureFileSpec.FullVolName != "")
 }
+
+func (s *ProxySpec) GetPureFullVolumeName() string {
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -1362,3 +1363,24 @@ func (s *ProxySpec) IsPureImport() bool {
...
+		return s.PureFileSpec.FullVolName
+	}
+
+	return ""
+}
+
+// GetAllEnumInfo returns an EnumInfo for every proto enum
+func GetAllEnumInfo() []protoimpl.EnumInfo {
+	return file_api_api_proto_enumTypes
+}
```
pureneelesh, line 1386: This function is not present in the commit that is being cherry-picked. How did this get added?

Pure-AdamuKaapan: That is... a good question, I don't know. I'll fix up the commit.

Pure-AdamuKaapan: Fixed. I remember having weird issues with that method the first time I added this method too a few months ago, so I dunno what's up with that.

pureneelesh (approved):

PR #2114, opened 2022-09-13
[CSI] Return non-final NodePublishVolume codes and cleanup

ggriffiths:

* The kubelet retry logic changes based on the NodePublishVolume / NodeUnpublishVolume status codes returned:
https://github.com/kubernetes/kubernetes/blob/64ed9145452d2d1d324d2437566f1ea1ce76f226/pkg/volume/csi/csi_client.go#L718-L724
* Same with provisioner: https://github.com/kubernetes-csi/external-provisioner/blob/b3d0a89b38dede277ea1a5c200f3da6a328bc3e7/pkg/controller/controller.go#L1803

Closes #

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -195,7 +195,7 @@ func (s *OsdCsiServer) ValidateVolumeCapabilities(
 	conn, err := s.getConn()
 	if err != nil {
 		return nil, status.Errorf(
-			codes.Internal,
+			codes.Unavailable,
 			"Unable to connect to SDK server: %v", err)
 	}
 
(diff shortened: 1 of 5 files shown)


pureneelesh:
Although the name of the function is tempting in concluding that "final" error means no retries, I could not find where kubelet uses the gRPC error code to decide whether to retry or not. It seems to use that just to keep track of whether the mount state is "uncertain". This is used when deciding whether an unmount is later needed or not?

ggriffiths:
Interesting, in the provisioner there's a direct translation between grpc error type and whether to retry/check:
https://github.com/kubernetes-csi/external-provisioner/blob/b3d0a89b38dede277ea1a5c200f3da6a328bc3e7/pkg/controller/controller.go#L1803

I'll look further in the kubelet.

pureneelesh (approved):

PR #2116, opened 2022-09-14
PWX-26792 Fix Race condition in unmount

dahuang-purestorage:
**What this PR does / why we need it**:
We hit a race condition in unmount with concurrent iteration and write. This was due to unlocking too early before reading ends. This commit adds the fix and UT to verify.

**Which issue(s) this PR fixes** (optional)
JIRA: PWX-26792

**Special notes for your reviewer**:
- Testing: UT passed
- UT error without the fix when run:
`go test  github.com/libopenstorage/openstorage/pkg/mount/... `
```
fatal error: concurrent map iteration and map write

goroutine 153 [running]:
runtime.throw(0xb74a14, 0x26)
	/usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc0001e7c78 sp=0xc0001e7c48 pc=0x4366f2
runtime.mapiternext(0xc0001e7e50)
	/usr/local/go/src/runtime/map.go:853 +0x552 fp=0xc0001e7cf8 sp=0xc0001e7c78 pc=0x411492
github.com/libopenstorage/openstorage/pkg/mount.(*Mounter).Unmount(0xc0001acb98, 0xc000594960, 0x1b, 0xc000594940, 0x1b, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/dahuang/go/src/github.com/libopenstorage/openstorage/pkg/mount/mount.go:613 +0xa07 fp=0xc0001e7ed0 sp=0xc0001e7cf8 pc=0x9b45e7
github.com/libopenstorage/openstorage/pkg/mount.mountTestParallel.func3(0xc000594960, 0x1b, 0xc000594940, 0x1b)
	/home/dahuang/go/src/github.com/libopenstorage/openstorage/pkg/mount/mount_test.go:111 +0x87 fp=0xc0001e7f48 sp=0xc0001e7ed0 pc=0x9bd8a7
github.com/libopenstorage/openstorage/pkg/mount.mountTestParallel.func4(0xc00051c310, 0xc000594920, 0x1a, 0xc000594940, 0x1b, 0xc00051c320, 0xc00051c330, 0xc000594960, 0x1b, 0xc000259e30)
	/home/dahuang/go/src/github.com/libopenstorage/openstorage/pkg/mount/mount_test.go:126 +0xc7 fp=0xc0001e7f90 sp=0xc0001e7f48 pc=0x9bd9f7
runtime.goexit()
[...]

PR #2118, opened 2022-09-15
PWX-26782 Fix Race condition in unmount (#2116)

dahuang-purestorage:
We hit a race condition in unmount with concurrent iteration and write. This was due to unlocking too early before reading ends. This commits adds the fix and UT to verify.

JIRA: PWX-26782

Testing: UT passed


pureneelesh:
Hey @dahuang-purestorage : can you check the load() function in the same file? Does it also need to use lock for that map?

PR #2124, opened 2022-10-19
PWX-27439-pt1 Pass trashlocation in NFS mount

dahuang-purestorage:

Previously, mount path was never removed because scheduled task was cancelled when PX is down. And NFS mount does not pass on the trashlocation for symlink removal. This commit adds the parameter and more logging for when removal happens.

PWX-27439

There will be a second part of related to the ticket on the porx side, to pass the trashlocation down from the mounter.New()

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 package mount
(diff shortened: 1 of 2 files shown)


pkg/mount/mount.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux
```
pureneelesh, line 1: is this intended? What does it do?

pureneelesh (approved):

pkg/mount/mount.go
dahuang-purestorage: It was added automatically by VSCode.
those serve the same purpose as `// +build linux`, but for go version 1.6+. 
Those are build tags. Basically saying the build should be under linux environment. 
Doc: [constraints are satisfied](https://pkg.go.dev/cmd/go#hdr-Build_constraints)

adityadani (approved):

PR #2127, opened 2022-10-20
Pass trashlocation in NFS mount

dahuang-purestorage:
Previously, mount path was never removed because scheduled task was cancelled when PX is down. And NFS mount does not pass on the trashlocation for symlink removal. This commit adds the parameter and more logging for when removal happens.

Testing: local testing

Closes #


dahuang-purestorage:
This is moving to 2.12.1, will add back next release.

PR #2129, opened 2022-10-25
Remove problematic go.pedge.io dependency

ggriffiths:

Remove problematic go.pedge.io dependency

Dependabot was failing due to this dependency and we don't use it much. Substituting it for our own prototime pkg.

Closes #


dahuang-purestorage (approved):
LGTM!

PR #2147, opened 2022-11-07
PWX-27736: Use 9001 as the default rest port when TLS is not enabled

pp511:

With TLS disabled,  9023 is being reported as REST port and all the communications to this endpoint are failing.

Closes #PWX-27736

* finalRestPort was set to restPortSecured
* This should happen only when TLS is enabled


jrivera-px (approved):
LGTM!!

ggriffiths (approved):
Looks good once unit tests are fixed

PR #2159, opened 2022-12-01
PWX-26915-pt1: Auto-TLS support

zoxpx:
* expand to "full" service names when using Kuberenetes DNS for service discovery
* adding Kuberenetes CA for gRPC dial-opts when SSL enabled

These changes are a part of the "auromatic SSL/TLS setup" effort

* change to "full" service names when using Kuberenetes DNS for discovery
  - required since the SSL certs are registered to FULL K8s dns-name `portworx-service.kube-system.svc.cluster.local`
  - using short `portworx-service.kube-system` dns-alias will result in SSL-validation errors
* adding Kuberenetes CA for gRPC dial-opts when SSL enabled
  - required since PX certs will be signed by K8s CA

Closes #  PWX-26915 (part 1)

### DIFF
--- a/volume/drivers/pwx/connection.go
+++ b/volume/drivers/pwx/connection.go
@@ -9,11 +9,10 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/libopenstorage/openstorage/pkg/grpcserver"
 	"github.com/portworx/sched-ops/k8s/core"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
-
-	"github.com/libopenstorage/openstorage/pkg/grpcserver"
 )
 
 // ConnectionParamsBuilder contains dependencies needed for building Dial options and endpoints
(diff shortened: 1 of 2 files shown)


volume/drivers/pwx/connection.go
```diff
@@ -225,6 +225,18 @@ func (cpb *ConnectionParamsBuilder) BuildDialOps() ([]grpc.DialOption, error) {
...
+		if len(rootCA) <= 0 {
+			rootCA = k8sCA
+		} else {
+			rootCA = bytes.Trim(rootCA, "\r\n")
+			rootCA = append(rootCA, '\n')
+			rootCA = append(rootCA, k8sCA...)
+		}
+	}
+
 	tlsDialOptions, err := grpcserver.GetTlsDialOptions(rootCA)
```
jrivera-px, line 240: If the above condition fails... is there a logging message we want to add?

jrivera-px (approved):
LGTM .. I was thinking that if the CA is required at this point for k8s do we want to log something if its not avail?

zoxpx:
> LGTM .. I was thinking that if the CA is required at this point for k8s do we want to log something if its not avail?

I think we should be OK "as is" --
* the `libopenstorage` should work also in non-K8s environments, so there is no guarantee the K8s-CA file will be present.
* additionally, if the CA is not set up properly, the SSL-connect should throw the appropriate error to indicate that host-cert has not been signed by a known/registered CA  (note, it's a very common error, so they had to make the baseline error-response very clear).

zoxpx:
note.. re-pushed `[6a479ac](https://github.com/libopenstorage/openstorage/pull/2159/commits/6a479ac0e4681a899930758983f62e23af936277)` diff with DCO in the comment  (travis-build was failing)

zoxpx:
Thanks for the review Jose -- merging the PR

PR #2162, opened 2022-12-07
PWX-23422: added master reverted changes of porx

ashinde-px:
- First these changes added in the vendor folder of porx repo, but due to other's commits our changes are reverted from porx master branch
- Added master reverted changes from the porx repo due to make vendor by others
- So these changes should be here and make vendor will auto pop these changes to the vendor dir of porx.

Closes #PWX-23422


api/api.go
```diff
@@ -62,6 +62,7 @@ const (
 	SpecExportOptions        = "export_options"
 	SpecExportOptionsEmpty   = "empty_export_options"
 	SpecMountOptions         = "mount_options"
+	SpecFaCreateOptions      = "createoptions"
```
ggriffiths, line 65: please follow the naming convention of other variables in the spec-
`fa_creation_options`

ggriffiths (changes requested):

api/api.go
ggriffiths: ok,  please document that as a comment above this line. Mention that this spec key cannot change due to parity with existing PSO storageclasses

ashinde-px: Sure

ggriffiths (approved):

PR #2166, opened 2022-12-08
[CSI] Add support for distributed deletes

ggriffiths:

Found that deletes for CSI are taking a long time, and not using the distributed provisioning model.

Closes #


dahuang-purestorage:
how did you find that the deletion takes a long time?
I'm guessing this adds the distributed deletes by calling a remote client that might be available, instead of using the local ones?

ggriffiths:
> how did you find that the deletion takes a long time? 
I was testing PX volume deletes and noticed all traffic was funneled to a single node. The remote client calls out to various nodes. 

> Probably worth adding a JIRA ticket if it's linked 
Thanks, yeah here it is- PWX-28149

> I'm guessing this adds the distributed deletes by calling a remote client that might be available, instead of using the local ones?
Yes that's correct - I'm doing some manual tests now, but it's the same RPC call but to a remote node.

dahuang-purestorage (approved):
lgtm. thanks for the explanation :)

PR #2169, opened 2022-12-08
PWX-26915-pt1.r2/PWX-28189: Auto-TLS support r2 (#2159)

zoxpx:
* moved K8s CA addition into `appendK8sCaCertBytes()`
* calling it from both `BuildTlsConfig()` and `BuildDialOps()`

* making the addition of the Kubernetes CA more generic, so it can be called from multiple methods
* will call `appendK8sCaCertBytes()` for both -- 
  - `BuildDialOps()`  (client for portworx-api:9020) as well as
  - `BuildTlsConfig()` (client for portworx-api:9023) 

Closes #
- PWX-26915 (part 1 - reworked), also
- PWX-28189 (stork auto-SSL support)

### DIFF
--- a/volume/drivers/pwx/connection.go
+++ b/volume/drivers/pwx/connection.go
@@ -181,6 +181,14 @@ func (cpb *ConnectionParamsBuilder) BuildTlsConfig() (*tls.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if upd, err := cpb.appendK8sCaCertBytes(rootCA); err == nil {
+		logrus.Debugf("Found Kubernetes CA, adding into TLS config")
+		rootCA = upd
+	} else if !os.IsNotExist(err) {
+		logrus.WithError(err).Warnf("Could not append Kubernetes CA")
+	}
+
 	tlsCfg := &tls.Config{}
 	setRootCA(tlsCfg, rootCA)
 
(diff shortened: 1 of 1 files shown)


volume/drivers/pwx/connection.go
```diff
@@ -181,6 +181,14 @@ func (cpb *ConnectionParamsBuilder) BuildTlsConfig() (*tls.Config, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	if upd, err := cpb.appendK8sCaCertBytes(rootCA); err == nil {
+		logrus.Debugf("Found Kubernetes CA, adding into TLS config")
+		rootCA = upd
+	} else if !os.IsNotExist(err) {
+		logrus.WithError(err).Warnf("Could not append Kubernetes CA")
```
ggriffiths, line 189: if the user intended to setup a rootCA but we can't find it, shouldn't we return error here? Otherwise we'll silently fail and not setup properly

zoxpx: The `cpb.appendK8sCaCertBytes()` appends the Kubernetes CA, and it's meant to be "opportunistic" --
* e.g. if file the is there (e.g. if we're running in Kubernetes env), go ahead and [always] _append_ K8s CA
  - but, if the file was not there (e.g. non-K8s env's), silently ignore the error
* note that it'll add this CA _in addition_ to CA-certs provided via secret
  - see `cpb.getCaCertBytes()` -- note that this func always returns its errors  (does not ignore them)

zoxpx:
This has been stuck on a review for 5 days, and 2 more PRs (JIRA tickets) depend in this change -- could I get a review on this PR please?

ggriffiths (approved):
Looks good, apologies for the delay

zoxpx:
Thanks Grant -- merging the PR

PR #2176, opened 2022-12-22
Add pvc/namespace labels to cloned csi volumes

ggriffiths:

Closes #


csi/controller.go
```diff
@@ -610,6 +610,14 @@ func (s *OsdCsiServer) CreateVolume(
 		if err != nil {
 			return nil, err
 		}
+
+		_, err = volumes.Update(ctx, &api.SdkVolumeUpdateRequest{
```
ggriffiths, line 614: volumes.Clone calls driver.Snapshot, which removes `pvc/namespace` labels to avoid misuse. Another possible fix is to change driver.Snapshot to take additional pvc/namespace parameters, but this fix is far less invasive for a patch.

csi/controller.go
```diff
@@ -610,6 +610,14 @@ func (s *OsdCsiServer) CreateVolume(
...
+			VolumeId: cloneResp.GetVolumeId(),
+			Labels:   getClonedPVCMetadata(locator),
+		})
+		if err != nil {
+			return nil, err
```
pureneelesh, line 619: is there any need to cleanup the cloned volume before returning an error from here?

csi/controller.go
```diff
@@ -610,6 +610,14 @@ func (s *OsdCsiServer) CreateVolume(
 		if err != nil {
 			return nil, err
 		}
+
+		_, err = volumes.Update(ctx, &api.SdkVolumeUpdateRequest{
+			VolumeId: cloneResp.GetVolumeId(),
+			Labels:   getClonedPVCMetadata(locator),
```
pureneelesh, line 616: do we need to worry about merging with the existing labels instead of overwriting?

pureneelesh (approved):

csi/controller.go
ggriffiths: volumes.Updates code suggests we don't, but I'll merge them here to be safe.

csi/controller.go
ggriffiths: yeah, will add this

csi/controller.go
ggriffiths: actually, going to move this to the SDK layer to handle this more broadly. will re-request review then.

csi/controller.go
ggriffiths: no longer an issue since we're handling at the sdk layer

ggriffiths:
test results:
restored pvc has `pvc=` and `namespace=` labels, as well as a bunch of other labels from the original volume.
```
[root@ggriffiths-k8s1-node1 /]# /opt/pwx/bin/pxctl v i 781948984762600138 | grep pvc=
	Labels          	 :  sharedv4=true,sharedv4_svc_type=ClusterIP,csi.storage.k8s.io/pv/name=pvc-b74dda8e-b258-41b0-a3f1-1abaa3e5bc41,csi.storage.k8s.io/pvc/name=mysql-data,csi.storage.k8s.io/pvc/namespace=default,namespace=default,pvc=hpvc-restore,repl=2
```

api/server/sdk/volume_ops.go
```diff
@@ -181,18 +182,45 @@ func (s *VolumeServer) create(
...
+		}
+
+		additionalLabelsNeeded := len(additionalCloneLabels) > 0
+		if additionalLabelsNeeded {
+			mergedLabels := make(map[string]string)
+			clonedLabels := make(map[string]string)
+			// Cloned volume may not have locator or volume.
+			// This is unlikely, but we'll avoid a panic here.
+			if clone.GetVolume() != nil && clone.GetVolume().GetLocator() != nil &&
+				clone.GetVolume().GetLocator().GetVolumeLabels() != nil {
```
pureneelesh, line 206: nit: the last check on line 206 is not needed since clonedLabels will just be nil on line 207 which is ok.

pureneelesh (approved):

PR #2193, opened 2023-01-18
[CSI] Adjust error codes for snapshotting to be non-final

ggriffiths:
We were seeing some cases where snapshotting retries do not happen. 

This happens when an error is non-final:
https://github.com/kubernetes-csi/external-snapshotter/blob/v6.0.1/pkg/sidecar-controller/snapshot_controller.go#L656-L678

In addition, we were checking the SDK Server response to be `kvdb.ErrNotFound`, but the SDK server only returns grpc status errors. We need to check for codes.NotFound instead.

Closes #

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -933,8 +933,12 @@ func (s *OsdCsiServer) CreateSnapshot(
 		Labels:   locator.GetVolumeLabels(),
 	})
 	if err != nil {
-		if err == kvdb.ErrNotFound {
-			return nil, status.Errorf(codes.NotFound, "Volume id %s not found", req.GetSourceVolumeId())
+		errStatus, ok := status.FromError(err)
+		if ok && errStatus.Code() == codes.NotFound {
+			// Return a non-final Aborted error, as the PVC may not have been created yet. Otherwise
+			// the CSI Snapshotter will not issue retries:
+			// https://github.com/kubernetes-csi/external-snapshotter/blob/v6.0.1/pkg/sidecar-controller/snapshot_controller.go#L656-L678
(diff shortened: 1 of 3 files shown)


jrivera-px (approved):

csi/controller.go
```diff
@@ -933,8 +933,12 @@ func (s *OsdCsiServer) CreateSnapshot(
...
 	})
 	if err != nil {
-		if err == kvdb.ErrNotFound {
-			return nil, status.Errorf(codes.NotFound, "Volume id %s not found", req.GetSourceVolumeId())
+		errStatus, ok := status.FromError(err)
+		if ok && errStatus.Code() == codes.NotFound {
+			// Return a non-final Aborted error, as the PVC may not have been created yet. Otherwise
+			// the CSI Snapshotter will not issue retries:
+			// https://github.com/kubernetes-csi/external-snapshotter/blob/v6.0.1/pkg/sidecar-controller/snapshot_controller.go#L656-L678
+			return nil, status.Errorf(codes.Aborted, "Volume id %s not found: %v", req.GetSourceVolumeId(), err)
```
pureneelesh, line 941: This brings back memories of similar issue with another volume op. My recollection is that we should return a "final" error if there is no state to be cleaned up later by CSI. Is that also the case here?

ggriffiths: "Final" errors are seen as errors that cannot be recovered from. I.e. if you provide an invalid argument, there is no sense in retrying.

In this case, we want to retry because a snapshot might have been created at the same time as a volume. With this race condition, a final error is immediately reached if the snapshot create fails before the volume has been created.

ggriffiths: And yeah, we had to make similar changes to make sure sharedv4 unmounts/mounts were retried properly.

PR #2194, opened 2023-01-19
PWX-25792: volume with mount_option nosuid fails to get mounted to the Pod

ashinde-px:
Signed-off-by: Amol Shinde <ashinde@purestorage.com>

volume with mount_option nosuid fails to get mounted to the Pod
Closes #PWX-25792

Adding new attribute CsiMountOptions to handle the user defined mount options as flags


api/api.proto
```diff
@@ -698,6 +698,8 @@ message VolumeSpec {
   bool winshare = 49;
   // Filesystem create options to be honored.
   string fa_create_options = 50;
+  // csi_mount_options user defined mount flags to be honored.
+  string csi_mount_options = 51;
```
ggriffiths, line 702: Since you're updating api.proto, you need to bump the version in this file and add a SDK_CHANGELOG line.

ashinde-px:
Apologies @ggriffiths We have decided to not create new attribute in proto to fix this issue. No longer needed these code changes, we are using other option to fix this ticket. Here is a new PR : https://github.com/portworx/porx/pull/10484

PR #2206, opened 2023-02-14
fix: update ready to use status based on snap error

Glitchfix:
Signed-off-by: Shivanjan Chakravorty <schakravorty@purestorage.com>

the ReadyToUse was hardcoded in every CSI Snapshot response
the aim is to return a false value incase there is an error associated with the snapshot

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -915,7 +915,7 @@ func (s *OsdCsiServer) CreateSnapshot(
 				SnapshotId:     v.GetId(),
 				SourceVolumeId: v.GetSource().GetParent(),
 				CreationTime:   v.GetCtime(),
-				ReadyToUse:     true,
+				ReadyToUse:     v.GetError() == "",
 			},
 		}, nil
 	}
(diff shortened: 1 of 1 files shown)


csi/controller.go
```diff
@@ -915,7 +915,7 @@ func (s *OsdCsiServer) CreateSnapshot(
 				SnapshotId:     v.GetId(),
 				SourceVolumeId: v.GetSource().GetParent(),
 				CreationTime:   v.GetCtime(),
-				ReadyToUse:     true,
+				ReadyToUse:     v.GetError() == "",
```
ggriffiths, line 918: Let's also check for volume up:
`*v.Status.Enum() == api.VolumeStatus_VOLUME_STATUS_UP`

Add v.GetError() and the above check in a helper function. i.e. isSnapshotReady

ggriffiths (changes requested):
One suggested change - can you also build a test image with this change and run the CSI e2e test suite with it? We want to make sure this is fixing the problem.

Glitchfix:
> One suggested change - can you also build a test image with this change and run the CSI e2e test suite with it? We want to make sure this is fixing the problem.

since the test was done and verified, can we merge this PR now?

ggriffiths:
> > One suggested change - can you also build a test image with this change and run the CSI e2e test suite with it? We want to make sure this is fixing the problem.
> 
> since the test was done and verified, can we merge this PR now?

No we cannot, we need to address the comments I have given.

csi/controller.go
Glitchfix: fixed in the latest commit

ggriffiths (approved):

PR #2228, opened 2023-02-23
PWX-23290: handle spec updates with stale config

sulakshm:
Redo https://github.com/libopenstorage/openstorage/pull/2022 to fix the unit tests additionally.

When volume spec updates are requested, there is a spec reconciliation based on the current state of the volume.
This is done to validate policy settings.
For some volume update operations the spec gets updated only upon action from px-storage and may take time to complete action. Ex. HA Update on a volume needs action from px-storage before the HA level change gets reflected in spec.
So, if there is a subsequent volume update, it is possible for the reconciliation action part of update to pick stale config,
and the final action may be inconsistent after.

To avoid this, whenever volume update is applied, only pass the relevant section of the spec that needs update, mask
rest of the specs that have not been updated.

This window of stale config is open only for defered spec updated and long running actions to complete the update operations. Boolean based state are mostly instantaneous updates to the spec and does not seem to have this problem.

Which issue(s) this PR fixes (optional)
Closes # PWX-23290
https://portworx.atlassian.net/browse/PWX-23290

Special notes for your reviewer:
The changes have been tested in 2.10.0-ea branch where the issue was reported.

### DIFF
--- a/api/server/sdk/volume_ops.go
+++ b/api/server/sdk/volume_ops.go
@@ -19,6 +19,7 @@ package sdk
 import (
 	"context"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


sharma-tapas (approved):
LGTM

PR #2239, opened 2023-03-06
Revert "near sync fields"

chiachunlin:
Reverts libopenstorage/openstorage#2230


chiachunlin:
as discussed, this reverts the near sync change to relase9.7 then later we bring all near sync related change to releas 9.7 in one shot.

PR #2243, opened 2023-03-06
PWX-29572: Warnf on CSI sched filter fail instead of erroring out

ggriffiths:
Warnf on CSI sched filter fail instead of erroring out

Closes #

### DIFF
--- a/csi/sched/interceptor.go
+++ b/csi/sched/interceptor.go
@@ -7,8 +7,6 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 )
 
 // FilterInterceptor is a wrapper for the filter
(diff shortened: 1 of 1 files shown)


pp511 (commented):
lgtm.

pp511 (approved):

csi/sched/interceptor.go
```diff
@@ -31,12 +29,11 @@ func (fi *FilterInterceptor) SchedUnaryInterceptor(
 		csiReq := req.(*csi.CreateVolumeRequest)
 		req, err = fi.Filter.PreVolumeCreate(csiReq)
 		if err != nil {
-			logrus.WithContext(ctx).Errorf("CSI pre-create filter failed: %v", err)
+			logrus.WithContext(ctx).Warnf("CSI pre-create filter failed: %v", err)
 
-			// Return an aborted code to retry from the csi-provisioner.
-			// We cannot ignore this error or else a volume will be created w/
-			// incorrect locator.VolumeLabels.
```
Pure-AdamuKaapan, line 38: Has the assumption/logic in this old comment changed? I always get leery about removing "we cannot ignore this" comments :yum:

ggriffiths: Yes, the assumption here was that the only client for this interceptor was the csi-provisioner sidecar. In this case, we should error out because pvc name/namespace will always be provided by the caller.

Since the px2pso tool calls csi.CreateVolume manually before a PVC has been created, this assumption is no longer correct.

Given both of these scenarios, we need to remain flexible when a pvc name/namespace isn't provided.

PR #2258, opened 2023-04-12
PWX-30083: change alert clearing from ttl to background scrub on an h…

sharma-tapas:
…ourly basis

Closes #


pnookala-px:
As per PWX-29519, long lived leases that we do using ttl on alerts is problematic. Given that our alerts ttl is in hours it makes sense for us to run a background thread that wakes up hourly and clean up the leases.

PR #2259, opened 2023-04-14
PWX-30083: Prune alerts using a background thread rather than ttl

pnookala-px:

Closes #

### DIFF
--- a/alerts/alerts.go
+++ b/alerts/alerts.go
@@ -4,6 +4,7 @@ package alerts
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"path/filepath"
 	"strconv"
 	"sync"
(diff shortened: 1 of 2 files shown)


pnookala-px:
As per PWX-30083, alert pruning using ttl is very costly given the time our alerts live. The alerts have a lifetime of 12 hours to 5 days which is not optimal for a ttl workflow which prunes the list every 500 ms. Instead use a background scheduler that runs at hourly intervals.

alerts/alerts.go
```diff
@@ -77,8 +84,36 @@ func newManager(options ...Option) (*manager, error) {
 				return nil, typeAssertionError
 			}
 			m.ttl = v
+			// make alertsbgcrawltime to half of ttl, so that we get two scans atleast.
+			AlertsBgCrawlTime = time.Duration(math.Min(float64(AlertsBgCrawlTime), float64((uint64)(time.Second) * (m.ttl/2))))
 		}
 	}
+	_, err := sched.Instance().Schedule(
```
adityadani, line 91: This schedule will run on all the nodes in a cluster. We should instead move this logic into porx, and ensure that this runs only as part of a cluster co-ordinator.

adityadani: You can introduce a new API/function called Purge in this package and invoke that as part of a schedule defined in porx

PR #2261, opened 2023-04-18
PWX-30083: Optional way to purge alerts

pnookala-px:
PWX-30083: In openstorage either use an option to use ttl to purge alerts or let the alert publisher purge the alerts. Default to the publisher purging the alerts.

### DIFF
--- a/alerts/alerts.go
+++ b/alerts/alerts.go
@@ -53,6 +53,8 @@ type Manager interface {
 	SetRules(rules ...Rule)
 	// DeleteRules deletes rules
 	DeleteRules(rules ...Rule)
+	// Purge alerts.
+	Purge() error
 }
 
 // FilterDeleter defines a list and delete interface on alerts.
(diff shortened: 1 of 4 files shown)


Glitchfix (approved):

alerts/alerts.go
```diff
@@ -248,6 +259,29 @@ func (m *manager) Filter(alerts []*api.Alert, filters ...Filter) ([]*api.Alert,
...
+	if kvdb.Instance() != nil {
+		kvps, err := enumerate(kvdb.Instance(), kvdbKey)
+		if err == nil {
+			for _, kvp := range kvps {
+				alert := new(api.Alert)
+				if err := json.Unmarshal(kvp.Value, alert); err != nil {
+					continue
+				}
+				curtime := (int64)(time.Now().Unix())
+				if (curtime - alert.Timestamp.GetSeconds()) > (int64)(alert.Ttl) {
```
adityadani, line 275: An easier way to do this in golang would be 
```suggestion
				curtime := (int64)(time.Now().Unix())
				if time.Now().Sub(alert.Timestamp)> time.Duration(alert.Ttl) {
```

alerts/alerts.go
```diff
@@ -248,6 +259,29 @@ func (m *manager) Filter(alerts []*api.Alert, filters ...Filter) ([]*api.Alert,
...
+					//Alert has lived its ttl. Time to delete it.
+					kvdb.Instance().Delete(kvp.Key)
```
adityadani, line 277: I would say check for how many deletes fail and return an error in this function indicating some alerts were not purged.

adityadani (approved):

PR #2264, opened 2023-04-25
Remove cred check for nearsync

dgoel-px:
Nearsync migrate does not need credentials and so the cred check is also not needed.


talakad (approved):

adikulkarni0:
Can you write a test for this and post the manual test results to the PR?

PR #2267, opened 2023-05-04
Allow round-robin load balancing of requests for SDK APIs.

adityadani:
- Move the current round-robin logic used by the CSI driver into a separate package.
- Currently only the VolumeMigrate's Start() API uses the round-robin load balancing functionality.
- Modify the CSI code to use the new round robin package.

The round robin package provides a GetRemoteNodeConnection() API
 - The API returns:
   - a gRPC client connection object.
   - a boolean value indicating whether the connection is to a remote node. It is set to false if the round robin node chosen is the current node itself.
   - an error which is returned when the API fails to create a gRPC connection.

Closes #

Eventually we could bake the round-robin balancer into the gRPC interceptor itself and allow each registered API to choose whether they want the round-robin functionality


ggriffiths (approved):
Looks good. Thanks for making this into a pkg!

PR #2269, opened 2023-05-09
Winshare support for openstorage

pnookala-px:

Closes #


.dockerignore
sulakshm, line 1: This change is not needed.

csi/controller.go
```diff
@@ -115,19 +118,95 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
+		logrus.Errorf("failed to get remote connection: %v, continuing with local node instead", err)
+		conn, err = s.getConn()
+		if err != nil {
+			return nil, status.Errorf(
+				codes.Unavailable,
+				"Unable to connect to the SDK server: %v", err)
+		}
+	}
+
+	// TODO: Add checks for capability and readonly, etc.
```
sulakshm, line 148: validation of volume capabilities missing.
see VolumeCreate API for example or others in this file pre-existing.
```
	if spec.IsPureVolume() {
		err = validateCreateVolumeCapabilitiesPure(req.GetVolumeCapabilities(), spec.GetProxySpec())
		if err != nil {
			return nil, err
		}
	} else {
		err = validateCreateVolumeCapabilities(req.GetVolumeCapabilities())
		if err != nil {
			return nil, err
		}
	}
```

csi/controller.go
```diff
@@ -115,19 +118,95 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
+		if s, ok := status.FromError(err); ok && s.Code() == codes.NotFound {
+			/// ignore volume not found error.
+			return &csi.ControllerUnpublishVolumeResponse{}, nil
+		}
+		return nil, err
+	}
+
+	/// have the volume, now submit a new grpc request to the px driver, to
+	/// cleanup volume that was published earlier.
+	/// TODO:
```
sulakshm, line 186: revise/remove comments.

sulakshm (approved):
minor nits inline, need cleanup and capability checks. otherwise okay.

With k8s in place, all volume attachments will follow
1/ ControllerPublish - porx ControllerPublish - attach vol, should setup nfs conns
2/ NodeStage - porx csi node driver does not use this capability but windows does.
3/ NodePublish - porx NodePublish - calls Attach/Mount on volume client.
4/ NodeUnpublish - porx NodeUnpublish - calls Unmount/Detach on volume client.
5/ NodeUnstage - porx csi node driver does not use this capability but windows does.
6/ ControllerUnpublish - porx ControllerUnpublish - should teardown nfs conns, no more refs, should detach volume.

For shared volumes (with/without windows):
Herein, controller side actions would involve, rerouting request to the right replica node, and attach the volume locally to that node and setup nfs server for exporting the same.

While node actions will involve, using the passed context and setting up the nfs client needs.

I wanted to highlight that with the above workflow, pxctl actions like "attach and mount" will not honour the above workflow. 

This workflow needs to be thought over, if windows volumes are not to be thought as an exception.

api/api.proto
```diff
@@ -2511,6 +2511,26 @@ service OpenStorageVolume {
           body: "*"
         };
     }
+
+    // Controller Publish Volume
+    // Publish volume for winshares, this receives the ip and sharename
```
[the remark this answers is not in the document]
pnookala-px: done.

sulakshm (changes requested):
csi/node.go needs change to report STAGE/UNSTAGE capabilities equivalent to windows csi driver
I am not totally sure about this, in addition also note that  csi.NodeServiceCapability_RPC_GET_VOLUME_STATS, is not supported by windows csi driver. 

@ggriffiths can you confirm the below.
A cluster is running a csi driver pxd.portworx.com
some nodes (read linux nodes) have one set of node capabilities for the same driver, while some other nodes (windows nodes) in the cluster have a different set. Is this a supported configuration with CSI+K8s or some undocumented/unsupported mix that we should avoid.


```
// NodeGetCapabilities is a CSI API function which seems to be setup for
// future patches
func (s *OsdCsiServer) NodeGetCapabilities(
	ctx context.Context,
	req *csi.NodeGetCapabilitiesRequest,
) (*csi.NodeGetCapabilitiesResponse, error) {
	clogger.WithContext(ctx).Debugf("csi.NodeGetCapabilities request received")

	caps := []csi.NodeServiceCapability_RPC_Type{
		// Getting volume stats for volume health monitoring
		csi.NodeServiceCapability_RPC_GET_VOLUME_STATS,
		// Indicates that the Node service can report volume conditions.
		csi.NodeServiceCapability_RPC_VOLUME_CONDITION,
	}
```

csi/controller.go
```diff
@@ -115,19 +118,122 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
+			return nil, status.Errorf(
+				codes.Unavailable,
+				"Unable to connect to the SDK server: %v", err)
+		}
+	}
+
+	if !vol.Spec.Winshare {
+		// Ignore for the timebeing
+		logrus.Debugf("ControllerPublishVolume: Not a Windows volume, return")
+		return &csi.ControllerPublishVolumeResponse{}, nil
```
adityadani, line 164: Maintain the original error which is NotImplemented for unsupported volume types.

api/server/sdk/volume_node_ops.go
```diff
@@ -122,6 +122,56 @@ func (s *VolumeServer) Detach(
...
+	if s.cluster() == nil || s.driver(ctx) == nil {
+		return nil, status.Error(codes.Unavailable, "Resource has not been initialized")
+	}
+
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.Unavailable, "VolumeId not provided.")
+	}
+
+	opts := make(map[string]string)
+	err := s.driver(ctx).ControllerUnpublish(ctx, req.GetVolumeId(), req.GetNodeId(), opts)
```
adityadani, line 164: Lets call the actual driver's Attach/Detach/Mount/Unmount functions here.

volume/volume.go
```diff
@@ -324,6 +328,10 @@ type BlockDriver interface {
 	// Detach device from the host.
 	// Errors ErrEnoEnt, ErrVolDetached may be returned.
 	Detach(ctx context.Context, volumeID string, options map[string]string) error
+	// ControllerPublish csi extension
```
adityadani, line 331: I am not in favor of adding ControllerPublish and Unpublish functions to the BlockDriver interface, since they are just a combination of Attach and Detach.

pnookala-px: These changes are needed for compilation support. A driver needs to support all interfaces.

api/server/sdk/volume_node_ops.go
pnookala-px: What you are implying here is that instead of having a ControlerPublish equivalent on the porx side, just have Attach()/Detach () and in CSI driver's CP/CUP call s.Driver.Attach()/s.Driver.Detach() respectively for windows shares? That though is a workable code, it is not clean code.

sulakshm: Hi @adityadani I understand your concern with bringing in new code flow to time tested workflow in sharedv4 code path. But in the long run, I see added value in clearly distinguishing functionality provided by this extension of using CSI ControllerPublish and ControllerUnpublish. Our convoluted way of using proxy objects and filter in the control path can be effectively removed/or atleast the code more made clean.

ControllerPublish and ControllerUnpublish will run only on linux nodes (but on any node). A proxy object redirection is necessary to perform this function on the replica node. Here volume gets attached and nfs server setup, along with sharedv4 service endpoint during setup phase. Controller can keep track of only portworx client nodes, but not external windows clients or outsiders. Not sure how its handled to keep refcounting correct.

For the client side, there would be NodeStageVolume followed by NodePublishVolume. 
NodeStageVolume as a first step, could setup the nfs client. 
NodePublishVolume could keep track of refcounts as necessary.

This is much cleaner abstraction than currently setup; and I would strongly vouch for the change.

api/api.proto
```diff
@@ -2511,6 +2511,26 @@ service OpenStorageVolume {
...
+    rpc ControllerPublish(SdkVolumeControllerPublishRequest)
+    returns (SdkVolumeControllerPublishResponse) {
+        option(google.api.http) = {
+          post: "/v1/volumes/publish/{volume_id}"
+          body: "*"
+        };
+    }
+
+    // Controller Unpublish Volume
+    // Drops the reference on the sharename
```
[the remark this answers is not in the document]
pnookala-px: Taken care off.

csi/controller.go
pnookala-px: The new check, checks that capabilities exist, do we need to check that we have read/write/access permissions?

csi/controller.go
```diff
@@ -115,19 +118,122 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
+	logrus.Debugf("Calling Controller Publish Volume in the csi controller")
+	// check if volumeID is present and volume is available on SP
+	if req.VolumeId == "" {
+		return nil, status.Error(codes.InvalidArgument, "VolumeID is mandatory in ControllerPublishVolume request")
+	}
+	clogger.WithContext(ctx).Infof("ControllerPublishVolume request recived. VolumeID: %s", req.GetVolumeId())
+
+	// if nodeID is set, and node is actually present on SP
+	if req.NodeId == "" {
+		return nil, status.Error(codes.InvalidArgument, "NodeID is mandatory in CPVolume request")
```
sulakshm, line 134: nit: replace CPVolume with ControllerPublishVolume

csi/controller.go
sulakshm: This is not an error condition, it is no work and an implicit pass if the request to ControllerPublish for a non windows volumes comes through. So current code is correct.

csi/controller.go
sulakshm: All relevant checks needs to be confirmed. If the capabilities do not match, then it is best to stop servicing request much earlier.

sulakshm (approved):
apart from minor nits, I feel the changes looks good to me.

csi/controller.go
```diff
@@ -115,19 +118,122 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
 // on to a node.
 func (s *OsdCsiServer) ControllerPublishVolume(
-	context.Context,
-	*csi.ControllerPublishVolumeRequest,
+	ctx context.Context,
+	req *csi.ControllerPublishVolumeRequest,
 ) (*csi.ControllerPublishVolumeResponse, error) {
-	return nil, status.Error(codes.Unimplemented, "This request is not supported")
+
+	logrus.Debugf("Calling Controller Publish Volume in the csi controller")
```
ggriffiths, line 125: For every log in the CSI package, please use `clogger.WithContext(ctx).*` instead of `logrus.*`. This include component metadata in the logs as well as correlation ID.

csi/node.go
```diff
@@ -293,6 +293,8 @@ func (s *OsdCsiServer) NodeGetCapabilities(
 	clogger.WithContext(ctx).Debugf("csi.NodeGetCapabilities request received")
 
 	caps := []csi.NodeServiceCapability_RPC_Type{
+		// Support for Stage/Unstage of Volume on a Node.
```
ggriffiths, line 296: What is the purpose of adding NodeStage volume as a no-op? Are you depending on the kubelet or k8s controller manager to do something here?

csi/controller.go
```diff
@@ -115,19 +118,122 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
+	if err != nil {
+		if s, ok := status.FromError(err); ok && s.Code() == codes.NotFound {
+			return nil, status.Error(codes.NotFound, "Volume not found")
+		}
+		return nil, err
+	}
+
+	// have the volumeID, submit a new grpc request to px driver to prepare
+	// volume to be published only if the volume propery is winshare.
+	conn, err := s.getRemoteConn(ctx)
```
ggriffiths, line 150: This will get a remote conn on a random node in the cluster. What is the benefit of using ControllerPublishVolume if both NodePublish/ControllerPublish will pick a random node in the cluster?

The purpose of ControllerPublish is to have 2-3 nodes in a cluster that run the controller drivers. The node driver runs on every node in the cluster, as we do with Portworx currently.

csi/controller.go
```diff
@@ -115,19 +118,122 @@ func (s *OsdCsiServer) ControllerGetCapabilities(
...
+	if err != nil {
+		logrus.Errorf("failed to get remote connection: %v, continuing with local node instead", err)
```
ggriffiths, line 152: use clogger.WithContext

csi/controller.go
ggriffiths: Have you tested this? If a CSI driver supports both NodePublish and ControllerPublish, which one is called?

Also, returning a non-error will tell kubernetes that this particular PVC is mounted in the desired container. I'm not sure if that's what we want here.

csi/controller.go
ggriffiths: Yes please address this if it hasn't been already. If it has been addressed, comment here or mark resolved

api/server/sdk/volume_node_ops.go
ggriffiths: It could be argued either way whether adding NodeStageVolume would increase or decrease complexity. It doesn't make sense to add in this PR as it's a no-op in the current implementation anyway.

@sulakshm @pnookala-px as this is an initial alpha drop with limited testing, let's avoid sweeping design changes that could cause more instability for a small subset of windows use cases

ggriffiths (changes requested):
Here's my first round of reviews on the CSI side - I didn't get very far yet as I want to review your design plan (as per my slack comment). 

Please address these comments in the meantime and I'll review the design/functional spec once it's sent over.

ggriffiths:
> csi/node.go needs change to report STAGE/UNSTAGE capabilities equivalent to windows csi driver
> I am not totally sure about this, in addition also note that csi.NodeServiceCapability_RPC_GET_VOLUME_STATS, is not supported by windows csi driver.

Why do you want to add STAGE/UNSTAGE?

ggriffiths:
> Is this a supported configuration with CSI+K8s or some undocumented/unsupported mix that we should avoid.

We should avoid this, I am not sure of any use cases where another driver has done this.

AFAIK, NodePublishVolume/NodeUnpublish volume can handle all of the use cases in this PR. We should avoid adding new capabilities as it will make our driver far more complex than it needs to be.

csi/controller.go
adityadani: AFAIK, based on the capabilities returned by the driver it will invoke NodePublish and/or ControllerPublish. LNS, I see what you are saying. Yeah may be it makes sense to return no error here since we are now saying we have the ControllerPublish capability.

api/server/sdk/volume_node_ops.go
adityadani: By adding the ControllerPublish/Unpublish or NodePublish/Unpublish all the way through the Driver and filter chain, we miss out on the functionality that we have added in Attach/Mount/Detach/Unmount in all the filters. Or you will have to add the same functionality in ControllerPublish & Unpublish in the filter implementation. I think this is more convoluted and duplication of effort.  We have a tried and battle tested path of Attach/Detach/Mount/Unmount and we should continue to use that. 

IMO, the controller/nodestage publish/unpublish is a contract between Kubernetes and the CSI driver. The CSI driver then needs to do what is required to achieve the publish/unpublish operation. And that doesn't mean adding this API all through the CSI driver's stack. So I am not bought on the clean code argument here.

volume/volume.go
adityadani: If you invoke driver's Attach/Detach in the Controller Publish/Unpublish implementation then you don't need this.

sulakshm:
> NodeServiceCapability_RPC_GET_VOLUME_STATS



> > Is this a supported configuration with CSI+K8s or some undocumented/unsupported mix that we should avoid.
> 
> We should avoid this, I am not sure of any use cases where another driver has done this.
> 
> AFAIK, NodePublishVolume/NodeUnpublish volume can handle all of the use cases in this PR. We should avoid adding new capabilities as it will make our driver far more complex than it needs to be.

@ggriffiths I sense there is a disconnect in the reasoning as to why this is needed.
A functional spec and phani has already shared his reasons, but to reiterate again here:

NodePublish/et al today works for linux nodes in the cluster because there is a common storage spec that is available through kvdb at all nodes in the cluster. This is not available for the windows node.

CSI driver already allows this part of its design. And the way to use it would be to use ControllerPublish to setup on the linux node which has more intelligence and pass an additional context setup for csi driver at node end to use. 
This is precisely what is being extended in this PR to support the windows nodes in the mix.

The only item I had concerns was with differning capabilities being reported for the same csi driver pxd.portworx.com based on where the query gets made. This is so far to me an undocumented but currently working behavior. And that is why I asked you to confirm if possible on this with the community.


In any case, my best recommendation would be to keep the same capabilities being reported from all nodes for the pxd.portworx.com csi driver but take necessary actions only if necessary.

In which case, the node capabilities on the linux and windows nodes should match up always. 
That includes linux nodes extending for all new calls and windows node keeping a strict compat check with the linux driver end. I believe windows currently does not export stats but still need to export it as a capability and return all zeros for now.

csi/controller.go
sulakshm: @ggriffiths I am not entirely convinced that the changes are understood here.

ControllerPublish will get called for every volume attachment in the k8s cluster inclusive of windows nodes where px does not run. The subsequent Node{Stage|Publish}... et al will get called on the scheduled windows node.
NodePublish is current implementation for volume attachment on linux nodes. 

> The purpose of ControllerPublish is to have 2-3 nodes in a cluster that run the controller drivers. 

not necessarily true given portworx csi driver is in itself an exception. This change is only an extension to supported capability to add ControllerPublish support.

api/server/sdk/volume_node_ops.go
sulakshm: >  I think this is more convoluted and duplication of effort.
Attach and Mount calls currently are overloaded for solving many different issues, which you should be woefully aware. In the long run, I assumed we would split the functionality here. 

> We have a tried and battle tested path of Attach/Detach/Mount/Unmount and we should continue to use that.

This is definitely true; but the suggestion to split based on current complexity is already made.

> IMO, the controller/nodestage publish/unpublish is a contract between Kubernetes and the CSI driver. The CSI driver then needs to do what is required to achieve the publish/unpublish operation. And that doesn't mean adding this API all through the CSI driver's stack.

The part about CSI contract is true, but the extension to porx is debatable. If my already posted comments on Attach/Mount being overloaded with complexity does not make a convincing argument to split the function, then I have no more arguments to place.

In the meantime, I will sync up with @pnookala-px and see if changes can be limited first cut.

csi/node.go
sulakshm: The changes needs to be seen in entireity.

This is Linux side CSI driver code changes to keep compatibility with Windows side.
As all csi drivers running in the single cluster as pxd.portworx.com will have to report same capability.
For the Linux side which traditionally has not used NodeStage this is a nop.

For the windows side, which is a new csi driver by the same 'pxd.portworx.com' it is defined as below.

a) NodeStageVolume - sets up nfs client side connections and mounts it to internal px staging site and sets up client refcounting.
b) NodePublishVolume - sets up a volume access from internal staging site to pod location, and increments ref to volume.
c) NodeUnpublishVolume - removes pod setup for volume access, decrements ref to volume.
d) NodeUnstageVolume - cleans up entire nfs client and staging path if no client references to volume exist.

volume/volume.go
sulakshm: will revisit this with @pnookala-px

ggriffiths:
> NodePublish/et al today works for linux nodes in the cluster because there is a common storage spec that is available through kvdb at all nodes in the cluster. This is not available for the windows node.

NodePublishVolume is specific to CSI. It is decoupled from the internal Portworx implementation. You can have an entirely separate workflow for windows inside of NodePublishVolume instead of introducing complexity with ControllerPublish/ControllerUnpublish and NodeStage/Unstage.

> The only item I had concerns was with differning capabilities being reported for the same csi driver pxd.portworx.com based on where the query gets made. This is so far to me an undocumented but currently working behavior. And that is why I asked you to confirm if possible on this with the community.

I understand this is your only concern, but you are introducing a lot of complexity with this change that will have unintended impact on thoroughly tested workflow. The same goal of windows compatibility can be done with NodePublishVolume. If you want to check with the community on this, you check with them here:
https://communityinviter.com/apps/kubernetes/community

> In which case, the node capabilities on the linux and windows nodes should match up always.
> That includes linux nodes extending for all new calls and windows node keeping a strict compat check with the linux driver end. I believe windows currently does not export stats but still need to export it as a capability and return all zeros for now.

I don't think you understand the impact this will have on our system. If we start supporting VolumeAttachment objects, we're basically starting from scratch in terms of handling Unmount/Mount edge cases with torpedo. We'll have to start cleaning up VolumeAttachment objects in the event of a system mismatch or unexpected failure. 

Extending the CSI spec past NodePublishVolume/UnpublishVolume is a blocker for getting this PR merged.

csi/node.go
ggriffiths: Where is the code for this new CSI driver? Is it this https://github.com/portworx/windows-csi-driver?

pnookala-px:
Are you suggesting that inside Windows's CSI driver NodePublishVolume
figure out how to mount portworx volumes?
Basically, have the windows client talk to porx side of code and entirely
bypass the linux csi layer?
Are you suggesting new grpc api calls altogether?
Because the current interfaces in the csi driver are inadequate, will need
to have portworx specific values exposed?

ggriffiths:
> Basically, have the windows client talk to porx side of code and entirely
bypass the linux csi layer?

Yes, in the windows CSI driver, you could talk directly to the SDK gRPC server. You could establish the mount or remotely mount the volume on a linux node, then setup the share from the Windows node to the linux node.

This can all be accomplished w/ NodePublishVolume alone.

sulakshm:
> > Basically, have the windows client talk to porx side of code and entirely
> > bypass the linux csi layer?
> 
> Yes, in the windows CSI driver, you could talk directly to the SDK gRPC server. You could establish the mount or remotely mount the volume on a linux node, then setup the share from the Windows node to the linux node.
> 
> This can all be accomplished w/ NodePublishVolume alone.

That would imply the windows side knows where the linux cluster exists, in addition should now handle cases where the network address changes. While in the design put forth, all the information is already done by the CSI driver and information presented to windows node during attach.

Just because an idea is feasible does not mean it is optimal.

PR #2271, opened 2023-05-16
feat: create cloud backup

Glitchfix:
CSI create snapshot will now support triggering snapshot that will initiate a cloud backup.

This only support with default cloud backup credentials to be used to create the backups with the same name as the snapshot.

This PR additionally proposes to try a different way of unit test.


csi/controller.go
```diff
@@ -980,6 +1008,99 @@ func (s *OsdCsiServer) CreateSnapshot(
 		},
 	}, nil
 }
+func (s *OsdCsiServer) getCloudBackupClient() error {
+	if s.cloudBackupClient == nil {
+		// Get grpc connection
+		conn, err := s.getConn()
```
ggriffiths, line 1014: We'll need to use `getRemoteConn` to spread the load of cloud backups across all nodes. As a result, it's best not to store this client locally in OsdCsiServer.

See this line for example:
https://github.com/libopenstorage/openstorage/blob/master/csi/controller.go#L548-L557

csi/controller.go
```diff
@@ -980,6 +1008,99 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+
+	var isBackupReady bool
+	var backupStatus *api.SdkCloudBackupStatusResponse
+
+	// Check if snapshot has been created but is in error state
+	backupStatus, errFindFailed := s.cloudBackupClient.Status(ctx, &api.SdkCloudBackupStatusRequest{
+		VolumeId: req.GetSourceVolumeId(),
+		TaskId:   backupID,
+	})
+	fmt.Printf("backupStatus: %+v %v\n", backupStatus, errFindFailed)
```
ggriffiths, line 1067: remove Printf

csi/controller.go
```diff
@@ -887,6 +887,34 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	defer cancel()
+
+	// Get any labels passed in by the CO
+	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
+	}
+	fmt.Printf("locator.VolumeLabels: %s", locator.VolumeLabels)
+	// Check ID is valid with the specified volume capabilities
+	snapshotType, ok := locator.VolumeLabels["snapshotType"]
```
ggriffiths, line 902: Can you externalize this to be a constant?

Also, let's name this with our standard convention: `csi.openstoarge.org/snapshot-type`.

See this area for an example of the prefix:
https://github.com/libopenstorage/openstorage/blob/master/csi/controller.go#L42-L46

csi/controller.go
```diff
@@ -887,6 +887,34 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	if !ok {
+		snapshotType = "volume"
+	}
+	switch snapshotType {
+	case "cloud":
```
ggriffiths, line 907: Values should be `local` or `cloud`

csi/controller.go
```diff
@@ -980,6 +1008,99 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	if err != nil {
+		errStatus, ok := status.FromError(err)
+		if ok && errStatus.Code() == codes.NotFound {
+			// Return a non-final Aborted error, as the PVC may not have been created yet. Otherwise
+			// the CSI Snapshotter will not issue retries:
+			// https://github.com/kubernetes-csi/external-snapshotter/blob/v6.0.1/pkg/sidecar-controller/snapshot_controller.go#L656-L678
+			return nil, status.Errorf(codes.Aborted, "Volume id %s not found: %v", req.GetSourceVolumeId(), err)
+		}
+
+		return nil, status.Errorf(codes.Aborted, "Failed to create snapshot: %v", err)
```
ggriffiths, line 1056:
```suggestion
		return nil, status.Errorf(codes.Aborted, "Failed to create cloud snapshot: %v", err)
```

csi/controller.go
```diff
@@ -980,6 +1008,99 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	})
+
+	if err != nil {
+		errStatus, ok := status.FromError(err)
+		if ok && errStatus.Code() == codes.NotFound {
+			// Return a non-final Aborted error, as the PVC may not have been created yet. Otherwise
+			// the CSI Snapshotter will not issue retries:
+			// https://github.com/kubernetes-csi/external-snapshotter/blob/v6.0.1/pkg/sidecar-controller/snapshot_controller.go#L656-L678
+			return nil, status.Errorf(codes.Aborted, "Volume id %s not found: %v", req.GetSourceVolumeId(), err)
+		}
```
ggriffiths, line 1054: This was a special case for local snapshots. We can remove this from cloud snapshots.

csi/controller.go
```diff
@@ -980,6 +1008,99 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+
+	// Get any labels passed in by the CO
+	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
+	}
+
+	backupID := req.GetName()
+	// Create snapshot
+	_, err = s.cloudBackupClient.Create(ctx, &api.SdkCloudBackupCreateRequest{
```
ggriffiths, line 1041: If we a return a non-final gRPC error code after this line, will a subsequent s.cloudBackupClient.Create cause issues here? Did you test how that workflow would work?

csi/controller.go
```diff
@@ -980,6 +1008,99 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	fmt.Printf("backupStatus: %+v %v\n", backupStatus, errFindFailed)
+	if errFindFailed != nil {
+		return nil, status.Errorf(codes.Aborted, "Failed to create snapshot: %v", err)
+	}
+	isBackupReady = backupStatus.Statuses[backupID].Status == api.SdkCloudBackupStatusType_SdkCloudBackupStatusTypeDone ||
+		backupStatus.Statuses[backupID].Status == api.SdkCloudBackupStatusType_SdkCloudBackupStatusTypeActive
+	if errFindFailed == nil && !isBackupReady {
+
+		logrus.Infof("cleaned up failed snapshot %v in order to retry snapshot create", backupID)
+		// If snapshot was created but has errors, cleanup and re-create it on the next iteration.
```
ggriffiths, line 1076: Will this cause an infinite cycle of snapshot delete/create? We want to return a non-final error let the snapshotter retry in this case.

csi/controller_test.go
```diff
@@ -3371,3 +3376,162 @@ func TestGetCapacity(t *testing.T) {
...
+	defer ctrl.Finish()
+
+	mockCloudBackupClient := mock.NewMockOpenStorageCloudBackupClient(ctrl)
+
+	ctx := context.Background()
+
+	mockErr := errors.New("MOCK ERROR")
+	creationTime := timestamppb.Now()
+
+	mockCloudBackupClient.EXPECT().Create(gomock.Any(), gomock.Any(), gomock.Any()).
```
ggriffiths, line 3391: in our call, you mentioned this needs to be a part of the Csi struct in order for this to work.

However, we'll need to find a way to generate this client per test. The reason is that we need a new client for each remote conn we get for round-robin connections.

ggriffiths, line 3391: in our call, you mentioned this needs to be a part of the Csi struct in order for this to work.

However, we'll need to find a way to generate this client per test. The reason is that we need a new client for each remote conn we get for round-robin connections.

ggriffiths, line 3391: in our call, you mentioned this needs to be a part of the Csi struct in order for this to work.

However, we'll need to find a way to generate this client per test. The reason is that we need a new client for each remote conn we get for round-robin connections.

csi/controller.go
Glitchfix: this could happen, or it could generate redundant copies

ggriffiths:
@Glitchfix can you fix the CSI unit tests? They are failing

csi/controller.go
ggriffiths: Can you test this implementation in a live PX cluster? I think this will continuously create new snapshots on every retry if it does not exit immediately.

Glitchfix:
> @Glitchfix can you fix the CSI unit tests? They are failing

There was a typo that occurred when I change the label which is why it was failing

csi/controller.go
```diff
@@ -980,6 +1016,88 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	// Get any labels passed in by the CO
+	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetParameters())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
+	}
+
+	backupID := req.GetName()
+	// Create snapshot
+	_, err = cloudBackupClient.Create(ctx, &api.SdkCloudBackupCreateRequest{
+		VolumeId: req.GetSourceVolumeId(),
```
adityadani, line 1048: The CloudBackupCreate request also takes in "CredentialID". We should take that as an input from the user.  Similar to `snapshot-type`, lets take an input `credential-id`

csi/controller.go
```diff
@@ -980,6 +1016,88 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+	snapSize, errSizeFailed := cloudBackupClient.Size(ctx, &api.SdkCloudBackupSizeRequest{
+		BackupId: backupID,
+	})
+	if errSizeFailed != nil {
+		return nil, status.Errorf(codes.Aborted, "Failed to get cloud snapshot size: %v", err)
+	}
+
+	return &csi.CreateSnapshotResponse{
+		Snapshot: &csi.Snapshot{
+			SizeBytes:      int64(snapSize.GetSize()),
```
adityadani, line 1093: You are using a deprecated size field from the response
```
message SdkCloudBackupSizeResponse{
  // Same value as total_download_bytes, will be deprecated
  uint64 size = 1;
  // Total download size of all the backups (full & incremental) required for a restore
  uint64 total_download_bytes = 2;
  // Size of the compressed individual backup in object-store or NFS
  uint64 compressed_object_bytes = 3;
  // Used size of the volume to be restored (in bytes)
  uint64 capacity_required_for_restore = 4;
}
```
According to CSI spec, what is the expected value from the size field, since it needs to map to one of these values.

csi/controller.go
```diff
@@ -980,6 +1016,88 @@ func (s *OsdCsiServer) CreateSnapshot(
...
+			SnapshotId:     backupID,
+			SourceVolumeId: req.GetSourceVolumeId(),
+			CreationTime:   backupStatus.Statuses[backupID].StartTime,
+			ReadyToUse:     isBackupReady,
```
adityadani, line 1097: Don't we need to change this flag to `true` at some point in time? 
I think previously our capability was only for LocalSnapshots which were instantaneous. We might have to change that since cloud backups take some time. 

Check this comment on ListSnapshots
```
// ListSnapshots lists all snapshots in a cluster.
// This is mainly implemented for Nomad, as Kubernetes will not call
// list snapshots for drivers which have synchronous snapshot creation.
// This is because ReadyToUse is set to true immediately upon on our snapshot creation.
```
cc : @ggriffiths

Glitchfix: I see an implementation in list snapshot where the snapshot status is fetched first before returning it to the caller of the API.
for cloud backups, we could first set it to a pending state and fetch the status of the snaps whenever the list APIs are invoked.
I will raise that change in the list snapshot PR

csi/controller.go
adityadani: Are you planning to fix this?

csi/controller.go
adityadani: But right now if the backup status is Active you will set this to `true`. Before we merge this change we should fix that. Only set isBackupReady to `true` iff status is SdkCloudBackupStatusType_SdkCloudBackupStatusTypeDone

adityadani (commented):
We need to handle the two outstanding comments before merging this change.

ggriffiths (approved):
Lgtm!

ggriffiths:
@Glitchfix I ran travis again, and it looks like a snapshot test is failing: 
https://app.travis-ci.com/github/libopenstorage/openstorage/jobs/605718589#L1559

Can we run this test a handful of times and see if it's flakey?

Glitchfix:
> @Glitchfix I ran travis again, and it looks like a snapshot test is failing: https://app.travis-ci.com/github/libopenstorage/openstorage/jobs/605718589#L1559
> 
> Can we run this test a handful of times and see if it's flakey?

there was one deprecated test case that was being executed I think it got added back somehow while rebasing
I have removed it now

Glitchfix:
@ggriffiths we can now merge it
I don't have the write access to merge it, need help

PR #2272, opened 2023-06-01
PWX-31098: openstorage changes for enabling fastpath by default

ppadalia:
Change Summary: Volume creation has a fastpath enablement option (fastpath preferred) which specifies if fastpath is enabled for the volume or not. It can be specified during a volume create or changed later during a volume update. Fastpath enablement is currently only allowed for dmthin pools. By default, the fastpath preferred option is false (until now). This change enables fp preference by default

PWX-31098

libopenstorage changes for default fastpath enablement

[PWX 31098 UT.txt](https://github.com/libopenstorage/openstorage/files/11627119/PWX.31098.UT.txt)


sulakshm (approved):
Looks good.

ggriffiths (approved):

PR #2281, opened 2023-06-21
Adding Watch API and initial implementation

dahuang-purestorage:
This PR implments a Watch gRPC call in libopenstorage on Volumes. 
In porx, the driver is expected to implement GetVolumeWatcher which will return a go channel used in libopenstorage server to relay the volumes.
In libopenstorage, there is a list of watch connections that keeps track of the current clients. Also, there is a go routine constantly waiting from a volume from porx. Once a volume is recieved in the go channel, openstorage will relay it to the watch clients. 

Closes # PWX-29706

TODO:
- Adding Unit test
- Figure out the best place to start watcher
- Determine if it's possible to have a generic channel for additional event/resources (nodes, disks, etc). So it would be easier to add watch endpoint for them as well. 

Testing:
- Smoke testing with porx on creating, deleting and attaching volumes. Snippet of working output:
```
grpcurl -plaintext -d '{"volume_event": {"labels": [{}]}}' 0.0.0.0:9020  openstorage.api.OpenStorageWatch.Watch | jq '.volumeEvent.volume.id + " " + .volumeEvent.volume.status + " " + .volumeEvent.volume.state'
"441214862979454928 VOLUME_STATUS_UP VOLUME_STATE_DETACHED"
"27660964167843585 VOLUME_STATUS_UP VOLUME_STATE_ATTACHED"
"27660964167843585 VOLUME_STATUS_UP VOLUME_STATE_ATTACHED"
"27660964167843585 VOLUME_STATUS_UP VOLUME_STATE_DELETED"
....
```


api/api.proto
```diff
@@ -2512,6 +2512,15 @@ service OpenStorageVolume {
           body: "*"
         };
     }
+
+    // Watch streams a list of volumes exist in the system.
+    rpc Watch(SdkVolumeWatchRequest)
```
lpabon, line 2517: This is good, but I would suggest creating a model of Watch for anything on the system, not in the Volumes service. That way we can watch for Pool, Disk, Volume, or any other type of event. Then the event message could use `oneof` for each type of storage event.

lpabon (changes requested):

api/api.proto
```diff
@@ -2512,6 +2512,24 @@ service OpenStorageVolume {
...
+
+
+}
+
+
+// OpenStorageWatcher is a service used to stream objects accessible by
+// openstorage API. Right now, we will only support Volumes.
+service OpenStorageWatch{
+
+    // Watch streams of volumes exist in the system.
```
lpabon, line 2524: This comment should be changed since this is a generic event API

volume/drivers/fake/fake.go
```diff
@@ -114,11 +118,35 @@ func newFakeDriver(params map[string]string) (*driver, error) {
 			}
 		}
 	}
-
+	// go volumeGenerator(inst)
```
ggriffiths, line 121: nit: remove commented code

api/api.proto
```diff
@@ -3763,6 +3781,39 @@ message SdkVolumeSnapshotScheduleUpdateRequest {
 message SdkVolumeSnapshotScheduleUpdateResponse {
 }
 
+// Defines the request to watch a volume
+message SdkWatchRequest {
+  // one of event type
```
ggriffiths, line 3786: To expand on Luis's comments, maybe expand on some of these comments as it's autogenerated for API docs

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,195 @@
...
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const (
+	volumeEventType = "Volume"
+)
+
+type WathcerServer struct {
```
ggriffiths, line 18: typo: WathcerServer -> WatcherServer

api/api.proto
```diff
@@ -2512,6 +2512,25 @@ service OpenStorageVolume {
           body: "*"
         };
     }
+
+
+}
+
+
+// OpenStorageWatcher is a service used to stream events in the system
```
adityadani, line 2520: nit: OpenStorageWatcher is a service that provides APIs for watching on resources and receive them as a stream of events.

api/api.proto
```diff
@@ -2512,6 +2512,25 @@ service OpenStorageVolume {
...
+service OpenStorageWatch{
+
+    // Watch streams of event exist in the system
```
adityadani, line 2523: nit: Watch on resources managed by the driver and receive them as a stream of events.

api/api.proto
```diff
@@ -3763,6 +3782,42 @@ message SdkVolumeSnapshotScheduleUpdateRequest {
...
+  // examples of options are volume, node, disk, etc
+  oneof event_type {
+    SdkVolumeWatchResponse volume_event = 1;
+  }
+}
+
+// Defines the request to watch an openstorage volume event for the given label
+// if the label is empty, it returns all the volume events
+message SdkVolumeWatchRequest {
+  // Options during watch
```
adityadani, line 3807: nit: fix comment - labels to filter out the volumes to watch on

api/api.proto
```diff
@@ -2512,6 +2512,25 @@ service OpenStorageVolume {
...
+// OpenStorageWatcher is a service used to stream events in the system
+service OpenStorageWatch{
+
+    // Watch streams of event exist in the system
+    //
+    // Requires access AccessType.Read
+    rpc Watch(SdkWatchRequest)
+    returns (stream SdkWatchResponse) {
+        option(google.api.http) = {
+          post: "/v1/watch/watch"
```
adityadani, line 2529: I think this should be just `/v1/watch`. We are not watching on specific resources anymore so we don't need the extra prefix of service name anymore.

api/api.proto
```diff
@@ -3763,6 +3782,42 @@ message SdkVolumeSnapshotScheduleUpdateRequest {
...
+}
+
+// Defines the response containing an volume with a state changed
+message SdkVolumeWatchResponse {
+  // Information about the volume
+  Volume volume = 1;
+  // Name of volume
+  string name = 2;
+  // Volume labels
+  map<string, string> labels = 3;
```
adityadani, line 3818: Do we need to return the labels separately? The labels are already a part of Volume

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+				resp := convertApiVolumeToSdkReponse(vol)
+				if err := stream.Send(resp); err != nil {
+					return err
+				}
+			}
+
+			for event := range client.eventChannel {
+				var vol *api.Volume
+				if vol, ok = event.(*api.Volume); !ok {
+					continue
```
ggriffiths, line 195: Log invalid type here so we know if there's any issues here

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+// Watch streams a list of events back to clients. Events are resources that can typically fetch with openstorage API, such as Volumes,
+// Nodes, Disks, etc.
+//
+// Implementation note: the flow of data starts from porx ETCD changes, then the data will be send to openstorage via a golang channel.
+// Once the event object arrives at openstorage, it will be redistributed to a list of watch connections via another set of channels.
+func (w *WatcherServer) Watch(req *api.SdkWatchRequest, stream api.OpenStorageWatch_WatchServer) error {
+	if req.GetVolumeEvent() != nil {
+		return w.volumeWatch(req.GetVolumeEvent(), stream)
+	}
+	return nil
```
adityadani, line 37: Should we return an error here instead since we were unable to parse the request and/or the object is unsupported?

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+
+func (s *WatcherServer) startVolumeWatcher(ctx context.Context, done chan bool) error {
+	if s.watchConnections == nil {
+		s.watchConnections = make(map[string][]*watchConnection)
+	}
+
+	// wait for driver to be initialized with an non-empty volume watcher
+	for {
+		if s.volumeServer.driver(ctx) == nil {
+			continue
```
adityadani, line 115: This will spin continuously if invoked before driver is initialized. Add a sleep may be?

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+
+		volumeChannel, _ := s.volumeServer.driver(ctx).GetVolumeWatcher(&api.VolumeLocator{}, make(map[string]string))
+		if volumeChannel == nil {
+			continue
```
adityadani, line 120: same here. Also I am thinking does make since to start the volume watcher when we get the first Watch API call instead of doing it while setting the gRPC server? 

In this way in the grpc Watch handler on line 34, you will check if watcher has been started if not, you start it and there you can do all these checks without requiring a for loop. If any of this fails, you return back an error to the caller. Also if the watcher is already setup, then you use the existing watcher.

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+	if ok {
+		// create a new context that will get done on deadline
+		ctx, cancel = context.WithTimeout(ctx, time.Until(deadline))
+		defer cancel()
+	}
+
+	client := watchConnection{
+		name:         uuid.New(),
+		eventType:    volumeEventType,
+		eventChannel: make(chan interface{}, 2),
```
adityadani, line 165: Why create a buffer of only 2? This could lead to issues where for some reason the channel is full (say the stream is blocked) and we keep spinning up go routines on line 132 as part of callback invocation. 

It would be better to have an unbounded channel, so that we keep queueing in the events, and avoid goroutines waiting on sending on this channel.

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+	})
+
+	// wait for err-group processes to be done
+	go func() {
+		errChan <- group.Wait()
+	}()
+
+	// wait only as long as context deadline allows
+	select {
+	case err := <-errChan:
```
adityadani, line 217: Why not just do this here?
```suggestion
	case err := <-group.Wait():
```

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+		s.watchConnections = make(map[string][]*watchConnection)
+	}
+
+	// wait for driver to be initialized with an non-empty volume watcher
+	for {
+		if s.volumeServer.driver(ctx) == nil {
+			continue
+		}
+
+		volumeChannel, _ := s.volumeServer.driver(ctx).GetVolumeWatcher(&api.VolumeLocator{}, make(map[string]string))
```
ggriffiths, line 118: Should we check the error here?

api/server/sdk/watcher.go
adityadani: I was referring to an unbounded channel and not unbuffered. Essentially just an unbounded queue that will keep collecting the events.

api/api.proto
dahuang-purestorage: there could be additional labels that were not part of the request

volume/volume.go
```diff
@@ -308,6 +309,12 @@ type Enumerator interface {
 	SnapEnumerate(volID []string, snapLabels map[string]string) ([]*api.Volume, error)
 }
 
+// Water provides a set of function to get volume
+type Watcher interface {
+	// Gets Volume notifier
+	GetVolumeWatcher(locator *api.VolumeLocator, labels map[string]string) (chan *api.Volume, error)
```
dahuang-purestorage, line 315: As discussed with Aditya offline, this might not be necessary. Looking into ways to remove this function.

api/server/sdk/watcher.go
dahuang-purestorage: we could but we will need a for loop to wait for it. group.Wait() returns an error instead of a channel.

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,242 @@
...
+			for event := range client.eventChannel {
+				var vol *api.Volume
+				if vol, ok = event.(*api.Volume); !ok {
+					continue
+				}
+				if !vol.IsPermitted(ctx, api.Ownership_Read) {
+					continue
+				}
+				resp := convertApiVolumeToSdkReponse(vol)
+				if err := stream.Send(resp); err != nil {
```
dahuang-purestorage, line 201: Figure out if we can set timeout whenever stream.Send fails

api/server/sdk/watcher.go
dahuang-purestorage: I'm not sure if that's the way to go. Since we have 2 grpcServer running, we might accidentally spin up two watcher. This goroutine should be stuck at line 129 once protodriver is initialize, which shouldn't take too long.

volume/volume.go
dahuang-purestorage: First pass on the code, I think it will be really hard to manage the channel passing from top to sdkServer and protoDriver. 

Another way to send the value is through channel, but it becomes hard to manage too when we throw business logic into contexts. 

@adityadani let me know what you think.

lpabon (commented):
👍🏼 I approve of of the changes to the API. I'll leave the implementation checks to Grant and Aditya.

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,264 @@
...
+					return err
+				}
+			}
+
+			for event := range client.eventChannel {
+
+				// create a new context that will return error if execution took more than streamTimeout
+				timeoutCtx, timeoutCancelled := context.WithTimeout(ctx, streamTimeout)
```
adityadani, line 204: I am not following how this timeoutCtx is being enforced? Isn't steam.Send() a blocking function?

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,264 @@
...
+
+				resp := convertApiVolumeToSdkReponse(vol)
+				err := stream.Send(resp)
+
+				if err != nil {
+					logrus.Warnf("error sending stream: %v", err)
+					return err
+				}
+				if timeoutCtx.Err() != nil {
+					logrus.Warnf("context error: %v", timeoutCtx.Err())
```
adityadani, line 223: nit: add more info in this log line

api/server/sdk/watcher.go
adityadani: Please add a sleep here before merging the PR

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,264 @@
...
+	for {
+		select {
+		case vol := <-s.volumeChannel:
+			s.RLock()
+			for _, client := range s.watchConnections[volumeEventType] {
+				go client.callBack(vol)
+			}
+			s.RUnlock()
+		case <-ctx.Done():
+			logrus.Infof("exiting volume watcher\n")
```
adityadani, line 145: Don't need the `\n`. Also can you change logs to start with uppercase letters?

api/server/sdk/watcher.go
```diff
@@ -0,0 +1,264 @@
...
+				codes.Internal,
+				"Failed to enumerate volumes: %v",
+				err.Error())
+		} else {
+			for _, vol := range vols {
+				if !vol.IsPermitted(ctx, api.Ownership_Read) {
+					continue
+				}
+				resp := convertApiVolumeToSdkReponse(vol)
+				if err := stream.Send(resp); err != nil {
```
adityadani, line 196: We need the timeoutCtx here too.

volume/volume.go
adityadani: Lets keep this interface and function in that case.

adityadani (commented):
overall looks good. I have some pending questions on the PR. Also do you think we should write client stubs for the streaming API as well, so that the callers can simply use it w/ the correct server keepalive timeouts.

api/server/sdk/watcher.go
```diff
@@ -40,7 +40,7 @@ func (w *WatcherServer) Watch(req *api.SdkWatchRequest, stream api.OpenStorageWa
 	if req.GetVolumeEvent() != nil {
 		return w.volumeWatch(req.GetVolumeEvent(), stream)
 	}
-	return status.Errorf(codes.InvalidArgument, "invalid request type for watcher %v", req)
+	return status.Errorf(codes.InvalidArgument, "Invalid request type for watcher %v", req)
```
adityadani, line 43: nit: errors should still be lower case

adityadani (approved):
lgtm

ggriffiths (approved):

PR #2293, opened 2023-07-11
PWX-30476: Upgrade GetMounts call to use the latest

pnookala-px:

Closes #


ggriffiths (changes requested):
I don't think this is still required based on the design changes, but if it is then please re-request review

ggriffiths (approved):

PR #2297, opened 2023-07-19
PWX-32459: Add missing REST handlers to openstorage SDK server

ggriffiths:
* Add missing REST handlers to openstorage SDK server 
* Adds a linter

Closes #

### DIFF
--- a/hack/check-registered-rest.sh
+++ b/hack/check-registered-rest.sh
@@ -1,0 +1,16 @@
+set -x 
+
+#!/bin/bash
+fail() {
+	echo "$1"
+	exit 1
+}
+
+registerFuncs=$(cat api/api.pb.go | grep 'func Register.*' | wc -l)
+registered=$(cat api/server/sdk/rest_gateway.go | grep 'Register.*Handler,' | wc -l)
+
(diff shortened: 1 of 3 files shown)


hack/check-registered-rest.sh
```diff
@@ -0,0 +1,16 @@
+set -x 
+
+#!/bin/bash
+fail() {
+	echo "$1"
+	exit 1
+}
+
+registerFuncs=$(cat api/api.pb.go | grep 'func Register.*' | wc -l)
+registered=$(cat api/server/sdk/rest_gateway.go | grep 'Register.*Handler,' | wc -l)
```
lpabon, line 10: Love this!

lpabon (approved):

hack/check-registered-rest.sh
zoxpx: Uh.. I would have used `egrep` or `grep -E` -- since I'm not sure "regular grep" will normally respect `.*`-regexp

Also, could reduce 3 commands into 1 ;) --> `grep -E 'Register.*Handler,' api/server/sdk/rest_gateway.go`

PR #2300, opened 2023-07-24
Windows share support without using controller publish

pnookala-px:

Closes #


ggriffiths (approved):
The changes to the API looks good

api/spec/spec_handler.go
```diff
@@ -606,6 +607,12 @@ func (d *specHandler) UpdateSpecFromOpts(opts map[string]string, spec *api.Volum
 			} else {
 				spec.FpPreference = fastpath
 			}
+		case api.SpecWinshare:
+			if winshare, err := strconv.ParseBool(v); err != nil {
+				return nil, nil, nil, err
+			} else {
+				spec.Winshare = winshare
+			}
```
sulakshm, line 615: else is not needed after return.

sulakshm (approved):
why isnt the changes still not merged?

api/spec/spec_handler.go
pnookala-px: but that is how code is written in all places IN THIS FUNCTION.

PR #2304, opened 2023-07-28
feat: delete cloud backup handler

Glitchfix:
Delete snapshots created for cloud backup

the deletion would require passing in the credential id in the secrets parameter along with the cloud backup id as the snapshot id

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -22,6 +22,7 @@ import (
 	"math"
 	"reflect"
 	"sort"
+	"strings"
 
 	"github.com/portworx/kvdb"
 	"github.com/sirupsen/logrus"
@@ -1092,13 +1093,7 @@ func (s *OsdCsiServer) DeleteSnapshot(
 		return nil, status.Error(codes.InvalidArgument, "Snapshot id must be provided")
 	}
 
-	// Get any labels passed in by the CO
-	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetSecrets())
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
-	}
-
-	_, ok := locator.VolumeLabels[osdSnapshotCredentialIDKey]
+	_, ok := req.GetSecrets()[osdSnapshotCredentialIDKey]
 	// Check ID is valid with the specified volume capabilities
 	snapshotType := DriverTypeCloud
 	if !ok {
@@ -1155,13 +1150,7 @@ func (s *OsdCsiServer) deleteCloudBackup(
 		return nil, err
 	}
 
-	// Get any labels passed in by the CO
-	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetSecrets())
-	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
-	}
-
-	credentialID := locator.VolumeLabels[osdSnapshotCredentialIDKey]
+	credentialID := req.GetSecrets()[osdSnapshotCredentialIDKey]
 
 	backupID := req.GetSnapshotId()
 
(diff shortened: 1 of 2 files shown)


csi/controller.go
```diff
@@ -1107,6 +1133,41 @@ func (s *OsdCsiServer) DeleteSnapshot(
...
+	ctx context.Context,
+	req *csi.DeleteSnapshotRequest,
+) (*csi.DeleteSnapshotResponse, error) {
+	cloudBackupClient, err := s.getCloudBackupClient(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get any labels passed in by the CO
+	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetSecrets())
```
ggriffiths, line 1146: Can we pass locator that we already got on line 1088? Instead of fetching a new copy

csi/controller.go
```diff
@@ -1107,6 +1133,41 @@ func (s *OsdCsiServer) DeleteSnapshot(
...
+	credentialID := locator.VolumeLabels[osdSnapshotCredentialIDKey]
+
+	backupID := req.GetSnapshotId()
+
+	// Delete snapshot
+	resp, err := cloudBackupClient.Delete(ctx, &api.SdkCloudBackupDeleteRequest{
+		BackupId:     backupID,
+		CredentialId: credentialID,
+	})
+	if err != nil {
```
ggriffiths, line 1160: Check for NotFound error here and return `return &csi.DeleteSnapshotResponse{}, nil` in case the underlying snapshot was deleted.

csi/controller.go
```diff
@@ -1107,6 +1133,41 @@ func (s *OsdCsiServer) DeleteSnapshot(
...
+		return nil, status.Errorf(codes.Aborted, "Failed to delete cloud snapshot: %v", err)
+	}
+
+	if resp != nil {
+		return &csi.DeleteSnapshotResponse{}, nil
+	}
+
+	return nil, status.Errorf(codes.Aborted, "unexpected error occured: %v", err)
```
ggriffiths, line 1168: This can just be simplified to `return nil`.

If we didn't get an error on line 1160, we know deleted succeeded and `err` will be nil.

csi/controller_test.go
```diff
@@ -3563,3 +3563,133 @@ func TestOsdCsiServer_CreateSnapshot(t *testing.T) {
...
+		},
+		{
+			"fail snapshot delete",
+			"delete-error",
+			"valid-cred",
+			nil,
+			true,
+		},
+		{
+			"volume id not found while creating",
```
ggriffiths, line 3639: should say while deleting

csi/controller_test.go
```diff
@@ -3563,3 +3563,133 @@ func TestOsdCsiServer_CreateSnapshot(t *testing.T) {
...
+			"delete-notfound",
+			"valid-cred",
+			nil,
+			true,
```
ggriffiths, line 3643: This should not return an error

csi/controller.go
ggriffiths: Can we check similar to this line?
https://github.com/libopenstorage/openstorage/blob/master/csi/controller.go#L154

We want to check the actual status code

ggriffiths: String check is error prone, we want to check for actual status code.

ggriffiths (changes requested):
A few changes still required. @Glitchfix can you also post testing output when you get a chance?

Glitchfix:
> A few changes still required. @Glitchfix can you also post testing output when you get a chance? Please re-request review here when ready

```
go test -timeout 30s -run ^TestOsdCsiServer_listCloudSnapshots$ github.com/libopenstorage/openstorage/csi
ok      github.com/libopenstorage/openstorage/csi       
```

ggriffiths:
@Glitchfix I mean integration testing - have we built a PX container with above changes and tried creating/deleting cloud snapshots?

PR #2305, opened 2023-07-28
feat: list cloud backup handler

Glitchfix:
list snapshots created for cloud backup

this feature would enumerate the cloud backups and will require passing in the credential id in the secrets parameter

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -1128,6 +1128,36 @@ func (s *OsdCsiServer) ListSnapshots(
 	ctx context.Context,
 	req *csi.ListSnapshotsRequest,
 ) (*csi.ListSnapshotsResponse, error) {
+	// Get any labels passed in by the CO
+	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetSecrets())
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Unable to get parameters: %v", err)
+	}
+
+	_, ok := locator.VolumeLabels[osdSnapshotCredentialIDKey]
+	// Check ID is valid with the specified volume capabilities
+	snapshotType := DriverTypeCloud
+	if !ok {
+		snapshotType = DriverTypeLocal
+	}
+	switch snapshotType {
+	case DriverTypeCloud:
+		return s.listCloudSnapshots(ctx, req)
+	case DriverTypeLocal:
+		fallthrough
+	default:
+		return s.listLocalSnapshots(ctx, req)
+	}
+}
+
+// ListSnapshots lists all snapshots in a cluster.
+// This is mainly implemented for Nomad, as Kubernetes will not call
+// list snapshots for drivers which have synchronous snapshot creation.
+// This is because ReadyToUse is set to true immediately upon on our snapshot creation.
+func (s *OsdCsiServer) listLocalSnapshots(
+	ctx context.Context,
+	req *csi.ListSnapshotsRequest,
+) (*csi.ListSnapshotsResponse, error) {
 
 	if len(req.GetSnapshotId()) > 0 {
 		return s.listSingleSnapshot(ctx, req)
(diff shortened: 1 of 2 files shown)


csi/controller.go
```diff
@@ -1115,6 +1115,36 @@ func (s *OsdCsiServer) ListSnapshots(
 	ctx context.Context,
 	req *csi.ListSnapshotsRequest,
 ) (*csi.ListSnapshotsResponse, error) {
+	// Get any labels passed in by the CO
+	_, locator, _, err := s.specHandler.SpecFromOpts(req.GetSecrets())
```
adityadani, line 1119: Why do we need to call `specHandler.SpecFromOpts` here? Why not just look at the map returned by `req.GetSecrets()` and check if it contains `osdSnapshotCredentialIDKey` ?

csi/controller.go
```diff
@@ -1287,6 +1317,76 @@ func (s *OsdCsiServer) listMultipleSnapshots(
...
+	}
+
+	for i, backup := range resp.Backups {
+		// Check if snapshot has been created but is in error state
+		backupStatus, errFindFailed := cloudBackupClient.Status(ctx, &api.SdkCloudBackupStatusRequest{
+			VolumeId: req.GetSourceVolumeId(),
+			TaskId:   backup.GetId(),
+		})
+		if errFindFailed != nil {
+			return nil, status.Errorf(codes.Aborted, "Failed to get cloud snapshot status: %v", err)
```
adityadani, line 1364: nit: errors should start with lower case

csi/controller.go
```diff
@@ -1287,6 +1317,76 @@ func (s *OsdCsiServer) listMultipleSnapshots(
...
+		}
+		isBackupReady := backupStatus.Statuses[backup.GetId()].Status == api.SdkCloudBackupStatusType_SdkCloudBackupStatusTypeDone
+
+		snapSize, errSizeFailed := cloudBackupClient.Size(ctx, &api.SdkCloudBackupSizeRequest{
+			BackupId:     backup.GetId(),
+			CredentialId: credentialID,
+		})
+		if errSizeFailed != nil {
+			return nil, status.Errorf(codes.Aborted, "Failed to get cloud snapshot size: %v", err)
```
adityadani, line 1373: nit: errors should start with lower case

csi/controller.go
```diff
@@ -1128,6 +1128,32 @@ func (s *OsdCsiServer) ListSnapshots(
...
+		fallthrough
+	default:
+		return s.listLocalSnapshots(ctx, req)
+	}
+}
+
+// ListSnapshots lists all snapshots in a cluster.
+// This is mainly implemented for Nomad, as Kubernetes will not call
+// list snapshots for drivers which have synchronous snapshot creation.
+// This is because ReadyToUse is set to true immediately upon on our snapshot creation.
```
ggriffiths, line 1152: can we update this comment? Looks like it was re-used from `ListSnapshots`

csi/controller.go
```diff
@@ -1128,6 +1128,32 @@ func (s *OsdCsiServer) ListSnapshots(
 	ctx context.Context,
 	req *csi.ListSnapshotsRequest,
 ) (*csi.ListSnapshotsResponse, error) {
+	// Get any labels passed in by the CO
+
+	_, ok := req.GetSecrets()[osdSnapshotCredentialIDKey]
```
ggriffiths, line 1133: It is possible to pass this value manually via Nomad:
https://developer.hashicorp.com/nomad/docs/commands/volume/snapshot-list

With k8s, I check and this should still work. CSI will find the snapshot class for a given snapshot, and pass the associated secrets into our ListSnapshots call:
https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/sidecar-controller/snapshot_controller.go#L258-L266

csi/controller.go
```diff
@@ -1300,6 +1326,70 @@ func (s *OsdCsiServer) listMultipleSnapshots(
...
+	}
+	resp, err := cloudBackupClient.EnumerateWithFilters(ctx, &api.SdkCloudBackupEnumerateWithFiltersRequest{
+		SrcVolumeId:       req.GetSourceVolumeId(),
+		CloudBackupId:     req.GetSnapshotId(),
+		CredentialId:      credentialID,
+		ContinuationToken: req.GetStartingToken(),
+		MaxBackups:        uint64(maxBackups),
+	})
+	if nil != err {
+		return nil, status.Errorf(codes.InvalidArgument, "unable to list cloud backups: %v", err)
```
ggriffiths, line 1353: This should be a codes.Internal, not invalid argument

csi/controller.go
```diff
@@ -1300,6 +1326,70 @@ func (s *OsdCsiServer) listMultipleSnapshots(
...
+	}
+
+	csiSnapshotResp := &csi.ListSnapshotsResponse{
```
ggriffiths, line 1356: Can you check if cloudBackupClient.EnumerateWithFilters returns a continuation token? Since we are passing in the ContinuationToken above, I'm guessing it does.

If so, we'll want to populate this `next_token` field in the response:
https://github.com/container-storage-interface/spec/blob/master/csi.proto#L1334

csi/controller.go
```diff
@@ -1300,6 +1326,70 @@ func (s *OsdCsiServer) listMultipleSnapshots(
...
+		Entries: make([]*csi.ListSnapshotsResponse_Entry, len(resp.Backups)),
+	}
+
+	for i, backup := range resp.Backups {
+		// Check if snapshot has been created but is in error state
+		backupStatus, errFindFailed := cloudBackupClient.Status(ctx, &api.SdkCloudBackupStatusRequest{
```
ggriffiths, line 1362: EnumerateWithFilters returns the status as part of the response:
https://github.com/libopenstorage/openstorage/blob/master/api/api.proto#L4727

Can we remove this per-backup status call and instead depend on the single enumerate with filters response?

ggriffiths (changes requested):
Some changes needed here, also can you send testing output here in the PR or via slack? Be sure to avoid any credentials of course.

PR #2329, opened 2023-08-24
fix: don't default to sharedv4/shared for RWX Volumes if proxy spec is set

Glitchfix:
When creating a proxy volume with CSI enabled, if you set RWX access mode, the user will hit an error. This is because we default to sharedv4 for all RWX volumes in CSI.

When defaulting to sharedv4, we need to first check if a proxy spec is set. In that case, we shouldn’t default to sharedv4.


csi/controller_test.go
```diff
@@ -3289,6 +3289,30 @@ func TestResolveSpecFromCSI(t *testing.T) {
...
+				},
+			},
+			existingSpec: &api.VolumeSpec{
+				ProxySpec: &api.ProxySpec{
+					ProxyProtocol: api.ProxyProtocol_PROXY_PROTOCOL_NFS,
+				},
+			},
+
+			expectedSpec: &api.VolumeSpec{
+				Shared: false,
```
ggriffiths, line 3310: Can you check `Sharedv4: false`, as well?

ggriffiths (approved):
LGTM! Thanks @Glitchfix

Glitchfix:
@ggriffiths I think you need merge the PR
Only those with write access to this repository can merge pull requests.

PR #2330, opened 2023-08-30
PWX-33379

alicelyu-px:
Needed by PWX-32897

https://github.com/portworx/porx/pull/11981/files


api/api.proto
```diff
@@ -234,6 +234,8 @@ message StorageResource {
   bool cache = 14;
   // True if the resource is used as thin pool metadata disk
   bool pool_metadata_dev = 15;
+  // Cloud drive type
+  string cloud_drive_type = 16;
```
ggriffiths, line 238: Can you bump the minor version in this file? Here is a previous PR as an example:
https://github.com/libopenstorage/openstorage/pull/2295/files#diff-1b40cafcd4234784bd31e2b14cb42a19926bd44764d27b4e12f57bc192b46a52L5438

Also, can you add an entry in SDK_CHANGELOG at the project root?

adityadani (approved):
lgtm

PR #2344, opened 2023-09-20
Pwx30476merge

pnookala-px:
PWX-30476: pick changes from master to release-9.7 branch.


ggriffiths (approved):
Please make sure to merge without squashing so we can maintain individual commits when cherry-picking. 

Otherwise looks good.

PR #2351, opened 2023-09-27
fix: chattr exits with status 1 for files that are not permitted without force flag

Glitchfix:
chattr exits with status 1 for files that are not permitted without force flag

[PWX-32298](https://portworx.atlassian.net/browse/PWX-32298)
It was obsereved that the package was throwing error despite of setting the immutability flags properly.
```sh
 Operation not supported while reading flags on /var/lib/kubelet/plugins_registry/pxd.portworx.com-reg.sock
```

**Testing Notes**  
Use the package to set immutability flags on node registration socket, 
use the `ForceAddImmutable` method to set the immutability flags 
it can be compared to `AddImmutable` that `ForceAddImmutable` does set the flags successfully

[PWX-32298]: https://portworx.atlassian.net/browse/PWX-32298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/pkg/chattr/chattr.go
+++ b/pkg/chattr/chattr.go
@@ -17,13 +17,32 @@ const (
 )
 
 func AddImmutable(path string) error {
+	return ForceAddImmutable(path, false)
+}
+
+func ForceAddImmutable(path string, force bool) error {
 	chattrBin := which(chattrCmd)
 	if _, err := os.Stat(path); err == nil {
-		cmd := exec.Command(chattrBin, "+i", path)
+		var cmd *exec.Cmd
+		if force {
+			cmd = exec.Command(chattrBin, "+i", "-f", path)
+		} else {
+			cmd = exec.Command(chattrBin, "+i", path)
+		}
 		var stderr bytes.Buffer
 		cmd.Stderr = &stderr
 
-		if err = cmd.Run(); err != nil {
+		err = cmd.Run()
+		if err != nil {
+			// chattr exits with status 1 for files that are not permitted without force flag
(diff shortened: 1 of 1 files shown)


pkg/chattr/chattr.go
```diff
@@ -16,34 +16,74 @@ const (
...
 
-	return nil
+	return err
 }
 
-func RemoveImmutable(path string) error {
+func RemoveImmutable(path string) (err error) {
+	return ForceRemoveImmutable(path, false)
+}
+func ForceRemoveImmutable(path string, force bool) (err error) {
```
ggriffiths, line 58: Same comments as above for `ForceAddImmutable` as the functions are essentially the same with `+i/-i`

ggriffiths (changes requested):
Let's try to avoid changing the functionality of code when not needed. This `chattr` package is called on every Mount/Unmount for every volume. So, changing the contract of how we return errors can lead to bugs.

ggriffiths:
@Glitchfix can you fix the travis failure?
```
git-validation -run DCO,short-subject
 * 9a01976 "refactor: [minor refactor] remove extra line and return nil to not break existing mount functionalities" ... FAIL
  - FAIL - commit subject exceeds 90 characters
 ```

ggriffiths (approved):

ggriffiths (approved):

PR #2356, opened 2023-10-09
Add volume id, gotten from mount request, to the mountpath

vlaxman-px:
The volume ID that we get from the http request to mount, is added to the mountpath to make it unique so that volume does not unmount from other pods to which it is attached.

Closes #
or
PWX-33076

### DIFF
--- a/api/server/docker.go
+++ b/api/server/docker.go
@@ -318,8 +318,8 @@ func (d *driver) status(w http.ResponseWriter, r *http.Request) {
 	io.WriteString(w, fmt.Sprintln("osd plugin", d.version))
 }
 
-func (d *driver) mountpath(name string) string {
-	return path.Join(volume.MountBase, name)
+func (d *driver) mountpath(nameWithID string) string {
+	return path.Join(volume.MountBase, nameWithID)
 }
 
 func (d *driver) getConn() (*grpc.ClientConn, error) {
@@ -651,6 +651,8 @@ func (d *driver) mount(w http.ResponseWriter, r *http.Request) {
 
 	// get spec and name from request
 	_, spec, _, _, name := d.SpecFromString(request.Name)
+	_, _, _, _, volID := d.SpecFromString(request.ID)
+	nameWithID := name + volID
 	attachOptions := d.attachOptionsFromSpec(spec)
 
 	// attach token in context metadata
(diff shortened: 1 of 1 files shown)


api/server/docker.go
```diff
@@ -651,6 +651,8 @@ func (d *driver) mount(w http.ResponseWriter, r *http.Request) {
 
 	// get spec and name from request
 	_, spec, _, _, name := d.SpecFromString(request.Name)
+	_, _, _, _, volID := d.SpecFromString(request.ID)
```
adityadani, line 654: You don't need this. Just use the provided ID as is.

api/server/docker.go
```diff
@@ -651,6 +651,8 @@ func (d *driver) mount(w http.ResponseWriter, r *http.Request) {
...
+	nameWithID := name + volID
```
adityadani, line 655:
```suggestion
	nameWithID := name + request.ID
```

api/server/docker.go
```diff
@@ -857,14 +859,17 @@ func (d *driver) unmount(w http.ResponseWriter, r *http.Request) {
 	}
 
 	_, _, _, _, name := d.SpecFromString(request.Name)
+	_, _, _, _, volID := d.SpecFromString(request.ID)
```
adityadani, line 862: same here

api/server/docker.go
```diff
@@ -882,6 +887,15 @@ func (d *driver) unmount(w http.ResponseWriter, r *http.Request) {
 	opts := make(map[string]string)
 	opts[options.OptionsDeleteAfterUnmount] = "true"
 
+	errWithoutID := v.Unmount(correlation.TODO(), id, mountpointWithoutID, opts)
```
adityadani, line 890: Can you add a comment here explaining why we are doing this double Unmount?

adityadani (changes requested):

api/server/docker.go
vlaxman-px: OK, I'll change that

api/server/docker.go
vlaxman-px: Okay

api/server/docker.go
vlaxman-px: Sure, will do that

adityadani (approved):

PR #2358, opened 2023-10-11
PWX-34326: CSI - fix GetPluginInfo version for CSI

ggriffiths:
Nomad shows this version in the CLI output, so we need to have the correct version exposed here via GetPluginInfo

PWX-34326

**Testing Notes**  
Not needed

n/a

### DIFF
--- a/csi/identity.go
+++ b/csi/identity.go
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	// CSI 1.4 compatible
-	csiDriverVersion = "1.4.0"
+	// CSI 1.7 compatible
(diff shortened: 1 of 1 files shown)


csi/identity.go
```diff
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	// CSI 1.4 compatible
-	csiDriverVersion = "1.4.0"
```
ggriffiths, line 28: Nomad shows this version in the CLI output, so we need to have the correct version exposed here via GetPluginInfo

pp511 (approved):
lgtm

PR #2367, opened 2023-10-25
PWX-28654 Add quorum member flag to the StorageNode SDK object

piyush-nimbalkar:
This field is populated from cluster database's NonQuorumMember field. Which in turn is calculated in QuorumMember() listener function impl in PX. It looks at the storage devices in the metadata store.

Closes #
or
[PWX-28654](https://portworx.atlassian.net/browse/PWX-28654)

**Testing Notes**  
- Modified UTs and they are passing.
- Validated the output of the new flag in all the APIs Inspect, InspectCurrent, EnumerateWithFilters. 
- Also, validated the flag at various stages of portworx bootup.

Sample output from a 6 node cluster with 3 storage nodes:
```
# curl -s localhost:9021/v1/nodes/filters | jq '.nodes[].non_quorum_member'
true
true
false
false
true
false
```

```
# curl -s localhost:9021/v1/nodes/inspect/74809bde-d9b6-4cdd-bcfe-337a0e215f52 | jq '.node.non_quorum_member'
true
```
```
# curl -s localhost:9021/v1/nodes/inspectcurrent | jq '.node.non_quorum_member'
true
```

[PWX-28654]: https://portworx.atlassian.net/browse/PWX-28654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/api.proto
```diff
@@ -1622,6 +1622,11 @@ message StorageNode {
   SecurityStatus security_status = 19;
   // Topology information of the node in scheduler context
   SchedulerTopology scheduler_topology = 20;
+  // Flag indicating whether the node is a quorum member or not. Using inverse value
+  // to handle intialization and upgrades. Node will always be counted as a quorum member
+  // when initializing until it reaches a point where we can definitely determine whether
+  // it is a quorum member or not.
+  bool non_quorum_member = 22;
```
pureneelesh, line 1629: why are we skipping 21?

cluster/manager/manager.go
```diff
@@ -746,6 +748,7 @@ func (c *ClusterManager) joinCluster(
 	selfNodeEntry.NonQuorumMember =
 		selfNodeEntry.Status == api.Status_STATUS_DECOMMISSION ||
 			!c.quorumMember()
+	c.selfNode.NonQuorumMember = selfNodeEntry.NonQuorumMember
```
pureneelesh, line 751: You probably checked it already but wondering if should set NonQuorumMember on line 597 in initNode() as well.

pureneelesh (approved):

api/api.proto
piyush-nimbalkar: 21 is defined above for `cpu_cores`.

cluster/manager/manager.go
piyush-nimbalkar: `initNode()` is called before the listeners are initialized. The caller of the `initNode()` is [initListeners()](https://github.com/libopenstorage/openstorage/blob/master/cluster/manager/manager.go#L1216), which does things in the following order - 
```
initNode()   - create a node object
initNodeInCluster() - initializes all the listeners
quorumMember() - this calls QuorumMember() on the now initialized PX listener
```
In this PR we are setting the quorum member flag after the above steps. We set it in both the selfNode and the node entry that will get stored in the cluster database.

piyush-nimbalkar:
Thank you for the quick review, @pureneelesh !

PR #2373, opened 2023-11-09
[cherry-pick] Don't default to sharedv4/shared for RWX Volumes if proxy spec is set

Glitchfix:
Don't default to sharedv4/shared for RWX Volumes if proxy spec is set

[PWX-33008](https://portworx.atlassian.net/browse/PWX-33008)

**Testing Notes**  
Add a proxy spec
RWX or a similar multi-node access mode the volume shouldn't default to sharedv4 if the proxy spec is defined

[PWX-33008]: https://portworx.atlassian.net/browse/PWX-33008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -783,6 +783,11 @@ func resolveSharedSpec(spec *api.VolumeSpec, req *csi.CreateVolumeRequest) (*api
 		return spec, nil
 	}
 
+	// don't default to sharedv4/shared for RWX Volumes if proxy spec is set
+	if spec.ProxySpec != nil {
+		return spec, nil
+	}
+
 	var shared bool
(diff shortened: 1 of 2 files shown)


pnookala-px (approved):
Have we tested compatibility before and after an upgrade? If tested with pv prior to upgrade and the behavior is not broken, we are good to go.

Glitchfix:
> Have we tested compatibility before and after an upgrade? If tested with pv prior to upgrade and the behavior is not broken, we are good to go.

this will not have any impact on compatibility or upgrades

PR #2375, opened 2023-11-14
Add a new type of Job: DefragJob

twang-ps:
This PR added DefragJob proto message under the Job data structure. It defines a defragmentation background job running on a cluster.

**Testing Notes**  
Add testing output or passing unit test output here.


api/api.proto
```diff
@@ -4106,6 +4109,16 @@ message CollectDiagsJob {
   repeated DiagsCollectionStatus statuses = 2;
 }
 
+message DefragJob {
+  double max_duration_hours = 1;
+  uint32 max_nodes_in_parallel = 2;
+  repeated string nodes_select = 3;
```
sangleganesh, line 4115: "nodes_select" and "node_selector" sound similar - can you rename ?
also add comments to the fields ?
do we need the schedule_id field in the job instance itself ?

twang-ps: Will change `nodes_select` to `select_nodes`; change `node_selector` to `node_label_selector`
Will add comments.
The `schedule_id` is needed here, because we want to keep track of on which nodes did the defrag schedule stop each time and resume from there next time. We will need to store this information under schedule data structure. 
Maybe the other way also works: put the job_id into schedule data structure, and do a EnumerateSelect.

api/api.proto
```diff
@@ -4106,6 +4109,30 @@ message CollectDiagsJob {
...
+  // ExcludeNodes is a list of node UUID: if provided, the job will skip these nodes;
+  // if not provided, will run on all nodes
+  // cannot coexist with IncludeNodes
+  repeated string exclude_nodes = 4;
+  // NodeSelector is a list of node label `key=value` pairs separated by comma, 
+  // which selects the nodes to be run on for the job
+  // can coexist with ExcludeNodes but cannot coexist with IncludeNodes
+  repeated string node_selector = 5;
+  // RunningNodes stores the nodes on which the job is currently running
+  repeated string running_nodes = 6;
```
adikulkarni0, line 4131: for more context I would name this `current_running_nodes`.

adikulkarni0 (approved):

api/api.proto
twang-ps: done

PR #2382, opened 2023-12-07
fix: rebuild connection cache on timeout

Glitchfix:
When creating the http client we cache the client against the host without any expiration
```
c = newHTTPClient(u, nil, 10*time.Second, 5*time.Minute)
httpCache[host] = c
```
In this scenario the node when rebooted was still picking the client from cache for any further request
The cached client had resolved the IP that was assigned to the node prior to reboot hence restarting resolved the issue
A fix for this would be to add a custom round tripper in the transport layer which can invalidate the cache and build a new http client in case of a timeout, this way the it will attempt to resolve the new IP for that host

[PWX-35254](https://portworx.atlassian.net/browse/PWX-35254)

**Testing Notes**  
restart one of the storage node and assign it a new IP address without changing the hostname
it will defer the cached client and should build a new client that resolves the correct IP address

[PWX-35254]: https://portworx.atlassian.net/browse/PWX-35254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -30,6 +30,8 @@ func NewClient(host, version, userAgent string) (*Client, error) {
 		return nil, fmt.Errorf("Unable to parse provided url: %v", host)
 	}
 	c := &Client{
+		host:        host,
+		tlsConfig:   nil,
 		base:        baseURL,
 		version:     version,
 		httpClient:  hClient,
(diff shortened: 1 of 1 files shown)


pnookala-px (commented):
The changes look good to me. Can you add Testing Notes?

pnookala-px (approved):

PR #2384, opened 2023-12-14
Fix spec handler and change internal name for fs format options

ppadalia:
Fix spec handler and change internal name for fs format options

Closes #PWX-18297

**Testing Notes**  
Please refer to porx PR for UT: https://github.com/portworx/porx/pull/12772


talakad (approved):
Lookd good to me. Adding Aditya Daani to review as well for any regressions

api/api.proto
```diff
@@ -5628,7 +5628,7 @@ message SdkVersion {
     // SDK version major value of this specification
     Major = 0;
     // SDK version minor value of this specification
-    Minor = 173;
+    Minor = 174;
```
adityadani, line 5631: if you are not changing any proto definitions, you don't need to bump up the SDK version. @lpabon Correct me if I am wrong.

adityadani (approved):

api/api.proto
ppadalia: Thanks @adityadani, I too had seen few PRs where the version was not bumped up. Since we are not sure about it, cautioning on the safer side and bumping it up even though its not strictly required.

PR #2390, opened 2023-12-19
PWX-35430: Honor cluster domains in round-robin balancer.

adityadani:

When a cluster is running with cluster-domains enabled, the round robin balancer should choose the nodes to forward the request within the same cluster domain.

PWX-35430

**Testing Notes**  
Added two UTs that exercise the changed node selection logic based on whether a cluster domain is specified or not.


pp511 (approved):
LGTM. Only caveat would be that in a Metro setup, nodes on the DR side which were participating in the roundrobin balancing will not be participating anymore. Which does reduce the number of nodes which will serve the request but sounds like the right thing to do.

diptiranjanpx (approved):
LGTM

PR #2392, opened 2023-12-19
PWX-35452: Fix mount/unmount race by using MountsInProgress to track …

pnookala-px:
…mounts that are in progress

There is a race condition between Mount and Unmount and the map that tracks the mounted devices can be removed from the mount even as mounts are in progress. This will result in threads that are trying to mount/unmount further hang or fail.

Closes #
or
PWX-35452

**Testing Notes**  
Testing in Progress.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -18,12 +18,12 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/moby/sys/mountinfo"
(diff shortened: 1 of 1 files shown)


adityadani (approved):
It feels a bit weird to use a different lock than the one inside info object to guard the MountsInProgress counter. But this way it reduces the scope of this look and avoid simultaneous mount and unmount request to happen.

PR #2399, opened 2024-01-10
PWX-35601 Add length check before accessing in round robin

dahuang-purestorage:
This PR adds a few check before accessing the filteredNodes:
1. If the length is zero, we return empty string and false (isRemote)
2. Before accessing the array, we take a reminder of the currentIndex with the length of the array.

PWX-35601

**Testing Notes**  
Added simple UT to check the function

### DIFF
--- a/pkg/loadbalancer/roundrobin.go
+++ b/pkg/loadbalancer/roundrobin.go
@@ -142,9 +142,10 @@ func (rr *roundRobin) getTargetAndIncrement(filteredNodes []*api.Node, selfNodeI
 		targetNodeNumber int
 		isRemoteConn     bool
 	)
-	if rr.nextCreateNodeNumber != 0 {
-		targetNodeNumber = rr.nextCreateNodeNumber
+	if len(filteredNodes) == 0 {
+		return "", false
 	}
+	targetNodeNumber = rr.nextCreateNodeNumber % len(filteredNodes)
 	targetNode := filteredNodes[targetNodeNumber]
 	if targetNode.Id != selfNodeID {
 		// NodeID set on the cluster object is this node's ID.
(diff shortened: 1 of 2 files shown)


pkg/loadbalancer/roundrobin_test.go
```diff
@@ -93,3 +93,31 @@ func TestGetRemoteNodeWithDomains(t *testing.T) {
 		}
 	}
 }
+
+// TestGetTargetAndIncrementNullPointer tests the case when getTargetAndIncrement
+// causes null pointer. This happens when the round robin index is not checked against node array
+// length before accessing it.
+//
+// https://portworx.atlassian.net/browse/PWX-35601
+func TestGetTargetAndIncrementNullPointer(t *testing.T) {
```
alicelyu-px, line 102: Super nit (sorry can't help it) -- Null is a Java concept

nrevanna (approved):
LGTM

pkg/loadbalancer/roundrobin.go
```diff
@@ -142,9 +142,10 @@ func (rr *roundRobin) getTargetAndIncrement(filteredNodes []*api.Node, selfNodeI
 		targetNodeNumber int
 		isRemoteConn     bool
 	)
-	if rr.nextCreateNodeNumber != 0 {
-		targetNodeNumber = rr.nextCreateNodeNumber
+	if len(filteredNodes) == 0 {
+		return "", false
```
pureneelesh, line 146: I think `len(filterNodes)` will not be 0. However, since we are checking for that, how will callers react to the empty string as the first return parameter? Will they be able to handle empty mgmt IP?

pkg/loadbalancer/roundrobin_test.go
```diff
@@ -93,3 +93,31 @@ func TestGetRemoteNodeWithDomains(t *testing.T) {
 		}
 	}
 }
+
+// TestGetTargetAndIncrementNullPointer tests the case when getTargetAndIncrement
```
pureneelesh, line 97: nit: It is "index out of bounds" error. Not "nil pointer".

pkg/loadbalancer/roundrobin.go
dahuang-purestorage: Yeah, i agree, it won't be zero.
I just wanna make sure whenever this codebase is expanded, we won't run into null pointer or index out of bound. 

Current caller checks the node length before passing it to the function.

pureneelesh (approved):

PR #2402, opened 2024-01-17
PWX-35559: add correlation tracing to nsm filter Stats

alicelyu-px:


api/client/client.go
```diff
@@ -138,6 +157,41 @@ func unix2HTTP(u *url.URL) {
...
+// shouldRoundTripRetry
+func (c *Client) shouldRoundTripRetry(res *http.Response, err error) bool {
+	if http.ErrHandlerTimeout == err || res == nil {
+		return true
+	}
+	return res.StatusCode == http.StatusRequestTimeout ||
+		res.StatusCode == http.StatusGatewayTimeout
+}
+
+// RoundTrip
```
ggriffiths, line 169: is this required for this change?

api/client/client.go
```diff
@@ -30,13 +30,17 @@ func NewClient(host, version, userAgent string) (*Client, error) {
 		return nil, fmt.Errorf("Unable to parse provided url: %v", host)
 	}
 	c := &Client{
+		host:        host,
```
ggriffiths, line 33: Is this needed? Seems like some extra files were copied here

api/client/client.go
alicelyu-px: > is this required for this change?

This change was on px master. Looks like openstorage master was not vendored into porx master and this change was mistakenly carried over.. I manually removed this change in a later commit. Thanks for the catch!

ggriffiths (approved):
lgtm thanks for addressing other concerns!

PR #2403, opened 2024-01-19
PWX-33580 Run stat command with timeout during NFS Unmount

vlaxman-px:
NFS Unmount call gets routed to RemoveMountPath, which can get stuck in “stats“ system call in case of an offline NFS server. To rectify this, changes has been made where stat command is run remotely along with a timeout of 60 seconds. This helps the system not to hang and retry the command every one minute.

PWX-33580

**Testing Notes**  
I tested the change with an Infof with my name to check if the code is running as expected. The change is working and the retry happens after the specified timeout.
Screenshot 2024-01-19 at 1 56 42 PM

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -4,12 +4,14 @@
 package mount
 
 import (
+	"context"
 	"crypto/md5"
 	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


pnookala-px (approved):
Nit: Try to distinguish ENOENT and Timeout errors.

PR #2408, opened 2024-01-30
Add generic schedule structure; add fields for defrag schedule

twang-ps:
This PR added a generic `Schedule` proto message. It defines a schedule with start time, max duration, and the tasks to run.
Also added `schedule_one_iteration_only` field to `DefragJob` for one-iteration-only defrag schedule.
Added `last_update_time` to `DefragNodeStatus`

Closes #
or
PWX-

**Testing Notes**  
Add testing output or passing unit test output here.


api/api.proto
```diff
@@ -4043,6 +4043,27 @@ message Job {
...
+  // StartTime is the scheduled time to start a run
+  // Valid formats: 
+  // daily=HH:mm  e.g. daily=19:15
+  // weekly=weekday@HH:mm  e.g. weekly=Sunday@19:15
+  // monthly=dd@HH:mm  e.g. monthly=22@19:15
+  string start_time = 2;
+  // MaxDurationHours defines the time limit of one run in hours 
+  double max_duration_hours = 3;
+  // Type is the type of job to run
+  Job.Type type = 4;
```
adikulkarni0, line 4058: Is Job.Type not part of Job that is repeated below?

adikulkarni0: If it is intended to be a Schedule Type lets use that

adikulkarni0 (approved):

api/api.proto
twang-ps: Yes it's intended to denote the schedule type. And it should be one of the Job.Type.

PR #2412, opened 2024-02-09
PWX-32899_pt4: K8s DNS fix

zoxpx:

Autopilot and Stork (which vendor in this library) should not assume that Kubernetes DNS is configured with default `.svc.cluster.local` domain

FIX:
* as a fix, we are modifying the hostnames to `<service>.<endpoint>` (and no `.svc.cluster.local` domain)
* instead, we will rely on the `/etc/resolv.conf` to provide the correct search `<dns-domain>` entry, so DNS resolution will still work correctly

PWX-32899  (part 4)

See also:
* part 1: https://github.com/libopenstorage/operator/pull/1437
* part 2: https://github.com/libopenstorage/stork/pull/1629
* part 3: https://github.com/pure-px/autopilot/pull/250

**Testing Notes**  

Please check stork and autopilot PRs for manual tests results

TODO: This PR change will need to be vendored into `stork` and `autopilot` projects

### DIFF
--- a/volume/drivers/pwx/connection.go
+++ b/volume/drivers/pwx/connection.go
@@ -125,7 +125,7 @@ func (cpb *ConnectionParamsBuilder) BuildClientsEndpoints() (string, string, err
 		return "", "", fmt.Errorf("failed to get k8s service specification: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


adityadani (approved):

zoxpx:
Thanks for the review Aditya -- merging the PR

PR #2419, opened 2024-02-20
add check for shared volumes in getVolumeStats

dahuang-purestorage:
Adds check for shared v4 volumes that have a different attach path than the assumed CSI volume path. 

PWX-35351

**Testing Notes**  
With the fix, we are able to see Used Space on OCP:
image

### DIFF
--- a/csi/node.go
+++ b/csi/node.go
@@ -365,7 +365,8 @@ func (s *OsdCsiServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetV
 
 	var attachPathMatch bool
 	for _, attachPath := range vol.AttachPath {
-		if attachPath == path {
+		sharedPath := fmt.Sprintf("%s/%s", api.SharedVolExportPrefix, id)
+		if attachPath == path || attachPath == sharedPath {
 			attachPathMatch = true
 		}
 	}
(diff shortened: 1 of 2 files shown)


csi/node.go
```diff
@@ -365,7 +365,8 @@ func (s *OsdCsiServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetV
 
 	var attachPathMatch bool
```
adityadani, line 366: In general, I don't understand the reason for this path match check. Regardless of the attachPath we can always return the volume usage for every volume. The attach path does not dictate the volume usage.

adityadani (approved):
lgtm after comment is addressed

dahuang-purestorage:
chatted with @adityadani offline. Removing the attach_path check will break how CSI sanity, so let's keep the attach_path validation.

csi/node.go
```diff
@@ -365,7 +365,8 @@ func (s *OsdCsiServer) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetV
...
 	for _, attachPath := range vol.AttachPath {
-		if attachPath == path {
+		sharedPath := fmt.Sprintf("%s/%s", api.SharedVolExportPrefix, id)
```
pureneelesh, line 368: nit: this line can be moved outside of the for loop

pureneelesh (approved):

PR #2427, opened 2024-03-18
Add Additional logging in Mount/Unmount/Load/Reload

pnookala-px:
There are instances where the mount cache table is going out of sync with the disk status and this is causing issues when mount/unmount calls are made. Log all place where we add / delete entry into the mount table.

PWX-36522.


Pure-AdamuKaapan (approved):
LGTM, hope this helps track it down!

PR #2431, opened 2024-03-20
PWX-36522: In case of devices use Exact match with mount and not a pr…

pnookala-px:
…efix match

PWX-36522: When the devices are of the format /dev/dm-<XX> and we do a prefix search, the search will lead incorrect match for ./dev/dm-<X>. Instead when searching for Target Devices, check for exact match.
Testing:
On the system that had devices in the specified format installed binary and verified that filesystems/devices are rightly identified and pods come up properly.
Tested on a Aetos setup with a mix of RWO/RWX/FADA pods and kept restarting px randomly on all nodes. Pods came up cleanly.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -368,7 +368,7 @@ func (m *Mounter) reload(device string, newM *Info) error {
 	}
 
 	// Purge old mounts.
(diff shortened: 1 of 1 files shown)


Pure-AdamuKaapan (approved):
LGTM, pending testing notes obviously

PR #2453, opened 2024-06-14
PWX-37650: Supports adding the pod name as a locator label for clone/snap restore

Pure-AdamuKaapan:
When we make a clone of a volume, or restore from a snapshot, we only get the name info of the volume. We do not get parameters like the pod name, which are required to determine where to place the volume. This adds the pod name as a label in the locator.

PWX-37650

**Testing Notes**  
Confirmed volume clones are placed correctly inside of pods as expected.


pp511 (approved):
lgtm

PR #2462, opened 2024-08-19
PWX-38618: Pods using FBDA volumes stuck in terminating state. Unmoun…

pnookala-px:
…t even if entry not found in mount table

If for any reason the mount table entry goes stale or like in the case of FBDA where we want to use IP address to mount but nfs internally mounted it using FQDN name there won't be an entry in the mount table corresponding to the mount point  in the mount table. In such cirucumstances when we get an Unmount() request for a path, trigger the Unmont()

Closes #
or
PWX-38618

**Testing Notes**  
Manually Tested the following:
1. Create a pure-file pvc.
2. Mount the pvc directly from one of the nodes using fqdn / hostname.
3. Trigger a pod that uses the pvc. (Verified that the mount uses hostname and mount table entry contains host IP).
4. Delete the pod.
5. Verify that the termination of the pod happens successfully.

### DIFF
--- a/pkg/mount/mount.go
+++ b/pkg/mount/mount.go
@@ -646,26 +646,37 @@ func (m *Mounter) Unmount(
 		// fuse mounts show-up with this key as device.
 		device = value
 	}
+	forceUnmount := false
+	if value, ok := opts[options.OptionsForceUnmount]; ok {
+		forceUnmount = true
+	}
 	info, ok := m.mounts[device]
 	if !ok {
 		logrus.Warnf("Unable to find device %q path %q: in mount table",
 			devPath, path)
 		m.printMountTable()
 		m.Unlock()
-		err := m.mountImpl.Unmount(path, flags, timeout)
-		if err != nil {
-			logrus.Infof("Unmount devPath[%s] from path[%s] failed with error [%v]", devPath, path, err)
-			// If Unmount failed, assume it is due to ErrEnoent and return the failure as enoent,
-			// so that the underlying directory can be removted.
-			err = ErrEnoent
-			return err
-		}
-		logrus.Infof("Unmount of path [%s] successful even though entry didn't exist in mount-table", path)
-		if options.IsBoolOptionSet(opts, options.OptionsDeleteAfterUnmount) {
-			m.RemoveMountPath(path, opts)
+		err := ErrEnoent
+		// If forceUnmount, we need to attempt Unmount even if entry is not there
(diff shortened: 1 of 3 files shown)


pkg/mount/mount.go
```diff
@@ -648,11 +648,23 @@ func (m *Mounter) Unmount(
...
 		m.printMountTable()
 		m.Unlock()
-		return ErrEnoent
+		err := m.mountImpl.Unmount(path, flags, timeout)
+		if err != nil {
+			logrus.Infof("Unmount devPath[%s] from path[%s] failed with error [%v]", devPath, path, err)
+			// If Unmount failed, assume it is due to ErrEnoent and return the failure as enoent,
+			// so that the underlying directory can be removted.
+			err = ErrEnoent
+			return err
```
pp511, line 661: nit.. no need to assign to err. Let's just return ErrEnoent

pp511 (commented):
Changes look ok to me.
@adityadani Can you take a look once and share any concerns?

vlaxman-px (approved):
LGTM

pkg/options/options.go
```diff
@@ -63,6 +63,11 @@ const (
 	// - Attach
 	// It indicates which IO path to use to complete user IO
 	OptionsFastpath = "FASTPATH_STATE"
+	// OptionsForceUnmount is an option to the following Opentstorage Volume API
+	// - Unmount
+	// It is used to issue an umount system call to the requested path even
+	// if the entry for the path is not present in the mount table
+	OptionsForceUnmount = "FORCE_UNMOUNT"
```
adityadani, line 70: There is an actual force flag that is passed to the Unmount function as part of the "flags" argument. So this name is a bit confusing. Instead can we rename this to OptionsUnmountOnErrNoEnt and be more explicit of its usage

pkg/mount/mount_test.go
```diff
@@ -96,6 +98,14 @@ func mountTest(t *testing.T, source, dest string) {
 	require.NoError(t, err, "Failed in unmount")
 }
 
+func forceUnmountTest(t *testing.T, source, dest string) {
```
adityadani, line 101: There should also be others tests to invoke Unmount twice so that your code path gets actually executed. 
- Invoke Unmount twice, second should fail when options are not set
- Invoke Unmount twice, second should succeed when options are set

adityadani (commented):
a few comments

PR #2463, opened 2024-08-20
PWX-38585: Removes extra pod name label

Pure-AdamuKaapan:
We are currently setting the pod name in two separate labels in a clone, which aside from a bad UX just makes things more complicated. This collapses it down to one variable.

It also passes all labels in one case instead of just the pod name label.

PWX-38585

**Testing Notes** 
Manually tested pxctl clone and PVC-based clone and confirmed that labels are piped through properly.
Letting UTs run and fail here so I can fix them.

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -590,9 +590,6 @@ func (s *OsdCsiServer) CreateVolume(
 		if spec.GetFADAPodName() != "" {
 			labels[api.SpecPurePodName] = spec.GetFADAPodName()
 		}
-		delete(labels, intreePvcNameKey)
-		delete(labels, intreePvcNamespaceKey)
-		delete(labels, api.SpecParent)
 		cloneResp, err := volumes.Clone(ctx, &api.SdkVolumeCloneRequest{
 			Name:             req.GetName(),
 			ParentId:         source.Parent,
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/libopenstorage/openstorage/api"
-	"github.com/libopenstorage/openstorage/api/server/sdk"
 	"github.com/libopenstorage/openstorage/pkg/grpcutil"
 	"github.com/libopenstorage/openstorage/pkg/units"
 	"github.com/libopenstorage/openstorage/pkg/util"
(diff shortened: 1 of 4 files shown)


csi/controller.go
```diff
@@ -587,14 +586,17 @@ func (s *OsdCsiServer) CreateVolume(
...
 	} else {
-		clonedMetadata := getClonedPVCMetadata(locator)
+		labels := locator.GetVolumeLabels()
 		if spec.GetFADAPodName() != "" {
-			clonedMetadata[sdk.FADAPodLabelKey] = spec.GetFADAPodName()
+			labels[api.SpecPurePodName] = spec.GetFADAPodName()
 		}
+		delete(labels, intreePvcNameKey)
+		delete(labels, intreePvcNamespaceKey)
+		delete(labels, api.SpecParent)
```
pp511, line 595: Can you help me understand why do we need to remove this label here?

csi/controller.go
```diff
@@ -587,14 +586,17 @@ func (s *OsdCsiServer) CreateVolume(
...
 		cloneResp, err := volumes.Clone(ctx, &api.SdkVolumeCloneRequest{
 			Name:             req.GetName(),
 			ParentId:         source.Parent,
-			AdditionalLabels: clonedMetadata,
+			AdditionalLabels: labels,
```
pp511, line 599: I am not entirely sure how the change is fixing the clone issue. Can you briefly describe what's happening?

pp511 (commented):
I have some questions. Just trying to understand how the change is helping. Can you respond?

csi/controller.go
Pure-AdamuKaapan: This is primarily so that other generic labels will be passed on - previously we were only going to pass the FADA pod value.

As discussed in slack, Porx PR coming soon which will help this make more sense.

csi/controller.go
Pure-AdamuKaapan: The PVC name and namespace keys don't make sense for a clone, as they are no longer attached to the same PVC and namespace.

Deleting the parent label was to make the tests happy (if it wasn't present before I wasn't sure if we wanted it present now), but I can change that to still be present. I don't _think_ it breaks any behavior to have it there.

csi/controller.go
pp511: Yes changes look ok. Was just not able to understand how it helps. Will take a quick look at the Porx PR and approve this if I don't have any concerns.

csi/controller.go
pp511: Yes my concern was with the parent label only. If it's not being used further then I guess it's ok to remove it.

pp511 (approved):
Looks good.

csi/controller.go
Pure-AdamuKaapan: Actually was able to remove this chunk!

PR #2464, opened 2024-08-20
PWX-38585: Removes extra pod name label

Pure-AdamuKaapan:
*Master CP of #2463*

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -614,14 +614,17 @@ func (s *OsdCsiServer) CreateVolume(
 		}
 		newVolumeId = createResp.VolumeId
 	} else {
-		clonedMetadata := getClonedPVCMetadata(locator)
+		labels := locator.GetVolumeLabels()
 		if spec.GetFADAPodName() != "" {
-			clonedMetadata[sdk.FADAPodLabelKey] = spec.GetFADAPodName()
+			labels[api.SpecPurePodName] = spec.GetFADAPodName()
 		}
+		delete(labels, intreePvcNameKey)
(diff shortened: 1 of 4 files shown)


csi/controller.go
```diff
@@ -614,14 +613,17 @@ func (s *OsdCsiServer) CreateVolume(
 		}
 		newVolumeId = createResp.VolumeId
 	} else {
-		clonedMetadata := getClonedPVCMetadata(locator)
+		labels := locator.GetVolumeLabels()
 		if spec.GetFADAPodName() != "" {
-			clonedMetadata[sdk.FADAPodLabelKey] = spec.GetFADAPodName()
+			labels[api.SpecPurePodName] = spec.GetFADAPodName()
 		}
+		delete(labels, intreePvcNameKey)
```
adityadani, line 621: Why do we need to delete these?

pp511 (approved):
LGTM

csi/controller.go
Pure-AdamuKaapan: My thought was that the PVC name and namespace would no longer be applicable on the clone (and in unit tests when I added this, the labels were empty). However, let me sanity check what happens when I remove this and try a clone on a PVC volume, I'll see if it gets the label of the source or an empty string.

Pure-AdamuKaapan: Actually this doesn't seem necessary - while I'll have to update unit tests, the end functionality is fine without these delete statements.

adityadani: I thought these would be the name and namespace of the PVC (clone) that is getting created and not of the parent volume. Isn't that the case?

PR #2482, opened 2024-09-09
PWX-38585: Adds additional labels to clone requests

Pure-AdamuKaapan:
Volume clone requests can have additional labels, which were not being plumbed through.

PWX-38585


harsh-px (approved):

Pure-AdamuKaapan:
Confirmed this with a manual test - ran `pxctl v clone` and set the fa pod name label, confirmed it was passed into the volume properly.

PR #2499, opened 2024-10-24
PWX-39688 Wai till local snapshot is taken to send create snapshot response in case of cloud snap

sradhakrishnan-px:
This PR will set the creation time of the snapshot in CSI response object as the local snapshot time. As per CSI spec, the creation time of the snapshot is the time when storage system took the point in time snapshot.

Closes #
or
PWX-39688

**Testing Notes**  
Tested creating a snapshot and made sure that the creation time is coming up in the kubectl get volumesnapshot <name> -o yaml output

### DIFF
--- a/csi/controller.go
+++ b/csi/controller.go
@@ -1069,6 +1069,13 @@ func (s *OsdCsiServer) createCloudBackup(
 			return nil, status.Errorf(codes.NotFound, "Snapshot %s not found in status map", csiSnapshotID)
 		}
 		logger.Infof("Status of snapshot %v", snapshotStatus.Status)
+
+		// if local snapshot is not taken, return an error so that the request will be retried
+		if !snapshotStatus.IsLocalSnapshotTaken() {
+			logger.Infof("Local snapshot for the cloud snapshot is not ready")
+			return nil, fmt.Errorf("local snapshot for the cloud snapshot is not ready")
+		}
+
 		isBackupReady := snapshotStatus.Status == api.SdkCloudBackupStatusType_SdkCloudBackupStatusTypeDone
 		return &csi.CreateSnapshotResponse{
 			Snapshot: &csi.Snapshot{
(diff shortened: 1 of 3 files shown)


csi/controller.go
```diff
@@ -1523,13 +1524,42 @@ func waitForVolumeId(ctx context.Context, req *csi.CreateVolumeRequest, volumes
...
+			TaskId: csiSnapshotID,
+		})
+		if err != nil {
+			logger.Errorf("Failed to get snapshot status after snapshot create")
+			return nil, true, err
+		}
+		snapshotStatus, ok := backupStatusResponse.Statuses[csiSnapshotID]
+		if !ok {
+			return nil, true, status.Errorf(codes.NotFound, "Snapshot %s not found in status map", csiSnapshotID)
+		}
```
pnookala-px, line 1549: This looks good to me. Only suggestion, is can we really test it by adding a sleep in porx code and see how this behaves when porx code doesn't respond within 2 minutes.

csi/controller.go
```diff
@@ -1523,13 +1524,42 @@ func waitForVolumeId(ctx context.Context, req *csi.CreateVolumeRequest, volumes
...
 	if err != nil {
 		return "", err
 	}
 
 	return id.(string), err
 }
 
+func waitForLocalSnapshot(ctx context.Context, csiSnapshotID string, cloudBackupClient api.OpenStorageCloudBackupClient, logger *logrus.Entry) (*api.SdkCloudBackupStatus, error) {
+	// the create snapshot operation does not take local snapshot immediately. As per CSI spec, the CreationTime of the snapshot
+	// is wen the local snapshot was taken, hence wait till local snap is available
```
pnookala-px, line 1537: spell check -> wen to when

pnookala-px (approved):
Changes look good, I only want to test the scenario wherein, let us say we creat

csi/controller.go
sradhakrishnan-px: Tested this out, byt commenting out my fix to send the local snap time. Initially the snapshot did not happen and the following error came 
```Status:
  Bound Volume Snapshot Content Name:  snapcontent-ae7cc560-b837-480d-bd27-acf1a11232ba
  Error:
    Message:     Failed to check and update snapshot content: failed to take snapshot of the volume 582374246902790351: "rpc error: code = Unknown desc = local snapshot for the cloud snapshot is not ready"
    Time:        2024-10-29T04:46:21Z
```
Then I restarted px with the fix, and the snaphsot completed
```
Events:
  Type    Reason            Age   From                 Message
  ----    ------            ----  ----                 -------
  Normal  CreatingSnapshot  30m   snapshot-controller  Waiting for a snapshot default/new-snapshot-test-10 to be created by the CSI driver.
  Normal  SnapshotCreated   16m   snapshot-controller  Snapshot default/new-snapshot-test-10 was successfully created by the CSI driver.
  Normal  SnapshotReady     16m   snapshot-controller  Snapshot default/new-snapshot-test-10 is ready to use.
```

PR #2500, opened 2024-10-26
PWX-39445: Option to disable quorum in gossip

dgoel-px:
This is required for Px Lite (Pure CSI) that will only run Storageless nodes and therefore will NOT obey any quorum.

PWX-39445

**Testing Notes**  
Tested these changes with tech preview build for customer. Further testing will be done with Px changes.


adityadani (approved):

adityadani:
@dgoel-px We should remove these changes from release-9.9

PR #2502, opened 2024-11-25
PWX-40172 (master): makes chattr ignore 'inappropriate ioctl' errors

Pure-AdamuKaapan:
In some OSes like SLE Micro, the `/var` directory is mounted using an overlay filesystem. Something in this setup does not support `chattr`. In this case, we should just ignore `inappropriate ioctl for device` errors as there is nothing we can do.

PWX-40172

**Testing Notes**
* Confirmed PX installs on a Harvester system
* Added UTs that make an NTFS loopback device (as NTFS does not support chattr either), mounts it, and confirms our chattr code still succeeds.

### DIFF
--- a/pkg/chattr/chattr.go
+++ b/pkg/chattr/chattr.go
@@ -16,71 +16,57 @@ const (
 	lsattrCmd = "lsattr"
 )
 
-func AddImmutable(path string) error {
-	return ForceAddImmutable(path, false)
+func isInappropriateIoctl(err string) bool {
+	lowerErr := strings.ToLower(err)
+	// "inappropriate ioctl for device" returned inside PX container
+	// "invalid argument while setting flags" returned on host
+	return strings.Contains(lowerErr, "inappropriate ioctl for device") || strings.Contains(lowerErr, "invalid argument while setting flags")
 }
 
-func ForceAddImmutable(path string, force bool) error {
+func runChattr(path string, arg string, force bool) error {
 	chattrBin := which(chattrCmd)
 	if _, err := os.Stat(path); err == nil {
-		var cmd *exec.Cmd
+		args := []string{arg}
 		if force {
-			cmd = exec.Command(chattrBin, "+i", "-f", path)
-		} else {
-			cmd = exec.Command(chattrBin, "+i", path)
+			args = append(args, "-f")
 		}
+		args = append(args, path)
+		cmd := exec.Command(chattrBin, args...)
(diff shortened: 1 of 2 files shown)


pkg/chattr/chattr.go
```diff
@@ -16,71 +16,57 @@ const (
...
 			//  "Operation not supported while reading flags"
 			if force && cmd.ProcessState.ExitCode() == 1 {
 				return nil
 			}
-			if force {
-				return fmt.Errorf("%s +i -f failed: %s. Err: %v", chattrBin, stderr.String(), err)
+
+			stderrStr := stderr.String()
+			if isInappropriateIoctl(stderrStr) { // Returned in case of filesystem that does not support chattr
+				return nil
```
zoxpx, line 48: Uh.. this override will unconditionally turn off the `inappropriate ioctl for device` (also `invalid argument while setting flags`) errors on `chattr` for ALL PLATFORMS.

I wonder if we could avoid it?

Some suggestions that might work:

1. have `AddImmutable()` return the `inappropriate ioctl for device` error -- and let the caller handle the "error ignoring", or
2. extend the `AddImmutable()` by adding `AddImmutable(..., errorsToIgnore []error)` and have the caller populate the "errors to ignore" list, or
3. have the `chattr` command "overrideable" -- so that the caller may _change_/_override_ the OpenStorage's `chattr` command with own implementation, before calling `AddImmutable()` [or, mounting volume]

zoxpx: Hmm.. I see that your comment above said _"// Returned in case of filesystem that does not support chattr"_
* so, I suppose you are making an argument that _"chattr aways returns the ENOTTY when filesystem does not support the chattr'ing"_
* my concern was that we'll also be masking the "legal errors" -- (maybe use a "wrong device" like perhaps `chattr /dev/stderr`) -- which is why I'd look for means to reduce the "blast radius" of this change, and limit it only to the `SLE-Micro` Linux platform

Let's see of others from the team would have any comments / feedback

Pure-AdamuKaapan: That was my assumption (it always returns one of those two errors I linked), but I do not know that for sure and I agree that it would mask "legal" errors.

The "override" option seems the most feasible, especially because when we run that platform detection code we can have it also set perhaps `IgnoreChattrIOCTLErrors = true` or something. Then the caller has the responsibility of knowing what cases that it matters in.

PR #2504, opened 2025-03-20
PWX-42338: add volumelocator id in query params for volume enumerate api call

strivedi-px:
Stork was sending the volume IDs in the Locator struct data but that was not being added to the query params in the volume enumerate REST API call. This PR fixes that.

[PWX-42338
](https://purestorage.atlassian.net/browse/PWX-42338)

**Testing Notes** 
```
In a 10 node cluster with 750 volumes and 750 snaps, when putting 30 pods in scheduling in parallel, we are seeing following

With released stork:
Filter request handling avg time taken is - 3-5secs
Priortize request handling avg time taken is - 5-7 secs

With the fix:
Filter request handling avg time taken is - 0-1sec
Priortize request handling avg time taken is - 0-1sec
```

### DIFF
--- a/api/client/volume/client.go
+++ b/api/client/volume/client.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
 	"github.com/libopenstorage/openstorage/api"
 	"github.com/libopenstorage/openstorage/api/client"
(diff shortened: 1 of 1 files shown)


diptiranjanpx (approved):

javedlingasur-px:
Merged this PR as per the request PTX-30530

